Issue #23, opened 2011-07-05
Authentication fails with more than one key in agent


gellule:
It seems that when I have more than one key held by the agent, any authentication that would require anything but the first key would fail. The trace is copied down below. I also added some dedug info at the packet level. Could it be that there should not be a [service-request] following the [userauth-failure] with the wrong key, but directly a new [userauth-request]?
Thanks,
-Gellule

DEBUG:paramiko.transport:starting thread (client mode): 0x2514550L
INFO:paramiko.transport:Connected (version 2.0, client 3.2.9.1)
Write packet [kexinit], length 406
Read packet [ignore], length 5
Read packet [kexinit], length 461
DEBUG:paramiko.transport:kex algos:['diffie-hellman-group1-sha1'] server key:['ssh-dss'] client encrypt:['aes128-cbc', '3des-cbc', 'twofish128-cbc', 'cast128-cbc', 'twofish-cbc', 'blowfish-cbc', 'aes192-cbc', 'aes256-cbc', 'twofish192-cbc', 'twofish256-cbc', 'arcfour'] server encrypt:['aes128-cbc', '3des-cbc', 'twofish128-cbc', 'cast128-cbc', 'twofish-cbc', 'blowfish-cbc', 'aes192-cbc', 'aes256-cbc', 'twofish192-cbc', 'twofish256-cbc', 'arcfour'] client mac:['hmac-sha1', 'hmac-sha1-96', 'hmac-md5', 'hmac-md5-96'] server mac:['hmac-sha1', 'hmac-sha1-96', 'hmac-md5', 'hmac-md5-96'] client compress:['none', 'zlib'] server compress:['none', 'zlib'] client lang:[''] server lang:[''] kex follows?False
DEBUG:paramiko.transport:Ciphers agreed: local=aes128-cbc, remote=aes128-cbc
DEBUG:paramiko.transport:using kex diffie-hellman-group1-sha1; server key type ssh-dss; cipher: local aes128-cbc, remote aes128-cbc; mac: local hmac-sha1, remote hmac-sha1; compression: local none, remote none
Write packet [kex30], length 133
Read packet [ignore], length 5
Read packet [kex31], length 630
Write packet [newkeys], length 1
Read packet [ignore], length 5
Read packet [newkeys], length 1
DEBUG:paramiko.transport:Switch to new keys ...
DEBUG:paramiko.transport:Trying SSH agent key fbd3bc799a422891ffc58f9212539a13
Write packet [service-request], length 17
[...]

PR #24, opened 2011-07-05
Only one service request per authentication serie

gellule:
This fixes the issue I had with the authentication key not being the first one delivered by ssh-agent. The problem was paramiko sending a service request for each ssh key held by the agent. The server would disconnect on the second service request. With the changes, (1) the authentication is instantiated only once for the full list of keys provided by the agent, (2) the service request is only performed once.

Cheers!


jansegre:
I checked and this fixes the issue I commented about on #23. I will port this to the latest master and open a new PR.

bitprophet:
Rolling this into #478

PR #58, opened 2011-12-16
banner functionality

Jhoanor:
I would like to be able to obtain the banner supplied by a server to which my python script connects using paramiko.
This is to be able to supply reasons for not be able to connect at a certain moment, (can be handy for legal issues as well)
The following changes are in line wich the get_username method.
Tested and works. :-)
I have not updated documentation in the pull request but this line can be added I suppose in the class Transport documentation on http://www.lag.net/paramiko/docs/ :
string get_banner(self)
            Return the banner supplied by the remote server.

### DIFF
--- a/paramiko/auth_handler.py
+++ b/paramiko/auth_handler.py
@@ -45,6 +45,7 @@ class AuthHandler (object):
         self.authenticated = False
         self.auth_event = None
         self.auth_method = ''
+        self.banner = None
         self.password = None
         self.private_key = None
         self.interactive_handler = None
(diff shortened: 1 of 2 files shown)


jossef:
```
# !/usr/bin/python

import paramiko


def grab_banner(ip_address, port):
    client = paramiko.SSHClient()
    try:
        client.connect(ip_address, port=port, username='username', password='bad-password-on-purpose')
    except:
        return client._transport.get_banner()


if __name__ == '__main__':
    print grab_banner('192.168.1.26', 22)
```

PR #62, opened 2012-03-21
Add timeout for handshake completion.

dacut:
This adds a mechanism for timing out a connection if the ssh handshake never completes.

When managing large fleets, we not-infrequently encounter cases where a host is alive enough to accept an SSH connection but not enough to accept commands.  (I've seen this with failing disks, for example -- my hypothesis is enough of the ssh daemon is cached in memory, but the shell or other libraries can't be read.)

(Internal p4 CLN 5034409; patch approved for release via TT/0013249552)
author: pala@amazon.com

### DIFF
--- a/paramiko/packet.py
+++ b/paramiko/packet.py
@@ -100,6 +100,10 @@ class Packetizer (object):
         self.__keepalive_last = time.time()
         self.__keepalive_callback = None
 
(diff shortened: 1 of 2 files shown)


bitprophet:
Rolling into #491

PR #65, opened 2012-04-17
Add method for "posix-rename@openssh.com" extension.

wt:
I have added this method manually, but what I am doing just feels wrong. I'd rather use the extension advertisement mentioned at section 3.3 of [1] and just make rename use the atomic method if it's available. Do you have any pointers for finding the info?

[1]http://www.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/PROTOCOL?rev=HEAD;content-type=text/plain

Thanks,
wt


bitprophet:
See #471

PR #78, opened 2012-08-23
Fail more gracefully when encountering a missing mtime.

btimby:
I actually have a file system that does this, and the current code results in
prepending (unknown date) to the file name in the SFTP client. This
prevents changing into the directory or downloading files, as the
client sends the name including the (unknown date) to the server.
By using a space, this case is handled properly (missing date, file
name unaffected).


bskinn:
Hi! I'm helping bitprophet by going through the issues and PRs backlog, and for anything more than a few years old, we're just closing them out for now.

If this is still a live bug, please reopen the PR and comment to let us know it's still of interest. (And, apologies for the long wait!)

Thanks!

PR #84, opened 2012-10-04
BaseSFTP._write_all no longer interprets 0 as error

gmacon:
When talking to a high-traffic SFTP server, send() may return 0,
but this is not an error.  This change treats the 0 return the same
way as Packetizer.write_all.


gmacon:
I think this can probably be closed as OBE (considering that the latest version is 1.15 and this patch is for 1.8).

PR #197, opened 2013-08-08
Fix hang when sending packets on closed connection

novas0x2a:
credit to Michael Schrijver michael@hating.biz for finding the problem


kparal:
Paramiko hangs indefinitely if connection goes down, which causes issues in our projects (and where doesn't?). There seem to be two different pull requests to fix this, this one and #470. Also there are several reported issues about this, at least #319 and #503. Could one of the patches please be moved forward and merged?

PR #209, opened 2013-09-14
Make evaluation of key auth available for two-factor auth analysis.

sunweaver:
When authenticating with an explicitly given key_filename, also evaluate
the key based authentication and check if two-factor auth is required.

Without the fix shipped with this pull request, two-factor auth fails
whenever the SSH key is specified via the key_filename parameter in
SSHClient().connect().


ploxiln:
looks like this was done in c9441c4a2

PR #211, opened 2013-09-16
README: clarify that API docs must be generated

meonkeys:
Also, remove an extra colon.


bitprophet:
Obviated by the doc shakeup a while back.

PR #238, opened 2013-12-01
Make comment field from Agent available

gregorlarson:
Add .comment field to AgentKey.

### DIFF
--- a/paramiko/agent.py
+++ b/paramiko/agent.py
@@ -73,8 +73,7 @@ class AgentSSH(object):
             raise SSHException('could not get keys from ssh-agent')
         keys = []
         for i in range(result.get_int()):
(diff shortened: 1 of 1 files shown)


bitprophet:
Note to self, this needs a tiny bit of tweaking in the class constructor to be backwards compatible.

bitprophet:
On second thought, I don't quite understand what this is doing (re: calling `.get_string()` [which is actually now `.get_binary()`] multiple times in a row) despite it being pretty small, and would rather punt for now.

@lndbrg if you recall the specifics here re: how this mechanism works & what its implications are for the general case (do all standard ssh-agent setups have this comment field? could this barf for a subset of users? etc), do drop me a hint :)

PR #272, opened 2014-02-22
Don't hash the input hostname when checking a host key if the input host...

sigmunau:
...name is already hashed

This happens in the loop checking if a hostkey is already loaded when
loading host key files. This reduces time to load host keys file with
about two seconds for me


prosegay:
I'm on paramiko 2.0.2 and this is still like 10x slower than what can be done with Popen ssh (I'm getting about 1s per command, consistent with what was noted above as a bit improvement). setting known_hosts=False brings that down to about 0.2s per cmd. using popen/ssh is < 0.1s per cmd (I didn't even measure it)

PR #304, opened 2014-04-06
handle "out of space" and "quota exceeded"

lukas227:
Please handle "out of space" and "quota exceeded" errors with an appropriate IOErrors. What I got from duplicity was:
BackendException: sftp put of [...](as [...]) failed: Unsupported operation

At the time I investigated, the incomplete backup had been cleaned up already (and there was quite some space left). It took me some time to identify the cause of the cryptic BackendException.


bitprophet:
See https://github.com/paramiko/paramiko/issues/321#issuecomment-642928315 for some context around this. tl;dr openssh doesn't actually support these higher protocol numbers so many/most of our users wouldn't see a benefit here. I'm curious what server the OP was targeting, assuming it was something else which might support more of the errors from the RFC.

If this were updated to account for the entire slate of numbers from the RFC (link in that comment above) and based on a more recent copy of the main branch, I'd still consider merging.

lukas227:
Thanks for your reply. I was using an sftp backup space provided by my server hoster, but I have no idea what software they were using.

I've personally no need for this any more and since (based on your comment) there is very limited usefulness to this overall I'm closing this pull request.

PR #326, opened 2014-05-14
look_for_keys behavior slightly simpler, easier to maintain

Janrain-Colin:
I've been using code from client.py, and I think this improvement would be for the better.


ploxiln:
a cleanup similar to this was done as a small part of #972

bitprophet:
Yea I think this can be considered covered now. Thanks! (Also of note, this area still needs more love re: #387 .)

Issue #327, opened 2014-05-19
In some cases paramiko raises a TypeError when connecting


boyvanduuren:
I'm using paramiko to connect to different types of servers. Some of these servers run Debian Squeeze, the openssh-server version for Squeeze has the following bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=659000

This means that even though I have a valid key for certain users on a Squeeze server, i'll still get the following error (output from `ssh <server> -l root -v`, ssh-agent enabled):

```
debug1: Remote: Your host '<snip>' is not permitted to use this key for login.
debug1: Remote: Your host '<snip>' is not permitted to use this key for login.
```

After that the connection is still setup successfully though.
When I connect to the same server using paramiko, the following happens:

```
Python 3.4.0 (default, Apr 27 2014, 23:33:09) 
[GCC 4.8.2 20140206 (prerelease)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from paramiko.client import *
>>> client = SSHClient()
>>> client.set_missing_host_key_policy(AutoAddPolicy())
>>> client.connect('<ip>', username='root', allow_agent=True)
>>> Unknown exception: Can't convert 'int' object to str implicitly
Traceback (most recent call last):
  File "/usr/lib/python3.4/site-packages/paramiko/transport.py", line 1431, in run
    self._parse_debug(m)
  File "/usr/lib/python3.4/site-packages/paramiko/transport.py", line 1982, in _parse_debug
    self._log(DEBUG, 'Debug msg: ' + util.safe_string(msg))
  File "/usr/lib/python3.4/site-packages/paramiko/util.py", line 121, in safe_string
    out += c
TypeError: Can't convert 'int' object to str implicitly
```

Casting `c` to a string using `chr()` in `util.py` seems to fix the issue:

``` diff
121c121
<             out += c

---
>             out += chr(c)
```

PR #328, opened 2014-05-19
Issue #327: safe_string TypeError

boyvanduuren:
In some cases safe_string gets an array
of integers as argument. safe_string then
tries to concatenate those integers onto a
string, resulting in a TypeError.

### DIFF
--- a/paramiko/util.py
+++ b/paramiko/util.py
@@ -117,8 +117,10 @@ def unhexify(s):
 def safe_string(s):
     out = ''
     for c in s:
-        if (byte_ord(c) >= 32) and (byte_ord(c) <= 127):
+        if (type(c) is int):
             out += chr(c)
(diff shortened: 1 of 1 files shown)


sbraz:
@bitprophet I still see this behaviour on Debian 8 with paramiko 1.15.1-1. I applied the fix and everything seems to work fine now.
Is there anything that needs to be done before this is merged?

ploxiln:
provide a good test case to show what triggers the error, in paramiko 1.17.x or later

sbraz:
@ploxiln I didn't realise Debian's paramiko was so ancient :/
It works fine with paramiko-2.2.1 from PyPI and Python 3.4. You can probably close this!

ploxiln:
FWIW, debian 9 "stretch" was released as stable a couple months ago and includes paramiko 2.0

Issue #161, opened 2013-04-26
Support stronger MAC schemes


EtiennePerot:
As of this writing, the strongest MAC supported by Paramiko is `hmac-sha1`. Since [OpenSSH 5.9](http://www.openssh.com/txt/release-5.9), stronger MAC schemes such as `hmac-sha2-512` and `hmac-sha2-256` are available.

Additionally, [OpenSSH 6.2](http://www.openssh.com/txt/release-6.2) added Encrypt-Then-MAC (ETM) schemes as well, which are [even](http://www.daemonology.net/blog/2009-06-24-encrypt-then-mac.html) [safer](http://crypto.stackexchange.com/questions/202/should-we-mac-then-encrypt-or-encrypt-then-mac) and are therefore used by default since 6.2. These are available under OpenSSH as `hmac-sha2-512-etm@openssh.com`, `hmac-sha2-256-etm@openssh.com`, `hmac-sha1-etm@openssh.com`, etc.

It would be nice if Paramiko kept up with this.

ThomasWaldmann:
OMG, I am stumbling over a 2013 ticket. :-(

I am trying to use paramiko as a client, doing sftp against some relatively securely configured server.

```
    t.connect(hostkey, user, passwd)
  File "...env/lib/python3.5/site-packages/paramiko/transport.py", line 1218, in connect
    self.start_client()
  File "...env/lib/python3.5/site-packages/paramiko/transport.py", line 587, in start_client
    raise e
  File "...env/lib/python3.5/site-packages/paramiko/transport.py", line 2010, in run
    self._handler_table[ptype](self, m)
  File "...env/lib/python3.5/site-packages/paramiko/transport.py", line 2125, in _negotiate_keys
    self._parse_kex_init(m)
  File "...env/lib/python3.5/site-packages/paramiko/transport.py", line 2365, in _parse_kex_init
    raise SSHException("Incompatible ssh server (no acceptable macs)")
paramiko.ssh_exception.SSHException: Incompatible ssh server (no acceptable macs)
```

From the log:
```
client mac:[
'hmac-sha2-512-etm@openssh.com',
'hmac-sha2-256-etm@openssh.com',
'hmac-ripemd160-etm@openssh.com',
'umac-128-etm@openssh.com'
]
server mac: [
'hmac-sha2-512-etm@openssh.com',
'hmac-sha2-256-etm@openssh.com',
'hmac-ripemd160-etm@openssh.com',
[...]

PR #356, opened 2014-07-15
Patches to support sha256 based hmac and kexgex

zamiam69:
This is based on forks by EtiennePerot and ashb who did most of the work. Changing the remaining sha1 references in `transport.py` let's me connect to hardened openssh servers again.


paramiko/transport.py
```diff
@@ -96,9 +96,13 @@ class Transport (threading.Thread, ClosingContextManager):
 
     _preferred_ciphers = ('aes128-ctr', 'aes256-ctr', 'aes128-cbc', 'blowfish-cbc',
                           'aes256-cbc', '3des-cbc', 'arcfour128', 'arcfour256')
-    _preferred_macs = ('hmac-sha1', 'hmac-md5', 'hmac-sha1-96', 'hmac-md5-96')
+    _preferred_macs = ('hmac-sha1', 'hmac-md5', 'hmac-sha1-96', 'hmac-md5-96',
+                       'hmac-sha2-256')
```
dstufft: Is this in order of preference? If it is then ``hmac-sha2-256`` should be first.

Akendo:
There has been a quite poplar blog entry https://stribika.github.io/2015/01/04/secure-secure-shell.html 
there are some good examples for harden everything in regards to openssh. 

A lot of people did this and there are more then ever lusting for this patch.

paramiko/transport.py
sindarina: If it works anything like the OpenSSH implementation, then yes, they are in order of preference, so I would suggest putting it first.

mgogoulos:
ping, this seems ready to be merged, right?

cyphase:
I know I can use the fork or modify the code myself for now, but is there an ETA for when this will be merged?

gertvdijk:
Sorry to comment on this again, but it's really taking a very long time to get SHA-2 support in Paramiko. Note that the original pull request (#161) was opened in *April 2013*. From that point on, it's mostly "here's an improvement" new pull request or complaints about why it isn't merged yet. It's not getting anywhere this way.

I'd suggest to merge this in and improve the situation from there, rather than keeping this open until it's perfect to the letter.

mgogoulos:
@gertvdijk  +1

aneeshusa:
+1 please merge this. If there's something holding this up, please let us know so we can help.

Russell-Jones:
@aneeshusa see #387

paramiko/transport.py
Russell-Jones: For reference, one of the tickets about auth key handling is https://github.com/paramiko/paramiko/issues/387

bitprophet:
Thought I noted this earlier, but this is definitely at/near the top of Paramiko's priority list, just FYI. Will be digging into said list once another (related) project reaches an important milestone, which should be soon.

cyphase:
While this version works with a server that 1.15.2 does not work with (due to no acceptable kex algorithm), it will raise an error when trying to connect to a different server that 1.15.2 does work with. Here's some example code and the traceback:

    # bug_in_paramiko_pull356.py

    import paramiko

    client = paramiko.SSHClient()

    hostname = 'works-with-1-15-2.example'
    username = 'USER'
    password = 'PASS'

    client.connect(hostname, 22, username, password, allow_agent=False)

Raises the following exception:

      File "bug_in_paramiko_pull356.py", line 12, in <module>
        client.connect(hostname, 22, username, password, allow_agent=False)
      File "paramiko/client.py", line 307, in connect
        look_for_keys, gss_auth, gss_kex, gss_deleg_creds, gss_host)
      File "paramiko/client.py", line 519, in _auth
        raise saved_exception
    paramiko.ssh_exception.SSHException: Invalid packet blocking

I'd rather not mention the particular server publicly, but I'm certainly willing to mention it privately. Ideally we can figure out the relevant OpenSSH server options that trigger the bug.

kiddick:
cyphase, same error here!
Is there any workaround for this issue?
So, what about merge?

igoraj:
+1

rata:
Any news when a release including this is planned ?

It's actually really of disturbing because of https://weakdh.org/sysadmin.html

gertvdijk:
@rata 
bq. It's actually really of disturbing because of https://weakdh.org/sysadmin.html

Well, `diffie-hellman-group14-sha1` in the key exchange is already supported it seems, so no panic regarding that. Please be aware that this PR/issue is about the symmetric cipher/MAC chosen after the key exchange and that the Logjam attack on OpenSSH is an attack on the key exchange. Don't confuse the two (OpenSSH configuration options `KexAlgorithms` vs `MACs`).

rata:
@gertvdijk sorry, I never actually read the pull req :-D. Just reported https://github.com/paramiko/paramiko/issues/532 and he told that it was addressed in this pull req.

So, I guess it should be re-opened :)

aaronbieber:
We just hardened our internal SSH daemons across the board to align with best practices, which now breaks our entire Paramiko toolchain. Any updates on when this can be merged?

cyphase:
@kiddick, the workaround I've been using is to manually use this patched version when connecting to a server that requires it. Only works because the particular program I'm maintaining only needs to connect to one SSH server right now, so I literally just replace the paramiko package in site-packages with this version.

aaronbieber:
I confirmed that the patch functions for me as well, so I will be building
my own egg/wheel to host on our internal Python repo. I do hope that this
gets into master soon, though, so that we can all benefit from bug fixes
and increased security.

baxeno:
Happy birthday #356 you are now 1 year old.

Just ran into this issue after hardening Dropbear in our embedded Linux product.
First thought was that our auto test servers was using an old version of Paramiko, nope 1.15.2 was newest pypi release (https://pypi.python.org/pypi/paramiko/1.15.2).

As progress with 2.0 seems kind of slow, have you considered releasing a 1.16 with this fix?

paramiko/kex_gex.py
```diff
@@ -87,7 +88,7 @@ def parse_next(self, ptype, m):
             return self._parse_kexdh_gex_reply(m)
         elif ptype == _MSG_KEXDH_GEX_REQUEST_OLD:
             return self._parse_kexdh_gex_request_old(m)
-        raise SSHException('KexGex asked to handle packet type %d' % ptype)
+        raise SSHException('KexGex %s asked to handle packet type %d' % self.name, ptype)
```
thusoy: This statement is missing a set of parentheses around the format arguments, it needs to be `% (self.name, ptype))`

GuyShaanan:
Using this patch, I was indeed able to connect using SHA2-256.
*However*, I've lost the ability to connect using SHA1 (got errors like @cyphase https://github.com/paramiko/paramiko/pull/356#issuecomment-101147799)

Reading through this comment (https://github.com/paramiko/paramiko/pull/356#discussion_r23203330) I changed `_preferred_macs` order and put `hmac-sha2-256` first - and it solved the problem for me.

bitprophet:
See also #581 which does at least some of the work needed to handle SHA512 (on top of SHA256).

travisthomas:
+1

pixelchutes:
Hoping that this support is able to be merged into the project soon, as it enables connectivity against hardened OpenSSH instances. Select projects relying on paramiko to connect to these hardened installs currently require manual patching or decreasing the security of the ssh server in order to communicate out-of-the-box, which isn't really an option...

deejross:
I've been following this issue for a while since I ran in to it myself. I'm astonished that this wasn't merged a year ago. It's borderline irresponsible.

bitprophet:
#596 is @GuyShaanan's updated/patched version of this PR, noting it for interlink purposes.

I'm back from vacation now and will be working to merge this & other high profile stuff ASAP. Sincerest regrets on how it's been languishing. Has been a pretty shite year all round :( Thanks for your understanding.

bitprophet:
Testing #596 now...

* Test target: Debian 8 server VM with its stock OpenSSH 6.7
* Noted that by default, this flavor of sshd still supports older MACs in its default `MACs` setting (haven't looked to see if this is specific to Debian tho), so Paramiko master still functions against it.
* Explicitly setting `Ciphers` and `MACs` as suggested in https://github.com/paramiko/paramiko/pull/356#issuecomment-67068052 toggles those older MACs off and triggers the expected `Incompatible ssh server (no acceptable macs)` error.
* Switching to #596 results in happy MAC agreement, but I'm running into `Invalid packet blocking` (which, because of #387 and all its friends, gets logged under DEBUG but manifests at interpreter exit as `Private key file is encrypted`)
    * @GuyShaanan implied earlier that his ordering tweaks cleared up his own instances of the packet blocking error - but clearly not always the case :(
    * Additionally frustrating, the error isn't consistently reproducible, though there is always *an* error. Rarely, I get this instead (though it also bubbles up at the end of key exchange, unlike `Invalid packet blocking`, because it is not an `SSHException`):
    ```
Unknown exception: signed integer is greater than maximum                               
  File "/Users/jforcier/Code/oss/paramiko/paramiko/transport.py", line 1615, in run     
    ptype, m = self.packetizer.read_message()                                           
  File "/Users/jforcier/Code/oss/paramiko/paramiko/packet.py", line 396, in read_message
    buf = self.read_all(packet_size + self.__mac_size_in - len(leftover))               
  File "/Users/jforcier/Code/oss/paramiko/paramiko/packet.py", line 249, in read_all    
    x = self.__socket.recv(n)                                                           
OverflowError: signed integer is greater than maximum                                   
````

I'm now digging deeper to figure out what's going on. Hoping I can get some variant of this merged without having to enter the #387 rabbit hole - I wanted to save that for 2.0 and get this here out for 1.16.

**Comment by bitprophet** [comment]
Poking at the common packet blocking error, it's here: https://github.com/paramiko/paramiko/blob/0b9d772a21a44af38ecceae0fdbae645e386bd9b/paramiko/packet.py#L394-L395 - when I add some debug, I notice the following:

* The modulo mismatch occurs on the very first packet
* The packet and modulo sizes vary every time; I don't know this part of the code well enough to know if that's surprising or not.
* The `OverflowError` occurs when those sizes just happen to line up & thus please the modulo test - it gets execution past the modulo check but falls down farther, unsurprisingly given something is rotten somewhere.

**Comment by bitprophet** [comment]
*  When connecting to a control host which only supports non-SHA-2 MACs, things work fine
* In that scenario, the 'leftover' len is always 12, and the packet size varies but is typically one of 12, 28 or 44.
* In the problem scenario, the packet size is far, far larger (samples: 2202837257, 4035551545, 191273182, 1589669346) but the 'leftover' len is still always 12.
* Given that the packet size comes from deconstructing the header, wondering if that's part of what's broken/misconfigured.

Turned on packet dumps; in the control scenario, the post-newkeys packet which is breaking, looks like this (the lines with the packet repr() and discussion of sizes/modulos, are debug lines I added temporarily):
```
Write packet <service-request>, length 17                               
OUT: 00 00 00 1C 0A 05 00 00 00 0C 73 73 68 2D 75 73    ..........ssh-us
OUT: 65 72 61 75 74 68 00 00 00 00 00 00 00 00 00 00    erauth..........
IN: 00 00 00 1C 0A 06 00 00 00 0C 73 73 68 2D 75 73    ..........ssh-us 
'\x00\x00\x00\x1c\n\x06\x00\x00\x00\x0cssh-us'                          
(packet_size 28 - leftover len 12) % block_size 16 was 0                
IN: 65 72 61 75 74 68 00 57 06 B6 17 34 04 4D 62 91    erauth.W...4.Mb. 
Got payload (28 bytes, 10 padding)                                                                 
```

In the broken scenario, we instead get this:
```
Write packet <service-request>, length 17                               
OUT: 00 00 00 1C 0A 05 00 00 00 0C 73 73 68 2D 75 73    ..........ssh-us
OUT: 65 72 61 75 74 68 00 00 00 00 00 00 00 00 00 00    erauth..........
IN: 75 67 F8 9E EA 0B FC 99 6A A0 5F C2 1D A3 78 35    ug......j._...x5 
'ug\xf8\x9e\xea\x0b\xfc\x99j\xa0_\xc2\x1d\xa3x5'                        
(packet_size 1969748126 - leftover len 12) % block_size 16 was 2        
```

The actual packet under discussion is the `IN` part, the `OUT` is the same in either case anyways. As expected, the data in the failure scenario looks like total garbage, it's clearly not actually a packet header. Question is why.

**Comment by bitprophet** [comment]
Actually thinking the issue may be in the key-exchange parts of the PR(s). I focused on `transport.py` yesterday because the error message was in the handling of packets and it seemed like a good place to start.

But...

* the changes to `transport.py` are pretty basic overall;
* high-level troubleshooting techniques like looking for suspicious diff chunks or examining what the diffs might have *overlooked* and neglected to change, aren't leading anywhere;
* the packets only get garbled after key renegotiation occurs (i.e. after the initial handshaking & agreement on keys wraps up and `MSG_NEWKEYS` is received/logged, as `Switch to new keys...`).
* many of the changes in the PR are in the kex/gex side of things.

So my next steps are to dig into the protocol details further & familiarize myself with the key exchange bits that got updated.

**Comment by bitprophet** [comment]
* Server's also spitting the expected errors about packet length/garbage, clearly the two ends aren't on speaking terms post-newkeys.
    * Meaning either paramiko has the right idea about which cipher or hmac to use but is bungling the implementation, **or** paramiko isn't using the ciper or hmac that it agreed to use.
* In my test cases, `aes128-ctr` is the agreed-upon cipher and `hmac-sha2-256` (vanilla, not etm, as we don't support that yet) is the HMAC.
* The kex algorithm is `diffie-hellman-group14-sha1` in both scenarios, though that shouldn't be relevant anyways - if the issue was about encrypting the key exchange we'd see failures at that step and not after.
    * Thus some of the kex logic I skipped earlier, isn't the culprit, as it implements the `sha256` variant of kex algorithms - which aren't being used at all in my scenarios.
* And the rest of the kex logic that was modified involves stripping out the DH group-oriented algorithms when the local implementation lacks a modulus pack, which also isn't the case here.
* This leaves me with "the code all looks fine, and it apparently works for other people, so what else could be breaking in my situation?"
    * To which the first answer is "the only other code that changes here is the actual hash algorithm - could `hashlib.sha256` be buggy in my local environment?" so I'm gonna head back up a few levels and try Python 2.7 (I'm on 2.6 right now) and Linux (my local OS is Yosemite).
    * After that, I'll continue mutating the allowed-MACs lists on my target sshd, as that's another easily controllable factor which might be differing for other users.

**Comment by bitprophet** [comment]
Python versions sanity check was good to perform but didn't turn anything up (honestly good because it'd be pretty alarming if the problem was "sha256 is broken on Yosemite and/or Python 2.6"):

* No dice on Yosemite's Python 2.7.10, identical behavior in the failure scenario
* Debian 8 has Python 2.7.9 (no 2.6 is available by default), but sadly, also identical failure/success behavior

Checking the other MACs Paramiko supports:

* `hmac-sha2-256`: is what we've been using as the failure case, so obviously it's busted
* `hmac-sha1`: this is the one I've been using to force disuse of sha-2 on Paramiko's end; it works fine
* `hmac-sha1-96`: also fine
* `hmac-md5`: breaks in identical fashion to `hmac-sha2-256`, which is interesting; need to test it with master to see if it was broken prior to this PR, because I wouldn't have expected this to break things...
    * *unless* the changes that switched a hardcoded `sha1` to "the function specified in `Transport._cipher_info` for the algorithm we're using", are overly aggressive somehow?
* `hmac-md5-96`: ditto, breaks

**Comment by bitprophet** [comment]
Confirmed, `hmac-md5` works on master but not on my branch based off of #596. So that's a regression, hopefully a useful one. Poking more.

**Comment by bitprophet** [comment]
Sure enough, if I simply modify `Transport._compute_key` to always use `sha1` again instead of the references to `md5` or `sha256` from `_mac_info`, suddenly all 3 HMAC schemes work correctly, including `hmac-sha2-256`.

This implies that switching the hash algorithm used in that particular step, is the wrong thing to do. What I don't get is why everybody testing had it "work" for them.

Only thing that makes sense off the top of my head is if people were testing on servers which were triggering use of SHA-2 key exchange but **not SHA-2 HMAC** (presuming that the swap-out of `sha1` for e.g. `sha256` in `KexGex` is correct/functional, which I will try to verify shortly). That would mask the bug that I appear to have identified.

**Comment by bitprophet** [comment]
Sanity-tested `KexAlgorithms` setting `diffie-hellman-group-exhange-sha1` (since there's no SHA-2 variant of the default, `diffie-hellman-group14-sha1`), then tested `diffie-hellman-group-exchange-sha256`.

As I half-expected, that broke, until I set `Transport._compute_key` back to using `sha256`. Clearly, we need to be setting this based on the agreed-upon kex, which in retrospect makes a lot of sense given how the protocol works & when `Transport._compute_key` is called.

**Comment by bitprophet** [comment]
FTR using `self.kex_engine.hash_algo` seems like the cleanest way to get at the hash algo set by the `Kex*` classes. (I also had to explicitly add it to some that lacked, since there's not much actual inheritance in that group of classes. Something to address later, perhaps.)

I need to see how easy it is to add tests for some of these kex/hmac combinations so there's at least some regression protection for this kind of silly mistake.

**Comment by bitprophet** [comment]
Grump, the way tests are currently broken down that'll be a lot of effort so I'm gonna punt for now :( (The tests, overall, need a ton of cleaning up anyways...) At least I identified a test failure my change triggered.

----

*So*. I'm now actually happy with my copy of this PR. It has been merged to master just recently. It passes all tests including those ran by Travis. It works for me with varying combinations of SHA1 vs SHA2 based kex *and/or* MAC settings on a target sshd.

I'd like to release it publicly as 1.16, but before I do so, I'd like at least 1-2 of the folks following this ticket to test it on their end, *preferably also with multiple kex and/or MAC settings in play*, please!

Going to cc some of the more recently active players in case their notifications for the ticket itself are disabled, please don't be offended if you're not in this list :) /cc @GuyShaanan @aaronbieber @cyphase @zamiam69

**Comment by cyphase** [comment]
I came just to check up on your progress - I've been following for the last few days - and was pleased to see that you solved this; thanks for pinging me though :). The particular SSH server I was having an issue with was replaced three weeks ago (which caused the program in question to raise the `paramiko.ssh_exception.SSHException: Invalid packet blocking` and `OverflowError: signed integer is greater than maximum` errors until I went back to vanilla 1.15.2), but I'm glad that this has been resolved. Thanks for your hard work @bitprophet :D.

**Comment by bitprophet** [comment]
Thanks for checking in, @cyphase :)

Realistically, I'm going to poke at the rest of [the 1.16 milestone](https://github.com/paramiko/paramiko/milestones/1.16) in the very near term, and that window of time is what I'll leave open for folks to yea/nay my copy of this functionality.

Worst case, if there are remaining bugs, they'll come to light relatively soon & get fixed in bugfix releases.

**Comment by GuyShaanan** [comment]
Hi @bitprophet 
Thank you so much for pinging me.

I just ran my tests again, now using v1.16, and it passed with success on all of my servers!
Thanks again!

**Comment by pixelchutes** [comment]
@bitprophet So far so good for me.

Tested against MySQL Workbench (v6.3.5) and current master: 935711b5a17370494a7b2b8b4587f5466badf1e8

Related https://bugs.mysql.com/74658

**Comment by bitprophet** [comment]
Thanks @GuyShaanan and @pixelchutes!

I released 1.16 today, so it's in the wild

**Comment by rata** [comment]
@bitprophet using it via Fabric, on a host that has:

```
KexAlgorithms curve25519-sha256@libssh.org
```

Can't connect with:
```
Fatal error: Incompatible ssh peer (no acceptable kex algorithm)
```

If I remove the KexAlgorithms directive from the host, Fabric/paramiko can connect just fine. And, of course, I can always ssh from my workstation when using the KexAlgorithms directive.

**Comment by pixelchutes** [comment]
@bitprophet Unfortunately, I can confirm same behavior as @rata when strictly using:
```
KexAlgorithms curve25519-sha256@libssh.org
```

bitprophet:
@rata @pixelchutes That's correct, this PR doesn't include `curve25519-sha256` (I listed the specifics that were added [in the changelog](http://www.paramiko.org/changelog.html#1.16.0)) but I'd merge one that did :)

bitprophet:
See also #325 (which is an issue, not a PR, but.)

rata:
@bitprophet oh, sorry. I thought it will be supported. Security wise, is kind of nice or a must to have it. Any plans?

bitprophet:
@rata I don't have the personal need (or the time) to implement it myself but I certainly recognize the utility of adding it, which is why I said I'd accept a PR :) For whichever reason, this original PR didn't include `curve22519`.

rata:
Fair enough. When you have the time, though, it will be nice as it's the
recommended to use so far :)

PR #368, opened 2014-07-28
Use context manager for `threading.Lock`

InPermutation:
It cleans up the code significantly in a couple places, and insignificantly in a lot of places.


paramiko/channel.py
```diff
@@ -726,8 +713,6 @@ def send_stderr(self, s):
             m.add_int(self.remote_chanid)
             m.add_int(1)
             m.add_string(s[:size])
-        finally:
```
InPermutation: f

PR #401, opened 2014-09-22
Fix line number for known_hosts error messages.

dylanahsmith:
Line number for files conventionally start with 1, but using `for lineno, line in enumerate(f):` would consider the first line to be line 0.

Passing an initial value of 1 to `enumerate` will fix this line number.

### DIFF
--- a/paramiko/hostkeys.py
+++ b/paramiko/hostkeys.py
@@ -92,7 +92,7 @@ class HostKeys (MutableMapping):
         :raises IOError: if there was an error reading the file
         """
         with open(filename, 'r') as f:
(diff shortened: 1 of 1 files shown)


bitprophet:
Took me a while to realize that this only impacts *log lines* and not actual exceptions. :weary: One can reproduce the incorrect-line-number messages by enabling logging, mangling a `known_hosts` file so it includes lines like e.g. `foo` (vs `foo ssh-rsa <key text>`) and then calling e.g. `SSHClient.load_system_host_keys`.

bitprophet:
Also, no idea why I punted this to 1.16.

PR #406, opened 2014-09-30
pass exc_info along when logging exs, dont split and log tb separately

### LINKED ISSUE #405: can transport log not split up error tracebacks?
when transport catches exceptions it splits tb into multiple strings:

https://github.com/paramiko/paramiko/blob/1af8db2331ec9d5f729ff62d8af87b9d8c1c7fb8/paramiko/transport.py#L1633-L1634

which are then logged separately:

https://github.com/paramiko/paramiko/blob/1af8db2331ec9d5f729ff62d8af87b9d8c1c7fb8/paramiko/transport.py#L1423-L1428

making it hard to e.g. filter a specific "error" in server mode.

would it make sense to just pass along whatever exc_info is being handled to logging and let it do formatting as a single msg?

bninja:
ends up looking like this now as single log msg:

```
2014-09-30 07:03:06,891 : ERROR : paramiko.transport : 16813 : Thread-2 : Exception: Error reading SSH protocol banner
  File "/usr/local/lib/python2.7/dist-packages/paramiko/transport.py", line 1582, in run
    self._check_banner()
  File "/usr/local/lib/python2.7/dist-packages/paramiko/transport.py", line 1707, in _check_banner
    raise SSHException('Error reading SSH protocol banner' + str(e))
SSHException: Error reading SSH protocol banner
```

closes #405

### DIFF
--- a/paramiko/server.py
+++ b/paramiko/server.py
@@ -69,7 +69,7 @@ class ServerInterface (object):
             - ``OPEN_FAILED_CONNECT_FAILED``
             - ``OPEN_FAILED_UNKNOWN_CHANNEL_TYPE``
             - ``OPEN_FAILED_RESOURCE_SHORTAGE``
-        
+
         The default implementation always returns
         ``OPEN_FAILED_ADMINISTRATIVELY_PROHIBITED``.
 
@@ -160,7 +160,7 @@ class ServerInterface (object):
         Note that you don't have to actually verify any key signtature here.
         If you're willing to accept the key, Paramiko will do the work of
         verifying the client's signature.
-        
+
         The default implementation always returns `.AUTH_FAILED`.
 
         :param str username: the username of the authenticating client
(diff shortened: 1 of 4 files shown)


exhuma:
Small improvment on the PR: You can simply use `exc_info=True` instead of `exc_info=sys.exc_info()`.

Also, is there any update on this? We've been struggling with this on our ELK stack as the splitting causes log-entries for every single line in the traceback. But what's worse, it make log entries not searchable. I was writing up a  bug report but did not see that there was already #405 until I finished writing it and GitHub pointing out the similarity to me ;)

I had written down a detail on it, and even though it is redundant with everything said in #405 I will copy/paste it below anyway, because it contains screenshots from how this affects ELK:

----

In some exceptions, the traceback is split into lines and each line is logged separately. For example:

* https://github.com/paramiko/paramiko/blob/ca304fef5ba3543323c294e5c85dec87bd17280e/paramiko/transport.py#L2130
* https://github.com/paramiko/paramiko/blob/ca304fef5ba3543323c294e5c85dec87bd17280e/paramiko/transport.py#L2147

The lines pass the traceback to [util.tb_strings](https://github.com/paramiko/paramiko/blob/ca304fef5ba3543323c294e5c85dec87bd17280e/paramiko/util.py#L138) which has a return-type of `List[str]`. That return-value is passed to [Transport._log](https://github.com/paramiko/paramiko/blob/ca304fef5ba3543323c294e5c85dec87bd17280e/paramiko/transport.py#L1814) which then logs each element of that list as a separate log line.

Fragmenting the traceback-message like this makes it almost impossible to search for log messages in a solution like an ELK stack. It also artificially inflates the count of error-messages considerably (which we monitor). See the following screenshot for example:

image

With messages like these, I cannot apply any useful filtering because if I do, I get only a fragment of the traceback. If I add a filter for `SSHConnectionError`, I get the following search result which is only one line of the traceback, and not having the full traceback is really problematic.

image


A better solution would be to use `exc_info=True` (or `exc_info=the_exception_instance`) in the log-statements. This will properly log the exception traceback as one field to an ELK stack, *and* also log it to the console, as mentioned in [logging.Formatter.format](https://docs.python.org/3/library/logging.html#logging.Formatter.format):

> [...] If there is exception information, it is formatted using formatException() and appended to the message. [...]

A quick "grep" around the code-base also shows that `util.tb_strings` is *only* used in logging of exceptions, so making use of `exc_info=True` in the log-statements would make this function redundant.

Issue #412, opened 2014-10-02
2-Factor auth does not work


owais:
2-Factor auth when using SSH + google authenticator lib does not work with paramiko. Password logins are disabled.

```
DEBUG:paramiko.transport:Switch to new keys ...
DEBUG:paramiko.transport:Adding ssh-rsa host key for server.com: #############
DEBUG:paramiko.transport:Trying SSH agent key *********************************************
DEBUG:paramiko.transport:userauth is OK
INFO:paramiko.transport:Authentication (publickey) failed.
DEBUG:paramiko.transport:Trying SSH agent key *********************************************
DEBUG:paramiko.transport:userauth is OK
INFO:paramiko.transport:Authentication (publickey) failed.
DEBUG:paramiko.transport:Trying SSH agent key *********************************************
DEBUG:paramiko.transport:userauth is OK
INFO:paramiko.transport:Authentication (publickey) failed.
DEBUG:paramiko.transport:Trying SSH agent key *********************************************
DEBUG:paramiko.transport:userauth is OK
INFO:paramiko.transport:Authentication continues...
DEBUG:paramiko.transport:Methods: [u'keyboard-interactive']
DEBUG:paramiko.transport:[chan 1] Max packet in: 32768 bytes
WARNING:paramiko.transport:Oops, unhandled type 3
```

`keyboard-interactive` is enabled on the server in addition to secret key auth but paramiko does not ask for authorization code

PaulFurtado:
@perryjrandall  Your changes do work for me when using duo for 2 factor authentication at work and MySQL Workbench's SSH tunnel feature to connect to MySQL servers. :beer:

What are the changes we can make Perry's changes production ready and merge them? I'd be willing to do some of the work if necessary.

bitprophet:
Rolling this into #467

Issue #387, opened 2014-09-06
Key handling is terribad


bitprophet:
There are a number of issues stemming from what feels like a simply too-naïve implementation of key handling in `SSHClient` and `PKey`.

I am making a single ticket for this because most of the existing PRs poking at it are too limited in scope; this sort of change has a high chance for bugs and breaking backwards compatibility (intentionally or no) and I feel it needs a broadly considered update.

[**EDIT**: See [this comment](https://github.com/paramiko/paramiko/issues/387#issuecomment-1492614554) for an update on the intent/scope of this issue!]

----

The high level details:
- First off, `SSHClient._auth` uses a multi-exit strategy combined with storing a single exception to raise at the end of the process. This frequently means the raised exception at auth time is flat out incorrect as to the true cause of the inability to authenticate.
  - For example, early on it attempts to check over all explicitly given `key_filename` values...and since it can't reliably know offhand which key types they are, it tries all known key types. In order. This will _always_ generate/store errors unless one happens to have only one key and it's of the arbitrarily-first-in-the-list key type.
  - Thus even if all keys given were valid keys of their intended type, and they also all happened to be invalid from the target server's perspective (i.e. the user doesn't actually have an accepted key) we raise to them a nice helpful "Hey your key is fucked up" error instead of "None of your keys were accepted by the server".
- Furthermore, `PKey._read_private_key` - used frequently within the above, indirectly - is similarly messy, and many, many types of key loading/reading problems end up raising an error about how `Private key file is encrypted` - on unencrypted files.
  - Why? Simply because the logic assumes if it gets to the end of the function, and the `password` kwarg is blank, that a password was required and not given.
  - While I haven't debugged this as deeply as I'd like yet, it comes up frequently when trying to deal with ECDSA keys, and/or when lower level key handling (or key related crap in the test suite) is presently botched.

A number of tickets have sprung up on this, I'm sure I haven't gotten all of them, but they include #358 and #224.

Strongly related are a bevvy of tickets concerned with exception handling & exception types/subclasses, such as [Fabric #85](https://github.com/fabric/fabric/issues/85) and our own #351 and #293.

Russell-Jones:
One approach to starting to debug it: 

1. change saved_exceptions to an array, i.e. 
```
saved_exceptions = []
...
saved_exceptions.append(e)
```
2. Do something like [this](https://mail.python.org/pipermail/python-list/2010-May/588272.html) so you get tracebacks for all exceptions.

[...]

PR #467, opened 2014-12-30
Add more flexible support for two factor authentication.

perryjrandall:
Allow paramiko to partially authenticate and continue by
echo'ing the prompt on the remote end and responding.

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -404,7 +404,8 @@ class SSHClient (ClosingContextManager):
         """
         saved_exception = None
         two_factor = False
-        allowed_types = []
+        allowed_types = set()
+        two_factor_types = set(['keyboard-interactive','password'])
 
         # If GSS-API support and GSS-PI Key Exchange was performed, we attempt
         # authentication with gssapi-keyex.
@@ -430,8 +431,8 @@ class SSHClient (ClosingContextManager):
         if pkey is not None:
             try:
                 self._log(DEBUG, 'Trying SSH key %s' % hexlify(pkey.get_fingerprint()))
-                allowed_types = self._transport.auth_publickey(username, pkey)
-                two_factor = (allowed_types == ['password'])
+                allowed_types = set(self._transport.auth_publickey(username, pkey))
+                two_factor = (allowed_types & two_factor_types)
                 if not two_factor:
                     return
             except SSHException as e:
@@ -444,7 +445,7 @@ class SSHClient (ClosingContextManager):
                         key = pkey_class.from_private_key_file(key_filename, password)
                         self._log(DEBUG, 'Trying key %s from %s' % (hexlify(key.get_fingerprint()), key_filename))
                         self._transport.auth_publickey(username, key)
-                        two_factor = (allowed_types == ['password'])
+                        two_factor = (allowed_types & two_factor_types)
                         if not two_factor:
                             return
                         break
@@ -458,9 +459,9 @@ class SSHClient (ClosingContextManager):
             for key in self._agent.get_keys():
                 try:
                     self._log(DEBUG, 'Trying SSH agent key %s' % hexlify(key.get_fingerprint()))
-                    # for 2-factor auth a successfully auth'd key will result in ['password']
-                    allowed_types = self._transport.auth_publickey(username, key)
-                    two_factor = (allowed_types == ['password'])
+                    # for 2-factor auth a successfully auth'd key password will return an allowed 2fac auth method
+                    allowed_types = set(self._transport.auth_publickey(username, key))
+                    two_factor = (allowed_types & two_factor_types)
                     if not two_factor:
                         return
                     break
@@ -497,8 +498,8 @@ class SSHClient (ClosingContextManager):
                     key = pkey_class.from_private_key_file(filename, password)
                     self._log(DEBUG, 'Trying discovered key %s in %s' % (hexlify(key.get_fingerprint()), filename))
                     # for 2-factor auth a successfully auth'd key will result in ['password']
-                    allowed_types = self._transport.auth_publickey(username, key)
-                    two_factor = (allowed_types == ['password'])
+                    allowed_types = set(self._transport.auth_publickey(username, key))
+                    two_factor = (allowed_types & two_factor_types)
                     if not two_factor:
                         return
                     break
@@ -512,7 +513,11 @@ class SSHClient (ClosingContextManager):
             except SSHException as e:
                 saved_exception = e
         elif two_factor:
-            raise SSHException('Two-factor authentication requires a password')
+            try:
+                self._transport.auth_interactive_dumb(username)
+                return
+            except SSHException as e:
+                saved_exception = e
 
         # if we got an auth-failed exception earlier, re-raise it
         if saved_exception is not None:
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -20,6 +20,7 @@
 Core protocol implementation
 """
 
+from __future__ import print_function
 import os
 import socket
 import sys
@@ -1283,6 +1284,27 @@ class Transport (threading.Thread, ClosingContextManager):
         self.auth_handler = AuthHandler(self)
         self.auth_handler.auth_interactive(username, handler, my_event, submethods)
         return self.auth_handler.wait_for_response(my_event)
+
+    def auth_interactive_dumb(self, username, handler=None, submethods=''):
+        """
+        Autenticate to the server interactively but dumber.
+        Just print the prompt and / or instructions to stdout and send back
+        the response. This is good for situations where partial auth is
+        achieved by key and then the user has to enter a 2fac token.
+        """
+
+        if not handler:
+            def handler(title, instructions, prompt_list):
+                answers = []
+                if title:
+                    print(title.strip())
+                if instructions:
+                    print(instructions.strip())
+                for prompt,show_input in prompt_list:
+                    print(prompt.strip(),end=' ')
+                    answers.append(raw_input())
+                return answers
+        return self.auth_interactive(username, handler, submethods)
 
     def auth_gssapi_with_mic(self, username, gss_host, gss_deleg_creds):
         """


perryjrandall:
Bumping :) I know it's still kinda the holidays but I'd greatly appreciate this getting pulled in! Thanks

perryjrandall:
Hey @bitprophet what specifically do you want to see in the way of docs / changelog?

perryjrandall:
Bumping again, hey @bitprophet I'd love to get this committed, what needs to be done?

chantra:
Is there anything we can help with to get this into master?

bitprophet:
Nope, just need time to check it over and test it on my end (just dug out of some start-of-year dayjob responsibilities, entering a pre-pycon open source sprint phase).

For now I am assuming the notes in #412 about reproducing/verifying, will apply to this implementation so that's what I'll do first.

cbaenziger:
For reference, I was able to use `client.get_transport().auth_interactive_dumb(username=user)` successfully for a proprietary two-factor authentication system where a simple `client.connect(host, user, static_password)` failed with:
````
paramiko.ssh_exception.BadAuthenticationType: ('Bad authentication type', [u'keyboard-interactive']) (allowed_types=[u'keyboard-interactive'])
````
However, I am new to Paramiko, so I do not know what could have been silently incorrect or wonky. Certainly all my password-y input was echoed back for better or worse (and I was just running in the Python REPL).

Either way, thanks for the awesome work!

**Comment by cbaenziger** [comment]
My apologies I may be too naive to be of use. I also used Paramiko version `1.10` as provided with Ubuntu 12.04 and ran the following to the same two-factor authentication system successfully:
````
from getpass import getpass
def my_handler(title, instructions, prompt_list):
   print title
   print instructions
   return [echo and raw_input(prompt) or getpass(prompt) for (prompt, echo) in prompt_list]

try:
  client.connect(host, username=user)
except paramiko.ssh_exception.SSHException:
  pass

client.get_transport().auth_interactive(username=user, handler=my_handler)
````
Also, just to correct my last post the handler is required to not echo the password; nothing wrong with `auth_interactive_dumb`.

**Comment by adimania** [comment]
bumping this. The checks have passed. Can we merge this sometime soon please?

**Comment by mattrobenolt** [comment]
I tested this patch out, and it _almost_ works.

Requires this extra patch to work in our case:

```diff
diff --git a/paramiko/client.py b/paramiko/client.py
index 0c8f6bd..2ecd487 100644
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -503,7 +503,7 @@ def _auth(self, username, password, pkey, key_filenames, allow_agent,
                     try:
                         key = pkey_class.from_private_key_file(key_filename, password)
                         self._log(DEBUG, 'Trying key %s from %s' % (hexlify(key.get_fingerprint()), key_filename))
-                        self._transport.auth_publickey(username, key)
+                        allowed_types = set(self._transport.auth_publickey(username, key))
                         two_factor = (allowed_types & two_factor_types)
                         if not two_factor:
                             return
```

https://github.com/mattrobenolt/paramiko/commit/75b64b80f7184a30609c65a35c1c80ca962d76dc

bitprophet:
Heyyyyy #467 what's good~

Prep work:

* Dropping a link to #387 so that when I handle that crap, I remember that this is a thing & needs to not break during the rewrite (but hopefully that will be soon...)
* I followed the tips from #412 to get google-based 2FA working via OpenSSH's client targeting a VM.
    * Both as the sole login method and as a literal "second" auth factor on top of password auth, depending on whether or not I `@include common-auth` in `/etc/pam.d/sshd`.
        * Which made me realize there's probably 2 related-but-distinct things to support here, a) support for multiple simultaneous authentication methods, and b) support for interactive non-password OTP type schemes, singularly or as a second factor.
    * I have to forcibly disable key auth on my client's end, since it seems to obviate any use of the  Google Authenticator plugin. Might dig more if I care about PAM.
    * The prompts OpenSSH client provides are useful (duh): says `Password:` for system password, `Verification code:` for google-auth.
* Confirmed that hitting the VM with Paramiko master (via Fabric) with PAM configured to only use google-auth, works, since it's basically just a different password mechanism.
    * However, since Fabric has to write the prompt itself, it always says 'password' not 'verification code'.
* Confirmed that once you tell PAM to want both password *and* google-auth, things don't work (only one prompt is presented and then failure occurs).

Trying to get the PR to function with the above setup fails with `AuthenticationException` after submitting a password, however, I suspect because the author & those testing it like @mattrobenolt have all convinced their sshd to do pubkey+google, whereas I am trying to test password+google. So I have to:

* Figure out what's missing from the PR to get password+google working (having looked, I think not super hard) or identify what else is causing my failure;
* Get pubkey+google working on my test target, then test that angle.

mattrobenolt:
>  I suspect because the author & those testing it like @mattrobenolt have all convinced their sshd to do pubkey+google, whereas I am trying to test password+google

We're specifically using pubkey + [Duo Unix](https://www.duosecurity.com/docs/duounix) with their PAM module.

And specifically using `AuthenticationMethods publickey,keyboard-interactive` in our `sshd_config` (following Duo's directions verbatim).

I'm happy to share any other information if you need it.

mattrobenolt:
Also, for more context on our use case if it's relevant:

We use Fabric to command our deploys, which happen through some automated magic. This means the 2FA needs to be bypassed in our case, which is a thing that Duo supports. Our `build` use doesn't need to prompt for the second auth and Duo just noops their case.

Not sure if this changes the behavior here, but for both of the auth cases, there are no prompts. `keyboard-interactive` just ends up passing through without needing any input since the PAM module allows it.

bitprophet:
Notes to self...

* https://gist.githubusercontent.com/bitprophet/b251973124e91f1e63db/raw/c0eccc407721ebca8c5249c8565dd3e5149b79eb/openssh-client.txt is what my sshd log looks like on a successful password+google login
* https://gist.githubusercontent.com/bitprophet/9f1e6b1dd5029aec7152/raw/cdb52d956fa3669eb365c8ec29d379b68c621c4c/gistfile1.txt is a failed attempt via Paramiko/Fabric (shorter because no cxn teardown occurs here, Fab was hanging out trying to re-prompt as if the password was bad, which isn't relevant right now)

Notable difference is that the userauth request method is `keyboard-interactive` on openssh's part but is `password` on Paramiko/Fabric's part. I assume this is why the failure scenario doesn't appear to respond as expected re: password, then auth code.

EDIT: yea, a bit of the ol' Google shows that `keyboard-interactive` basically supersedes `password`, and maps to telling OpenSSH's sshd `ChallengeResponseAuthentication yes` (which was required for the google plugin, and presumably any interesting PAM stuff, to function).

So when Paramiko's (horrible, terrible, no good, very bad) `Client._auth` falls down to the end and calls `self.transport.auth_password`, this triggers code in `AuthHandler` that explicitly submits a `password`-method auth request (as we saw in the logs).

That's *supposed* to trigger a `BadAuthenticationType` exception internally in this kind of scenario and fall back to `auth_interactive`. But this isn't happening in my case, probably because my sshd is actually supporting `password` so it's a different failure scenario from the server's perspective.

I think this can be considered a very niche case and I'm not likely to try fixing it right now, if I can get pubkey based auth working instead. Will try to confirm by telling the server to stop supporting `password`.

bitprophet:
@mattrobenolt thanks for the deets (& the missed-spot patch you commented earlier), that (so far) seems to map to what I'm seeing. I'm guessing your Duo Unix setup behaves closely enough like Google Auth that it won't matter on my end. We'll see...

bitprophet:
* Confirmed that when I explicitly state `PasswordAuthentication no` on the sshd, openssh client is still happy in both 2FA and "just password" situations, *and* Fabric/Paramiko are still happy in the "just password" situation, proving that said fallback code works / is happening (also verified this in paramiko's log).
* Paramiko still being a jerk in the "password + 2FA" scenario, trying to figure out why exactly - even when I update the logic at the bottom of `Client._auth` to not assume password and 2FA are mutually exclusive, the initial password step is failing during the fallback to `keyboard-interactive`.
    * Log from server: https://gist.githubusercontent.com/bitprophet/b9add39087050137f3ea/raw/c40766139d9258b23d1a4b17c38c82cdbbc34eed/gistfile1.txt
    * Need to compare that to the success case, but wondering if somehow the server is just being obstinate because of the initial failed `password` attempt, and if cutting that out would make a difference.
    * If testing this case takes much longer, will just fall back to testing key+2FA, because it's super unlikely anybody conscious enough to enable 2FA is going to be pairing it with password instead of privkey...

bitprophet:
Aside from the failure log saying `attempts 1 failures 1` (again, assuming that is because of the first `password` attempt) vs a success saying `attempts 1 failures 0`, the two logs look the same until the failure log is all `sshd(pam_google_authenticator)[2101]: Invalid verification code`.

Probably need to peep the packets being sent by openssh client vs paramiko cuz that still doesn't make sense to me, unless sshd is literally treating attempt 1 as the passwordy login and attempt 2 as the google-auth login.

I'm gonna tinker real quick to try and short circuit around the `password` login attempt to see if that's it, and if not, will skip ahead to key auth because ugh time.

bitprophet:
Aha, some more reading/logging shows that `auth_interactive` is being fallen back to via a handler callback, which yields the password that was to be used for the `password`-type message. (I.e. it's `auth_password` trying to non-interactively execute `auth_interactive`.)

So what's happening is this handler is called *twice*, once for each prompt, which is why the google-auth pam plugin is all mad - it really is being given input, and the wrong input.

This is apparently the documented behavior of `Transport.auth_interactive`, to wit, `The handler may be called more than once if the server continues to ask questions`. Specifically it's implemented inside `AuthHandler` as "store the handler, respond with it when you get a new `INFO` request of the expected type", and clearly the client gets two of those in a row from the server.

This was unexpected in part because all these auth handlers are implemented/documented as "in the rare case of multi-step auth, this returns other acceptable auth types" - but this specific "password + 2FA" setup is *not actually multi-step auth in that sense* because it's two `keyboard-interactive` "questions" instead of e.g. one privkey and one keyboard-interactive.

tl;dr this really is the fault of me trying to get it working without a key, so I am definitely doing that next and putting this aside for now.

bitprophet:
FTR I needed to:

* Explicitly configure `AuthenticationMethods` in sshd config (to `pubkey,keyboard-interactive`, though there's no actual default apparently, just "any one thing will do", thus the behavior I was seeing)
* Remove inclusion of PAM's general account module in the `sshd` module, since otherwise I need to give my pubkey, the 2FA token *and* the system user's password, which is a bit much.
    * Though post-#387 we'll hopefully be able to handle that as well...it's not worth futzing with before then IMHO.

This gets me a working pubkey+2FA setup which presumably matches the other folks' configurations.

With master Paramiko, things just hang after logging `WARNING:paramiko.transport:Oops, unhandled type 3`.

With my branch based on this PR, I am successfully prompted for the verification code, enter it, and successfully connect. Notes:

* No changes needed on Fabric's end.
* The prompt echoes my input, unlike vanilla OpenSSH which hides my input.
    * This is non-ideal, but not a deal-breaker given the nature of 2FA tokens. Suspect it too can wait until after #387 rearchitects auth handling.

Think this means we can merge and move on, finally.

perryjrandall:
Woo hoo! Glad to see this merged, thanks @bitprophet

Symmetric:
It wasn't immediately obvious that this would be the case from skimming this thread, but this also works with sftp servers using the slightly odd configuration

```
AuthorizationMethods publickey,password
```

:+1:

etmoore:
>but this also works with sftp servers using the slightly odd configuration
>AuthorizationMethods publickey,password

@Symmetric Does this mean that #519 should be closed?

PR #470, opened 2015-01-06
Added dead server detection code.

tigran-avanesov:
This change is required to detect the situations when SSH server dies. This code checks for the latest message receive time and compares it with the current timestamp almost as it is done in OpenSSH.
Also added none authentication support in order to connect to SSH without password at all. This function exists in current implementataion, but it was not used.

### DIFF
--- a/paramiko/_version.py
+++ b/paramiko/_version.py
@@ -1,2 +1,2 @@
-__version_info__ = (1, 15, 2)
+__version_info__ = (1, 15, 3)
 __version__ = '.'.join(map(str, __version_info__))
--- a/paramiko/_version.py
+++ b/paramiko/_version.py
@@ -1,2 +1,2 @@
-__version_info__ = (1, 15, 3)
+__version_info__ = (1, 15, 4)
 __version__ = '.'.join(map(str, __version_info__))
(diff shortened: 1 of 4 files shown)


bitprophet:
Can you clarify this? What exactly is this fixing & how can we reproduce it to prove it works? Thanks!

tigran-avanesov:
This change is done to detect the situation when SSH server dies. This code checks for the latest message receive time and compares it with the current timestamp. 
This situation can be easily reproduced if you connect to virtual machine and pause it when executing some commands.
In older version of paramiko connection will hang forever, while with this change it will be closed. This is critical for our project and causes many problems, so we did this fix.

MrMino:
Are there any plans to at least remedy the `auth_none` situation with this PR? I tried to implement it in #1399, but failed miserably.

IMHO the behavior of `SSHClient` and `Transport` should partially follow the OpenSSH implementation, as in: try method-none first, if it doesn't work then at least log the possible authentication methods.

tigranavanesov:
@MrMino I don't plan to work on this anymore. Actually this was required for one of our projects in past and we fixed it locally. Now I don't even remember what was the problem :)

MrMino:
Yes, I understand, but this is the only PR that I've found so far that tries to add `auth_none()` to `SSHClient`. My comment was made in the hope of getting an answer on this from @bitprophet :).

PR #471, opened 2015-01-06
SFTP: Add support for "posix-rename@openssh.com" extension for SFTP client.

mikapfl:
Hi,

this is the same feature as https://github.com/paramiko/paramiko/pull/65, but with docstrings and a test.

Note that at the moment, the test fails early if run against the sftp stub server, as the stub server actually uses os.rename() to implement the sftp rename, which is wrong as per the sftp spec. However, the test works if run against openssh.

Cheers,

Mika


mikapfl:
Hi,

I guess I can add the SFTP server code for the stub server and the interfaces, and also make the stub server pass the non-posix-rename-test such that the tests all pass.
For the changelog/docs I'm not sure what to do.

Cheers,

Mika

bitprophet:
Changelog's in `sites/www/changelog.rst` if you want to take a crack at it, just add new entry at top following style of the others. Otherwise I'll add one when I merge. Thanks!

jlp78:
Why has this not been merged?

PR #473, opened 2015-01-08
Take known_hosts into account when sending the preferrerd key order.

mbr:
A common problem when using parameter on systems like Ubuntu or Debian
is that the default host key request by the regular SSH client is
ecdsa-sha2-nistp256, while Paramiko will ask for ssh-rsa first. This will
cause a miss upon the first key lookup and will trigger a "host is not
in known_hosts" warning, even though it's present.

This patch will check if the host is known first and alter the list of
preferred host key types based upon that information before connecting.

I'm not 100% sure everything is in the right place, since I spent most of my time actually tracking down thte issue. However I believe that the current state is a catatrophe! While in theory this should cause affected people to cautiously file bug reports, this is what reality looks like: https://stackoverflow.com/questions/10670217/paramiko-unknown-server - cannot get it working? oh well, better disable host key checking (which is conveniently named "auto add" instead).

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -252,6 +252,26 @@ class SSHClient (ClosingContextManager):
 
         t = self._transport = Transport(sock, gss_kex=gss_kex, gss_deleg_creds=gss_deleg_creds)
         t.use_compression(compress=compress)
+
+        if port == SSH_PORT:
+            server_hostkey_name = hostname
+        else:
+            server_hostkey_name = "[%s]:%d" % (hostname, port)
+
+        # if we already have a host key stored, change our key preference
+        known_host_keys = self._system_host_keys.get(server_hostkey_name,
+                                                     {})
+        if not known_host_keys:
+            self._host_keys.get(server_hostkey_name, {})
+        if known_host_keys:
+            # order the keys as follows: known keys in preferred-keys order,
+            # then unknown keys in preferred-keys order
+            valid_known_keys = [k for k in known_host_keys
+                                if k in t._preferred_keys]
+            t._preferred_keys = (valid_known_keys
+                                 + [k for k in t._preferred_keys
+                                    if k not in valid_known_keys])
+
         if gss_kex and gss_host is None:
             t.set_gss_host(hostname)
         elif gss_kex and gss_host is not None:
@@ -267,11 +287,6 @@ class SSHClient (ClosingContextManager):
 
         server_key = t.get_remote_server_key()
         keytype = server_key.get_name()
-
-        if port == SSH_PORT:
-            server_hostkey_name = hostname
-        else:
-            server_hostkey_name = "[%s]:%d" % (hostname, port)
 
         # If GSS-API Key Exchange is performed we are not required to check the
         # host key, because the host is authenticated via GSS-API / SSPI as


bitprophet:
This sounds very familiar, I don't have time to dig deeply into the rest of the stuff flagged as key-related (https://github.com/paramiko/paramiko/labels/Keys) but suspect it's probably closely related. Adding to the label so I come back and look when I tackle that area (see mostly #387)

mbr:
Thanks. I actually missed that it was somewhat of a known problem. I do hope it's a priority, because as I said above, it seriously compromises the actual security of many systems out there.

If I can do more to help, let me know.

j9ac9k:
Can confirm this is the same issue as #529 ...applying https://github.com/mbr/paramiko/commit/d51683709d32d8126911a6a489eba97005d6268e
https://github.com/mbr/paramiko/commit/b7f10c7b19f6e083f62ff9329511ca7654871155
and https://github.com/mbr/paramiko/commit/b7fbb8def62dd4ac8b99059fd8b57b58dd5cf9ec fixed the issue for me.

finjulhich:
Also confirm what j9ac9k commented on 28 Jun.
Although I only applied the first 2 commits in his comment, as the 3rd one is identical to the 2nd one.

mc3:
Being not compatible with command line ssh tools is a big issue for me:
```
  File "/usr/local/py_venv/p35ca/lib/python3.5/site-packages/paramiko/client.py", line 361, in connect
    server_key)
  File "/usr/local/py_venv/p35ca/lib/python3.5/site-packages/paramiko/client.py", line 648, in missing_host_key
    raise SSHException('Server %r not found in known_hosts' % hostname)
paramiko.ssh_exception.SSHException: Server 'bh3' not found in known_hosts
(p35ca)[root@hermes /usr/local/py_app/CA]# host bh3
bh3.lrau.net has address 91.216.35.170
bh3.lrau.net has IPv6 address 2a02:d40:2:18::70
(p34ca)[root@hermes /usr/local/py_app/CA]# grep bh3 ~/.ssh/known_hosts 
bh3 ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBHy/V8HrE1oMGmZdCSo+6TXFcVuDY6Vx8bCsm2Y8jshNvgE+oOGHb+9ZqltmZrsq7qoG4CMYbB9cNKvRAw6/ufE=
bh3.lrau.net ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBHy/V8HrE1oMGmZdCSo+6TXFcVuDY6Vx8bCsm2Y8jshNvgE+oOGHb+9ZqltmZrsq7qoG4CMYbB9cNKvRAw6/ufE=
(p34ca)[root@hermes /usr/local/py_app/CA]# ssh -V
OpenSSH_6.6.1p1, OpenSSL 1.0.1p-freebsd 9 Jul 2015
```
That missing functionality prevents me from migrating from mbr/paramiko to paramiko/paramiko.

Caligatio:
I just ran into this problem today and it set me back an hour or two.  The proposed patch by @mbr looks like it's going down the right path but doesn't appear to take into account non-system known_hosts.

Issue #353, opened 2014-07-08
OverflowError: Python int too large to convert to C long


fz0dbx:
I am getting the following stack trace on a 32bit Debian jessie computer:

File "/usr/lib/python2.7/dist-packages/paramiko/sftp_client.py", line 155, in listdir
    return [f.filename for f in self.listdir_attr(path)]
  File "/usr/lib/python2.7/dist-packages/paramiko/sftp_client.py", line 193, in listdir_attr
    attr = SFTPAttributes._from_msg(msg, filename, longname)
  File "/usr/lib/python2.7/dist-packages/paramiko/sftp_attr.py", line 91, in _from_msg
    attr._unpack(msg)
  File "/usr/lib/python2.7/dist-packages/paramiko/sftp_attr.py", line 104, in _unpack
    self.st_uid = msg.get_int()
  File "/usr/lib/python2.7/dist-packages/paramiko/message.py", line 140, in get_int
    return util.inflate_long(self.get_binary())
  File "/usr/lib/python2.7/dist-packages/paramiko/message.py", line 198, in get_binary
    return self.get_bytes(self.get_size())
  File "/usr/lib/python2.7/dist-packages/paramiko/message.py", line 108, in get_bytes
    b = self.packet.read(n)
OverflowError: Python int too large to convert to C long

It does not happen on a 64bit machine.

As you can see, all that I am doing is a listdir.
When I run in a debugger and print the value of n, I get 4294967039L
Seems like a bug in Paramiko.

dboreham:
I ran into this issue today, found this bug report and did some investigation. As far as I can see the problem is in some (to me at least) quite odd code in message.py. In particular the function Message.get_int() treats the case where the leading byte is 0xff differently (and to my eye incorrectly). This in turn leads to the > 32-bit value for the uint32 being decoded, causing the Python integer overflow exception. That code path was exercised while decoding the sftp channel open response from the server because the server I connected to sent 0xffffffff as the window size. I coded a quick fix that for me at least makes my sftp work. I don't know enough about the ssh protocol encoding to know if I broke other stuff though. I also noticed in passing that there are two instances of the method get_size() in Message. Surely that's not correct? If someone familiar with the code could let me know if I'm on the right track (or not), I can submit a pr with my fix. Curiously, I worked with Robey many years ago -- had no idea he had written this code!

dboreham:
[...]

PR #475, opened 2015-01-08
Fix get_int()

dboreham:
Fix get_int (make it decode a 4-byte unsigned 32-bit integer) and remove extra copy of get_size.

I came to the conclusion the existing implementation for get_int() was incorrect while investigating this issue : https://github.com/paramiko/paramiko/issues/353 but it is likely also causing other problems as well (e.g. hung sftp sessions) due to the server's channel window size being incorrectly decoded by the client.

This change reverts get_int() to the version prior to this commit: https://github.com/paramiko/paramiko/commit/0e4ce3762a5b25c5d3eb89335495d3bb9054e3e7

Also remove a second redundant (and wrong) copy of the function get_size() that came from the same commit but was subsequently eclipsed by a good version in this merge (the good version is defined second in the file): https://github.com/paramiko/paramiko/commit/e7f41de2f2dac5d03404f35edc5514f12e42c49f

### DIFF
--- a/paramiko/message.py
+++ b/paramiko/message.py
@@ -135,24 +135,7 @@ class Message (object):
 
         :return: a 32-bit unsigned `int`.
         """
-        byte = self.get_bytes(1)
-        if byte == max_byte:
-            return util.inflate_long(self.get_binary())
-        byte += self.get_bytes(3)
-        return struct.unpack('>I', byte)[0]
-
-    def get_size(self):
-        """
-        Fetch an int from the stream.
-
-        @return: a 32-bit unsigned integer.
-        @rtype: int
-        """
-        byte = self.get_bytes(1)
-        if byte == max_byte:
-            return util.inflate_long(self.get_binary())
-        byte += self.get_bytes(3)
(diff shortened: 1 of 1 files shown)


dboreham:
Looking at the test failure a bit, I'm pretty sure that a) the test is invalid (that is : my code is correct and the test is wrong) and b) now I think I see how the get_int() code ended up the way it was. It looks like there was an assumption that you can decide on the encoding of an integer by inspecting its value (either uint32 or mpint). That's just not valid. e.g. if you are given a value larger than 3^32 by the application for a uint32 , that would be an error -- not an indication that you should go ahead and encode it as an mpint. Similarly you might have a protocol value that is defined as an mpint but that has a small integer value. On the decode side, there are valid uint32 values that have a leading byte of 0xff. The existing code appears to take this as a signal to decode the subsequent bytes as an mpint, which leads to the symptoms in the bug report.

ploxiln:
looks like the equivalent fix was done in https://github.com/paramiko/paramiko/pull/482

PR #478, opened 2015-01-21
Fix issue #23, port of PR24

jansegre:
Paramiko was sending multiple `MSG_SERVICE_REQUEST`, though openssh behaviour is to not resend the service request after it was accepted.

Though this may not be a problem for some implementations it is for the Mikrotik ROSSSH, which is the case I'm trying to fix. More details on #23.

Note: the diff is a bit weird because I de-indented a portion of the code so it could live in a separate function, which by the way could have its name reviewed.

Note2: I know, I know, I haven't made any test. Although I believe coverage shouldn't be affected I little time right now and just needed the issue fixed. If tests are really required please ask again that I'll try to take some time to make it.

### DIFF
--- a/paramiko/auth_handler.py
+++ b/paramiko/auth_handler.py
@@ -154,105 +154,12 @@ class AuthHandler (object):
 
     def _request_auth(self):
         if not self.userauth_service_accepted:
+            m = Message()
             m.add_byte(cMSG_SERVICE_REQUEST)
             m.add_string('ssh-userauth')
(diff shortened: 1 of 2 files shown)


bitprophet:
Thanks! Tossing this into the "oh god keys are horrible" bucket and will examine it when I sit down and unravel all that junk. Also thanks for linking it to other apparently-related tickets, appreciated :)

ploxiln:
also related: #165

PR #479, opened 2015-01-23
Raise AuthenticationException instead of TypeError

### LINKED ISSUE #402: Attempting to forward agent w/o actual local agent raises TypeError
E.g. when running [this integration test on Fabric](https://github.com/fabric/fabric/blob/d77ddb9818cfefb176bcdb6af0034ce719af919e/integration/test_operations.py#L133) where I stupidly neglected to set up an actual local agent first, we get:

``` python
Exception in thread Thread-2:
  File "/opt/python/2.6.9/lib/python2.6/threading.py", line 532, in __bootstrap_inner
    self.run()
  File "/home/travis/virtualenv/python2.6.9/src/paramiko/paramiko/agent.py", line 114, in run
    self._communicate()
  File "/home/travis/virtualenv/python2.6.9/src/paramiko/paramiko/agent.py", line 124, in _communicate
    events = select([self._agent._conn, self.__inr], [], [], 0.5)
TypeError: argument must be an int, or have a fileno() method.
```

This is the same as [Fabric #642](https://github.com/fabric/fabric/issues/642).

Presumably we can just check for the value being handed to the `select` call to ensure it's not `None` or whatnot, and raise some more useful exception (or, possibly, just emit a log warning, though since this only happens on explicit attempts to forward an agent, it not functioning should probably be considered a stop-worthy error.)

jordo1ken:
Fixes #402 

Added a bit of logic to test for a Ssh agent and raise an AuthenticationException if unable to connect to one.

### DIFF
--- a/paramiko/agent.py
+++ b/paramiko/agent.py
@@ -32,7 +32,7 @@ from select import select
 from paramiko.common import asbytes, io_sleep
 from paramiko.py3compat import byte_chr
 
-from paramiko.ssh_exception import SSHException
+from paramiko.ssh_exception import SSHException, AuthenticationException
 from paramiko.message import Message
 from paramiko.pkey import PKey
 from paramiko.util import retry_on_signal
@@ -109,9 +109,15 @@ class AgentProxyThread(threading.Thread):
     def run(self):
         try:
             (r, addr) = self.get_connection()
+            # Found that r should be either a socket from the socket library or None
             self.__inr = r
-            self.__addr = addr
+            self.__addr = addr # This should be an IP address as a string? or None
             self._agent.connect()
+            print("Conn Value: ", self._agent._conn)
+            print("Has fileno method: ", hasattr(self._agent._conn, 'fileno'))
+            print("Verify isinstance of int: ", isinstance(self._agent, int))
+            if self._agent._conn is None or not (hasattr(self._agent._conn, 'fileno') or isinstance(self._agent, int)):
+                raise AuthenticationException("Unable to connect to SSH agent")
             self._communicate()
         except:
             #XXX Not sure what to do here ... raise or pass ?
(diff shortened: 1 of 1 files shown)


choochootrain:
:+1:

paramiko/agent.py
```diff
@@ -109,9 +109,12 @@ def __init__(self, agent):
     def run(self):
         try:
             (r, addr) = self.get_connection()
+            # Found that r should be either a socket from the socket library or None
             self.__inr = r
-            self.__addr = addr
+            self.__addr = addr # This should be an IP address as a string? or None
             self._agent.connect()
+            if self._agent._conn is None or not (hasattr(self._agent._conn, 'fileno') or isinstance(self._agent, int)):
```
bitprophet: Probably need `isinstance` first, otherwise if it *is* an int, the first test(s) on this line will fail since `int` types do not have `._conn` attributes :)

I actually don't recall seeing reports of an issue where `self._agent` is an int - where did this come from?

jordo1ken: @bitprophet I was trying to find my original traceback message, but the original from #402 is close enough. The very last line mentions that it's looking for it be an ```int``` object or have the method. So I added in the either/or test because both of those tests need to be ```False``` in order to avoid the traceback apparently. I forget why I put the check for the ```None``` type first. 

Anyway, if you still want me to change it I will.

jordo1ken: @bitprophet Apologies for not understanding what you were saying sooner. I have seen the error of ways and I believe corrected them.

bitprophet: Yea that looks better, thanks!

bitprophet:
Confirmed that I can recreate the original problem (via fabric) and that this appears to work OK :) Thanks!

bitprophet:
Rebased onto 1.15 branch

PR #394, opened 2014-09-15
Switched everything to use cryptography instead of pyCrypto

alex:
Motivation:

* Adds PyPy support
* Performance improvement
* OpenSSL and friends are better audited than PyCrypto
* Easier windows install flow (Cryptography provides statically linked wheels
  on Windows)

This PR is basically complete on the code side, of course it can always use
more review :-)

Tests all pass locally (tested with PyPy!)

Still needs some docs work, and to figure out how to do this with the version
numbers so people's stuff doesn't suddenly get broken.

PR #486, opened 2015-02-09
Bump ecdsa: 0.11 -> 0.13

frewsxcv:
https://github.com/warner/python-ecdsa/blob/5a6fc047222cf21ad89f6cbf8782d0f1e3ddacda/NEWS#L1-L12

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ try:
     kw = {
         'install_requires': [
             'pycrypto >= 2.1, != 2.4',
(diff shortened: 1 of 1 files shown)


bitprophet:
Thanks! We're actually slated to drop ecdsa in 2.0 (in favor of the `cryptography` library - see #394) which is also going to be the next minor-or-major release candidate. So I'm going to close this for now, but if plans change and we put out an e.g. 1.16 sometime, it might get reconsidered.

PR #487, opened 2015-02-11
Fix non-blocking reads from ChannelFile

pheller:
`read(-1)` on a `ChannelFile` presently blocks even when that channel has been set not to block with an appropriate call to `setblocking()` or `settimeout()`.  This patch restores the expected operation of returning to the caller upon the expiration of a timeout in the underlying `BufferedPipe` `read()` method.


pheller:
Hmmh, only the python 3.2 build failed, which oddly did not fail on a build against this branch on my personal fork.  Will investigate.

PR #488, opened 2015-02-17
Typo causes failure if Putty Pageant is running.

stevecohen42:
Pageant connections fail due to a typo in the setter for the security agent descriptor.  Fixing the typo takes care of the problem.  Note that the Pageant connection is not covered in the test-cases and probably needs to be checked by hand from time-to-time due to being Windows specific.

### DIFF
--- a/paramiko/_winapi.py
+++ b/paramiko/_winapi.py
@@ -219,8 +219,8 @@ class SECURITY_ATTRIBUTES(ctypes.Structure):
 
     @descriptor.setter
     def descriptor(self, value):
-        self._descriptor = descriptor
-        self.lpSecurityDescriptor = ctypes.addressof(descriptor)
+        self._descriptor = value
+        self.lpSecurityDescriptor = ctypes.addressof(value)
 
 def GetTokenInformation(token, information_class):
     """
--- a/sites/www/changelog.rst
+++ b/sites/www/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :bug:`461` Correct the obvious typo in SECURITY_ATTRIBUTES descriptor setter.
 * :bug:`402` Check to see if an SSH agent is actually present before trying to
   forward it to the remote end. This replaces what was usually a useless
   ``TypeError`` with a human-readable ``AuthenticationError``. Credit to Ken


stevecohen42:
This is a duplicate of 469, which was not pulled due to a missing changelog entry.

Folks should note that this is only applicable to the Windows platform and only when Putty's Pagaent is running when the Paramiko using application is run.  As such automated testing via Travis CI is impractical (impossible?).

bitprophet:
Yea, we can't CI for Windows (nor do the core devs run Windows) so support is 2nd tier only unfortunately. Thanks for updating this, adding it to release milestone.

Note to self, need to add credit for the 2-3 different tickets involved here.

remram44:
@bitprophet [appveyor](http://www.appveyor.com/) offers free CI on Windows, I use it on rpaths.

Also, I'm affected by this bug. I'll try to bisect the last working version.

remram44:
Caused by c5836ad5 (#419), between v1.15.1 and v1.15.2.

bitprophet:
Yea since I wrote that comment I actually had another user for another project help me set up appveyor, so I will add it to the others including paramiko once I have time :) (Like other Windows related support, though, it's never top priority, sorry :D)

remram44:
This is a very obvious typo that really breaks paramiko on Windows, bump

JoelBennett:
Chalk another one up for it breaking paramiko on Windows...

remram44:
Please? What do we have to do to see this get in?

Russell-Jones:
I guess we need more volunteers to be core devs on Paramiko. I'd quite like to do this myself, but a) I'm not sure I've the time required b) I think I'd need to deepen my knowledge of the relevant RFCs first

My impression is that @bitprophet  looks after this because it's integral to his Fabric project, but his requirements are a subset of all use cases, and he doesn't have time to do much more than keep it doing what he needs it to at the moment (Jeff, is that a fair summary?)

remram44:
I'm all for having more tests and a larger, more responsive development team. But here the we've had the fix for four months, it's a two lines long, and it's just sitting in a pull request waiting for someone to press the button. Am I going to have to merge this and vendor it in my software?

jaraco:
I'd be happy to help cut releases if @bitprophet would be willing to add me as an author/maintainer.

bitprophet:
I'm still figuring out the situation re: giving other folks commit bits.

Re: recent time allocation, have been focusing on Fabric v2 to the exclusion of most else, but that time is nearly done and I will be going through the release milestones here (note that this ticket *is* in the bugfix milestone!) real soon now.

shellster:
I'd also like to add a plea that this gets merged as soon as possible.
The fix works and is necessary for using paramiko in Windows environments.

Issue #489, opened 2015-02-20
paramiko channel default exit_status indistinguashable from a program error 


oxpa:
Hello.

I had a strange problem with a server either returning "-1" as an exit code or not returning an exit code at all.
These cases are indistinguishable in paramiko: channel.exit_status() returns -1 in either case.

I suppose, it would be useful to change default return value from "-1" to "None", to explicitly denote if there was an exit status from a command. "None" as a default value would allow handling "no exit code" specially, while checks like "if channel.exit_status() <0: handle_errors()" won't be broken.

Please, comment if you support this change: I would send a pull request with code and documentation changes.

lndbrg:
Ah, i definitely see you point. Do send a PR (it's a small one anyway) and we can have the discussion there.

bitprophet:
Rolling into #492

PR #492, opened 2015-02-24
default value for channel.recv_exit_status() changed: -1 -> None

oxpa:
As discussed at https://github.com/paramiko/paramiko/issues/489 :
If a channel returned -1 it's difficult to find out if it's a channel default exit code or there was no exit code from server.
This PR changes default from -1 to None. 
This doesn't seem to break any code inside paramiko: nothing uses exit_code directly.
This may break programs testing for "exit_code == -1" and facing a 'broken' server

### DIFF
--- a/paramiko/channel.py
+++ b/paramiko/channel.py
@@ -114,7 +114,7 @@ class Channel (ClosingContextManager):
         self.event = threading.Event()
         self.event_ready = False
         self.combine_stderr = False
-        self.exit_status = -1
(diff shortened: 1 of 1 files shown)


bitprophet:
As @lndbrg said, yea, this sounds like a good idea offhand.

Agree that this has implications for client software, so I'm gonna dump this into the 2.0 bucket (at the risk of making that a huge release...meh!)

PR #496, opened 2015-03-03
Various GSSAPI fixes

akruis:
The GSSAPI support in 1.15 and master contains a few minor bugs.

 - paramiko/auth_handler.py uses MSG_NAMES without importing it.
 - The documentation of of the return value and exception of ssh_check_mic in paramiko/ssh_gss is wrong.
 - in paramiko/auth_handler.py the variable "retval" is used uninitialised.
 - paramiko/kex_gss still uses the old PyCrypto random generator.

This pull request includes one commit for each bug.


wrouesnel:
I'm running this locally - paramiko as installed from pip is broken for GSS auth at the moment. This should be a high-priority merge.

mkomitee:
FYI, without at LEAST accepting the  change in c158db9833c70b2f57db0a58ab6133595c740d6f, GSSAPI support is completely broken.

bitprophet:
Referencing #584 for linkage, it's not directly related as far as I can tell (insofar as the newer GSSAPI module breaking things is not the same as this bundle of fixes) but presumably without fixing #584, many users will not be able to reap the benefits of these fixes here.

bitprophet:
I don't have the ability to readily test GSSAPI functionality (would like to sometime though, ugh) but since I've got at least a few people in here saying the PR fixes it, I'm going to trust them and just-merge :)

bitprophet:
Ah, #584 was a little confused, if people are actually following instructions and pip installing `python-gssapi` and not `gssapi`, there's no problem currently. Still good to have these linked tho.

PR #504, opened 2015-03-28
Support for using ForwardServer with the with keyword and some cleanup i...

haridsv:
I used the demo code from `forward.py` to create an adhoc tunnel and use it like this:

```python
with ForwardServer(0, remote_host, remote_port, ssh_transport) as serv:
    adhoc_port = serv.socket.getsockname()[1]
    # Use the adhoc_port
```

I found it useful enough that I wanted to contribute back so that others who are using `forward.py` as the their startup code can get benefit out of it.

### DIFF
--- a/demos/forward.py
+++ b/demos/forward.py
@@ -84,7 +84,7 @@ class ForwardServer(SocketServer.ThreadingTCPServer,object):
     def join(self):
         if self.th:
             # Workaround for Thread.join() prevents KeyboardInterrupt from getting raised.
-            while self.th.isAlive():
+            while self.th.is_alive():
                 self.th.join(0.1)
 
     def __enter__(self):
--- a/demos/forward.py
+++ b/demos/forward.py
@@ -46,7 +46,7 @@ DEFAULT_PORT = 4000
 
 g_verbose = True
 
-class ForwardServer(SocketServer.ThreadingTCPServer,object):
+class ForwardServer(SocketServer.ThreadingTCPServer, object):
     """
     Use ForwardServer in two ways:
 
--- a/demos/forward.py
+++ b/demos/forward.py
@@ -30,6 +30,7 @@ import getpass
 import os
 import socket
 import select
+from threading import Thread
 try:
     import SocketServer
 except ImportError:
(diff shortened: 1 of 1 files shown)


demos/forward.py
```diff
@@ -45,31 +46,75 @@
...
 
-class Handler (SocketServer.BaseRequestHandler):
+    def stop(self):
+        self.shutdown()
+        self.socket.close()
 
+    def join(self):
+        if self.th:
+            # Workaround for Thread.join() prevents KeyboardInterrupt from getting raised.
+            while self.th.isAlive():
```
lndbrg: Please use the newer ```is_alive```, there is no difference, but we are trying to keep the code base modern and forward compatible. :)

demos/forward.py
```diff
@@ -45,31 +46,75 @@
 
 g_verbose = True
 
+class ForwardServer(SocketServer.ThreadingTCPServer,object):
```
lndbrg: You should not need to inherit from object here.

demos/forward.py
haridsv: Yes, I will fix that.

demos/forward.py
haridsv: It is needed for super() to work as ThreadingTCPServer is an "old style" class. See this thread: http://stackoverflow.com/questions/9698614/super-raises-typeerror-must-be-type-not-classobj-for-new-style-class

lndbrg: Ok, cool, could you add a space after the comma? :)

haridsv: Done :)

bitprophet:
I need to dig deeper to see how this compares to the tunnel support in Fabric, as it may be close enough that it's better to recommend folks use Fabric's implementation instead.

That will be even more of a good idea after Fabric 2 launches as it'll be less opinionated and more library-driven, some Paramiko users don't find Fabric 1 a good fit for their needs.

lndbrg:
The thing that got me to say a-ok was the fact that is was in the demos folder, which are scripts that aren't getting that much love anyway and it's a great example of how to use the api.
But it's not changing any internals or exposing new supported classes (I do understan that the demos might end up as conventions of how to use it though)

bitprophet:
That's a fair point, though I've always been conflicted on the demos because so many folks clearly use them as literal jumping off points; I'd rather, post Fabric-2, be able to nuke the demos and say "look at this other codebase, it has well organized modules leveraging Paramiko, you can probably just use them by themselves, otherwise they're a great real world example". We'll see.

haridsv:
Unless I am mistaken, Fabric's tunnel is local forward only (equivalent to -L option of ssh), where as this code is meant to do the opposite (equivalent to -R option of ssh). At least that is what the documentation meant to me, and if not, I would myself prefer Fabric's implementation. In fact, I am already trying to use Fabric and Cuisine to the extent it is possible.

That said, I intended to create a patch for Fabric as well to add the reverse proxy functionality, again assuming it didn't already offer it. May be I didn't understand the documentation well enough, I will go through it again.

haridsv:
Here is theFabric's remote_tunnel() method: https://github.com/fabric/fabric/blob/master/fabric/context_managers.py#L487

"Create a tunnel forwarding a locally-visible port to the remote target.

For example, you can let the remote host access a database that is
installed on the client host"

In my case, I needed to access a remote rsync server from my local machine through a firewall. How would I use the above method for this use case? Is there another method that I am missing to see?

bitprophet:
Gotcha. There's definitely an open ticket or five for adding the reverse tunnel type to Fabric, so if you can please look for it/them :)

haridsv:
@bitprophet there is already a patch for this here: https://github.com/fabric/fabric/pull/939, so no point creating a patch with this change, no?

bitprophet:
Sounds about right, @haridsv, thanks! I will close this one for now, but it'll always be preserved here in case we want to try consolidating things down into Paramiko proper.

PR #522, opened 2015-05-04
add sha256,sha512 HMAC algorithm

xyxjin:
if the host only supported sha256/sha512,  the paramiko will report no supported HMAC.

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -26,7 +26,7 @@ import sys
 import threading
 import time
 import weakref
-from hashlib import md5, sha1
+from hashlib import md5, sha1, sha256, sha512
 
 import paramiko
 from paramiko import util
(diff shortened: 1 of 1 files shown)


Russell-Jones:
_preferred_macs = ('hmac-sha1', 'hmac-md5', 'hmac-sha1-96', 'hmac-md5-96', 'hmac-sha2-256', 'hmac-sha2-512')

Aren't sha2-512 and sha2-256 stronger than sha1 (at least [a bit](http://www.cigital.com/justice-league-blog/2014/01/21/sha2-vs-sha1/)) ? So shouldn't the order be

_preferred_macs = ('hmac-sha2-512', 'hmac-sha2-256', 'hmac-sha1', 'hmac-md5', 'hmac-sha1-96', 'hmac-md5-96',  )

? Would this cause problems somehow?

xyxjin:
i don't know which algorithm is stronger.  in our product requirement, there is only sha2-256 and sha2-512 will be supported. 
with  ('hmac-sha1', 'hmac-md5', 'hmac-sha1-96', 'hmac-md5-96', 'hmac-sha2-256', 'hmac-sha2-512') this order,  the paramiko can work normal and i can login our product.

if this two algorithm can be added into official branch, we need not to maintain our own branch about paramiko.

bitprophet:
This looks like a dupe of #356?

PR #524, opened 2015-05-08
Hang on large initial window size

oleksiy-markovets:
Hi,

I'm using _paramiko_ for file upload to SFTP server, but I faced problem that with some SFTP 
servers it just hangs in _putfo_. Nevertheless _openssh_'s _sftp_ uploads file without any probles.
After hours of debuging of _paramiko_ and _openssh_ I found out that it happens when servers 
returns _MAX_UINT_ in _intial window size_ (which complies SSHv2 standard).  Because of strage 
packing/unpacking logic _paramiko_ gets wrong _window size_ and hangs in a while.

I studied specifications (https://wiki.filezilla-project.org/SFTP_specifications) to find explanation of this
logic, but didn't find any. 

Anyway my change fixes my problem, but I have strong feeling that this _int_ packing/unpacking logic was implemented on purpose. Could you advice what the purpose of it is?

best regards
Oleksiy

### DIFF
--- a/paramiko/message.py
+++ b/paramiko/message.py
@@ -148,19 +148,6 @@ class Message (object):
         @return: a 32-bit unsigned integer.
         @rtype: int
         """
-        byte = self.get_bytes(1)
-        if byte == max_byte:
-            return util.inflate_long(self.get_binary())
-        byte += self.get_bytes(3)
-        return struct.unpack('>I', byte)[0]
-
-    def get_size(self):
-        """
-        Fetch an int from the stream.
-
-        @return: a 32-bit unsigned integer.
-        @rtype: int
-        """
         return struct.unpack('>I', self.get_bytes(4))[0]
 
     def get_int64(self):
@@ -257,19 +244,6 @@ class Message (object):
         self.packet.write(struct.pack('>I', n))
         return self
             
-    def add_int(self, n):
-        """
-        Add an integer to the stream.
-        
-        :param int n: integer to add
-        """
-        if n >= Message.big_int:
-            self.packet.write(max_byte)
-            self.add_string(util.deflate_long(n))
-        else:
-            self.packet.write(struct.pack('>I', n))
-        return self
-
     def add_int(self, n):
         """
         Add an integer to the stream.
--- a/paramiko/message.py
+++ b/paramiko/message.py
@@ -38,8 +38,6 @@ class Message (object):
     paramiko doesn't support yet.
     """
 
-    big_int = long(0xff000000)
-
     def __init__(self, content=None):
         """
         Create a new SSH2 message.
@@ -135,11 +133,7 @@ class Message (object):
 
         :return: a 32-bit unsigned `int`.
         """
-        byte = self.get_bytes(1)
-        if byte == max_byte:
-            return util.inflate_long(self.get_binary())
-        byte += self.get_bytes(3)
-        return struct.unpack('>I', byte)[0]
+        return struct.unpack('>I', self.get_bytes(4))[0]
 
     def get_size(self):
         """
@@ -251,11 +245,7 @@ class Message (object):
         @param n: integer to add
         @type n: int
         """
-        if n >= Message.big_int:
-            self.packet.write(max_byte)
-            self.add_string(util.deflate_long(n))
-        else:
-            self.packet.write(struct.pack('>I', n))
+        self.packet.write(struct.pack('>I', n))
         return self
 
     def add_int64(self, n):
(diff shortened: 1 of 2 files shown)


Russell-Jones:
It was added in https://github.com/paramiko/paramiko/commit/0e4ce3762a5b25c5d3eb89335495d3bb9054e3e7 @scottkmaxwell do you remember why you added "int handling to use mpint with a flag whenever the int is greater than 32 bits" ? Also, unless I'm misunderstanding it, the modified get_int() is looking for 0xff as a lead byte whilst in RFC 4251 an mpint is zero (00 00 00 00) or a 32-bit length followed by that number of bytes to be taken as a twos compliment integer unless the most significant bit of the first byte is 1, in which case a leading zero byte should be added, which appears quite different. Is this a workaround for something, perhaps?

Russell-Jones:
It appears to relate to https://github.com/paramiko/paramiko/commit/0e4ce3762a5b25c5d3eb89335495d3bb9054e3e7 add_int() does the reverse, adding 0xff when an int is > Message.big_int ( long(0xff000000) ) Both appear to be related to a set of Python 3 compatibility fixes.

oleksiy-markovets:
Yes, I saw that commit too. But it doesn't make sense to me, I didn't find any mention of such data format in SSHv2 specification. Also, I undestand that it was implemented like this intentionaly, but I don't understand why.

Russell-Jones:
My suspicion is that it's used internally and the message gets rewritten. I should probably just wait for Scott, but I find this intriguing and want to read through the code of message.py now :)

I notice that get_size() is defined twice, too, once in the way you did it (the second time it's defined), and once in the altered way.

scottkmaxwell:
I believe this is what I fixed in #482

oleksiy-markovets:
@Russell-Jones I also thought that it may be some kind of paramiko's extension of SSH protocol to address some internal issues, but it brakes compatibility with others SSH implementations.
@scottkmaxwell Yes, #482 should fix my problem, but I'm just wondering what is the purpose of get_adaptive_int/add_adaptive_int? I mean this datatype is not defined in SSH specification so potentially dangerous

Russell-Jones:
@oleksiy-markovets  I just noticed that there's some explanation from Scott in https://github.com/paramiko/paramiko/issues/353

oleksiy-markovets:
@Russell-Jones thank you for comprehensive explanation. 
IMHO, for sure, the best approach would be 2). Since SSHv2 specs operates only with *uint32*, *uint64* and *mpint* which just don't map to python's native *int* and *long*. So having generic *add()* method with *isinstance* magic inside is a very bad idea. 

After days of debuging of openssh code I became more or less familiar with its code. And yes, they use explicit functions for *uint32*, *uint64* and *mpint*.

ps. As I can see 3 different pull requests which fix this problem, so I hope it will be fixed in master shortly....

bitprophet:
Thanks for the detailed explanation, @Russell-Jones.

@oleksiy-markovets I haven't unpacked this problem in my head in a while but I think merging #482 is the best initial step and I hope to get to that bugfix milestone relatively soon.

Given a lot of these problems are *due* to an earlier "quite complex, requiring significant refactoring" change - the Python 3 support merge - I'm a bit scared to jump into another. Though I recognize it may be the best way forward depending on how #482 is received post-release.

scottkmaxwell:
Yes, I'd say get #482 merged, then look at removing the magic int code. Since that code only appears to be used in tests and would be totally invalid in an actual SSH context, best to change or remove those tests and drop the magic code. I am guessing that code was a nice shortcut back when it was originally written but (fortunately) it was either never used in any actual code or those usages were later made explicit.

oleksiy-markovets:
agree

bitprophet:
Good point, @scottkmaxwell - thanks.

corbinbs:
Hi - I encountered this issue earlier today with sftp uploads using paramiko to ProFTPD running on ubuntu.   This patch with the paramiko 1.15.2 release fixed it up for me.   In my case, I wasn't having any problems until the file size became larger than 32KB or so.  When transmitting really small files, I didn't encounter it.  Just wanted to pass that along in case that helps develop more test scenarios for it.
Thanks for doing the research on this and offering a solution until it can be worked into an official release

Tatsh:
This really works. All of a sudden downloading large files works correctly with `SFTPFile.get()`, downloading at full speed.

Server I'm downloading from uses ProFTPd with mod_sftp 0.9.9.

I don't know what to do until this gets merged or the fix in general gets into Paramiko, except to vendorise Paramiko :cry: (via Git submodule or some other decent way).

oleksiy-markovets:
I would suggest you to do what I did for our project, just download python-paramiko debian source package (assuming that you are using debian/ubuntu), patch it, rebuild and install on all machines that you need.

Tatsh:
At first I manually patched the file, but then...

It seems that Paramiko 1.16 already has the duplicate methods removed, and the `get_int()` patch. I removed paramiko completely from my virtualenv even by deleting it outright, and then ran `pip install --upgrade paramiko` and everything still worked.

The result is here: https://github.com/Tatsh/xirvik-tools/blob/master/xirvik/sftp.py#L102 <- the `except socket.timeout` is almost unnecessary as I have not encountered it since I upgraded Paramiko. I believe I was on 1.15.1 and it was not working at all. Got the hanging issue as described in this bug.

bskinn:
Hi! I'm helping bitprophet by going through the issues and PRs backlog, and for anything more than a few years old, we're just closing them out for now.

If this is still a live bug, please re-test and reproduce using a current version of Python (3.9+) and paramiko (3.0+), check to be sure the change still solves the problem, and create a new PR for it.

Thanks, and apologies for the long wait!

bskinn:
> Pretty sure I fixed that in #482. Glad to see you going and closing out these old issues Brian.

:+1:, thanks for following up, @scottkmaxwell.

No problem - going to be a gradual process, but I think we'll get there.

Issue #498, opened 2015-03-10
Access violation in _winapi.py


taumeson:
I'm running MySQL Workbench on Windows 8.1 and had to upgrade paramiko as part of a suggested bug fix on the MySQL bug tracker involving having pageant running at the same time (http://bugs.mysql.com/bug.php?id=76168). This was done successfully but I'm now receiving an access violation:

`ERR [20150310-15:49:00.816] thr=1   paramiko.transport:   File "C:\Program Files\MySQL\MySQL Workbench 6.2 CE/python/site-packages\paramiko\_winapi.py", line 150, in read`
`ERR [20150310-15:49:00.816] thr=1   paramiko.transport:     ctypes.windll.kernel32.RtlMoveMemory(out, self.view + self.pos, n)`
`ERR [20150310-15:49:00.818] thr=1   paramiko.transport: WindowsError: exception: access violation reading 0x00007FFE258A0000`

Here is the gist of the full traceback:
https://gist.github.com/taumeson/8853980184673ba88188

I'm running Windows 8.1 x64, and I know that Windows is a secondary concern. Hopefully it's something obvious but I am out of my element.

thorstenkampe:
Same issue here (trying a simple `dir`):
```
 ctypes.windll.kernel32.RtlMoveMemory(out, self.view + self.pos, n)
OSError: exception: access violation reading 0xFFFFFFFFFFFFFFFF
```
Issue occurs with Python 3.4.3 64bit Python 2.7.9 32bit on same machine is fine. Python3 issue or 64-bit issue?

thorstenkampe:
The issue only occurs in Python3 64-bit, neither in Python 32-bit/64-bit nor in Python3 32-bit

jaraco:
In back-porting #193 to [jaraco.windows](https://bitbucket.org/jaraco/jaraco.windows), the unit tests uncovered this bug as well. [The fix](https://bitbucket.org/jaraco/jaraco.windows/commits/76e7694c572e6fcf4cbf8141343028860ce6ebd3) has been released in jaraco.windows 3.3.

bitprophet:
Rolling into #525, thanks @jaraco

Issue #193, opened 2013-08-02
paramiko error on Windows


hobbsd:
I'm using paramiko via Fabric. Version 1.10.1 works fine, but 1.11.0 failed with this traceback (just the paramiko part):

  File "c:\python27\lib\site-packages\paramiko\client.py", line 342, in connect
    self._auth(username, password, pkey, key_filenames, allow_agent, look_for_keys)
  File "c:\python27\lib\site-packages\paramiko\client.py", line 474, in _auth
    self._agent = Agent()
  File "c:\python27\lib\site-packages\paramiko\agent.py", line 345, in __init__
    self._connect(conn)
  File "c:\python27\lib\site-packages\paramiko\agent.py", line 71, in _connect
    ptype, result = self._send_message(chr(SSH2_AGENTC_REQUEST_IDENTITIES))
  File "c:\python27\lib\site-packages\paramiko\agent.py", line 87, in _send_message
    self._conn.send(struct.pack('>I', len(msg)) + msg)
  File "c:\python27\lib\site-packages\paramiko\win_pageant.py", line 113, in send
    self._response = _query_pageant(data)
  File "c:\python27\lib\site-packages\paramiko\win_pageant.py", line 83, in _query_pageant
    pymap.write(msg)
  File "c:\python27\lib\site-packages\paramiko_winapi.py", line 133, in write
    ctypes.windll.msvcrt.memcpy(self.view + self.pos, msg, len(msg))
ValueError: Procedure probably called with too many arguments (12 bytes in excess)

jaraco:
@akx Thanks for fixing this issue. I'm curious as to why I never encountered the error in my testing. Do you know under what circumstances the error occurs? Was the issue only on 32-bit Python perhaps (as I develop/test primarily on 64-bit Python)?

alexeiramone:
The 1.16.0 is not working for me (Windows 7-64bit, Python 2.7 32bit).

jaraco:
@alexeiramone Same error [as above](https://github.com/paramiko/paramiko/issues/193#issue-17543938)?

alexeiramone:
Yes, I don't have the stack but the message was "ValueError: Procedure probably called with too many arguments (4 bytes in excess)". I had to downgrade it to 1.10.2 and then my fabric's fabfile run commands worked.

jaraco:
@alexeiramone You're looking for #613, fixed in #619.

Issue #505, opened 2015-03-28
Implement Windows CI with Appveyor


akx:
Since issues such as #488 mention there is no CI for Windows platforms, how about looking into http://www.appveyor.com/ (free for open-source projects!) to set that up? :)

lndbrg:
Might be interesting, there is a python demo here:
https://github.com/ogrisel/python-appveyor-demo/blob/master/appveyor.yml

bitprophet:
I & a contributor actually just implemented this for Invoke as part of pyinvoke/invoke#213 and while I'm not in the habit of checking it, it's there and seemed to work (latest master as of right now seems to still pass, yay: https://ci.appveyor.com/project/bitprophet/invoke/build/1.0.32/job/8asevwt4kp8c7hfd)

So if we add it here I'd follow the setup used there, most likely. It's not high on my priority list but at some point I want to also apply similar stuff also in use there (flake8, for example) to Paramiko and my other projects. High level goal for sure.

PR #525, opened 2015-05-08
Updated vendored jaraco.windows APIs

jaraco:
In this commit, I refresh the vendored _winapi.py module from its reference implementation. Merging this commit incorporates the fix for #193 and also includes (tested) fixes for #488 and #498.

It does restore the tab indentation, which makes it easier to sync the code with the reference implementation. I recommend keeping the tab indentation for that reason, though if you want to convert it to spaces, feel free to do so or ask and I can do that as a separate commit.

This PR comes against the 1.13 branch as I wanted to make the fixes available to be back ported to that version. It should be able to be readily merged into 1.14, 1.15, and master by simply accepting the whole `_winapi` file.


bitprophet:
Thanks!

Re: tabs, I am starting to add linting to my processes for projects and I bet that might get mad about tabs eventually, but that can be handled then (& probably just told to skip).

bitprophet:
Aw boo I thought we already implemented #505 and was trying to find the paramiko appveyor page. But no, I was thinking of invoke.

Tests pass for me on Unix with this merged, though, so full speed ahead...

PR #527, opened 2015-05-11
abot socks5 proxy

1again:
socket-like object via pass parameter of ‘connect’ function to use
socks5 proxy


uhe1231:
sorry!
i don't know how to use github 'pull request' change
so, i create other 'pull request'! :(

PR #531, opened 2015-05-26
client support for RSA ssh certificates

jasonrig:
I have extended the RSAKey class to support the ssh-rsa-cert-v01@openssh.com type of certificate authentication in a new RSACert class. A test case was added to test_pkey.py to test loading an RSA certificate (a signed public key that's the same one given in the tests already).
I think this class should be a reasonable example from which DSS and ECDSA certificate classes can also be created.

Let me know what you think!

Jason

### DIFF
--- a/paramiko/__init__.py
+++ b/paramiko/__init__.py
@@ -43,6 +43,7 @@ from paramiko.ssh_exception import (
 )
 from paramiko.server import ServerInterface, SubsystemHandler, InteractiveQuery
 from paramiko.rsakey import RSAKey
+from paramiko.rsacert import RSACert
 from paramiko.dsskey import DSSKey
 from paramiko.ecdsakey import ECDSAKey
 from paramiko.ed25519key import Ed25519Key
--- a/paramiko/rsacert.py
+++ b/paramiko/rsacert.py
@@ -1,0 +1,166 @@
+import base64
+
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.asymmetric import rsa
+
+from paramiko import util
+from paramiko.message import Message
+from paramiko.rsakey import RSAKey
+from paramiko.ssh_exception import SSHException
+
+
+class RSACert(RSAKey):
+    """
+    Certificate equivalent of RSAKey
+    See the `official SSH certificate format specification
+    <http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/PROTOCOL.certkeys?rev=1.9&content-type=text/x-cvsweb-markup>`_
+
+    OpenSSH versions newer than 5.4 support certificate-based authentication,
+    which can simplify authentication without relying upon heavy-weight
+    services such as LDAP. A good introductory guide to SSH certificates can
+    be found here:
+
+    https://www.digitalocean.com/community/tutorials/how-to-create-an-ssh-ca-to-validate-hosts-and-clients-with-ubuntu
+    """
+
+    def __init__(self, msg=None, data=None, privkey_filename=None,
+                 cert_filename=None, password=None, privkey_file_obj=None,
+                 cert_file_obj=None):
+        self.nonce = None
+        self.key = None
+        self.serial = None
+        self.type = None
+        self.key_id = None
+        self.valid_principals = None
+        self.valid_after = None
+        self.valid_before = None
+        self.critical_options = None
+        self.extensions = None
+        self.reserved = None
+        self.signature_key = None
+        self.signature = None
+        self.d = None
+        self.p = None
+        self.q = None
+
+        if cert_filename is not None:
+            msg = self._load_cert_from_file(cert_filename)
+        elif cert_file_obj is not None:
+            msg = self._load_cert(cert_file_obj)
+        elif cert_filename is None and cert_file_obj is None and data is None:
+            raise SSHException(
+                'Either a data object or a certificate file must be given')
+
+        if privkey_file_obj is not None:
+            self._from_private_key(privkey_file_obj, password)
+        elif privkey_filename is not None:
+            self._from_private_key_file(privkey_filename, password)
+
+        if (msg is None) and (data is not None):
+            msg = Message(data)
+
+        if msg is None:
+            raise SSHException('Key object may not be empty')
+        if msg.get_text() != 'ssh-rsa-cert-v01@openssh.com':
+            raise SSHException('Invalid key')
+
+        self.nonce = msg.get_string()
+
+        e = msg.get_mpint()
+        n = msg.get_mpint()
+        # Key might've been set by a private key file. If not, set it from the
+        # cert
+        if self.key is None:
+            self.key = rsa.RSAPublicNumbers(e=e, n=n).public_key(
+                default_backend())
+
+        self.serial = msg.get_int64()
+        self.type = msg.get_int()
+        self.key_id = msg.get_string()
+        self.valid_principals = msg.get_string()
+        self.valid_after = msg.get_int64()
+        self.valid_before = msg.get_int64()
+        self.critical_options = msg.get_string()
+        self.extensions = msg.get_string()
+        self.reserved = msg.get_string()
+        self.signature_key = msg.get_string()
+        self.signature = msg.get_string()
+
+    @property
+    def size(self):
+        return util.bit_length(self.public_numbers.n)
+
+    def get_name(self):
+        return 'ssh-rsa-cert-v01@openssh.com'
+
+    def get_public_key(self):
+        return RSAKey(key=self.key)
+
+    def _message_without_signature(self):
+        m = Message()
+        m.add_string('ssh-rsa-cert-v01@openssh.com')
+        m.add_string(self.nonce)
+        m.add_mpint(self.public_numbers.e)
+        m.add_mpint(self.public_numbers.n)
+        m.add_int64(self.serial)
+        m.add_int(self.type)
+        m.add_string(self.key_id)
+        m.add_string(self.valid_principals)
+        m.add_int64(self.valid_after)
+        m.add_int64(self.valid_before)
+        m.add_string(self.critical_options)
+        m.add_string(self.extensions)
+        m.add_string(self.reserved)
+        m.add_string(self.signature_key)
+        return m
+
+    def _message_with_signature(self):
+        m = self._message_without_signature()
+        m.add_string(self.signature)
+        return m
+
+    def asbytes(self):
+        return self._message_with_signature().asbytes()
+
+    def verify_certificate_signature(self):
+        return RSAKey(data=self.signature_key).verify_ssh_sig(
+            self._message_without_signature().asbytes(),
+            Message(self.signature))
+
+    def _load_cert_from_file(self, cert_file):
+        with open(cert_file, 'r') as f:
+            data = self._load_cert(f)
+        return data
+
+    def _load_cert(self, cert_file_obj):
+        data = cert_file_obj.read().replace('\n', '')
+        data = data.split()
+        if len(data) > 1:
+            data = data[1]
+        else:
+            data = data[0]
+        return Message(base64.b64decode(data.encode('ascii')))
+
+    @staticmethod
+    def generate(bits, progress_func=None):
+        """
+        Not implemented in RSACert because a certificate must be signed by a CA
+        and therefore loaded from some pre-existing data
+        """
+        raise Exception('Not implemented in RSACert')
+
+    @classmethod
+    def from_private_key_file(cls, filename, password=None):
+        """
+        Not implemented in RSACert because certificates cannot be generated
+        from private key files
+        """
+        raise Exception('Not implemented in RSACert')
+
+    @classmethod
+    def from_private_key(cls, file_obj, password=None):
+        """
+        Not implemented in RSACert because certificates cannot be generated
+        from private key files
+        """
+        raise Exception('Not implemented in RSACert')
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -64,6 +64,7 @@ from paramiko.message import Message
 from paramiko.packet import Packetizer, NeedRekeyException
 from paramiko.primes import ModulusPack
 from paramiko.py3compat import string_types, long, byte_ord, b, input, PY2
+from paramiko.rsacert import RSACert
 from paramiko.rsakey import RSAKey
 from paramiko.ecdsakey import ECDSAKey
 from paramiko.server import ServerInterface
@@ -203,6 +204,7 @@ class Transport(threading.Thread, ClosingContextManager):
     }
 
     _key_info = {
+        'ssh-rsa-cert-v01@openssh.com': RSACert,
         'ssh-rsa': RSAKey,
         'ssh-dss': DSSKey,
         'ecdsa-sha2-nistp256': ECDSAKey,
--- a/sites/docs/api/keys.rst
+++ b/sites/docs/api/keys.rst
@@ -17,6 +17,11 @@ RSA
 
 .. automodule:: paramiko.rsakey
 
+RSA Certificate
+===============
+
+.. automodule:: paramiko.rsacert
+
 ECDSA
 =====
 
--- a/tests/test_pkey.py
+++ b/tests/test_pkey.py
@@ -27,13 +27,14 @@ from binascii import hexlify
 from hashlib import md5
 import base64
 
-from paramiko import RSAKey, DSSKey, ECDSAKey, Ed25519Key, Message, util
+from paramiko import RSAKey, RSACert, DSSKey, ECDSAKey, Ed25519Key, Message, util
 from paramiko.py3compat import StringIO, byte_chr, b, bytes, PY2
 
 from tests.util import test_path
 
 # from openssh's ssh-keygen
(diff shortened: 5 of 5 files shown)


dimrozakis:
Any updates on this one? Are there any plans to get this merged?

bitprophet:
Will try and take a look next feature release pass. If you can update it so it merges/works cleanly with master that would help its chances :) thanks!

bitprophet:
Also, real-world reproduction steps for testing the functionality would be awesome, I haven't used cert-style authentication much yet.

jasonrig:
Hey @bitprophet and @dimrozakis, I'll rebase on master and clean up where required. I have reasonable experience with SSH certs, so I'll provide some extra docs to help introduce the concepts and real-world steps for use. Sometime next week, I hope :)

By the way, my requirements for certs+paramiko have been exclusively RSA where paramiko is the client and not server, so this functionality is not included here, but would be awesome if someone with time and expertise could contribute it... just for completeness.

jasonrig:
@bitprophet, I've rejigged the PR to merge cleanly with master, added a few tests and some docs (mostly references to people who write about it a lot better than I).
As for real-world reproduction, [DigitalOcean have a very good article](https://www.digitalocean.com/community/tutorials/how-to-create-an-ssh-ca-to-validate-hosts-and-clients-with-ubuntu) about how to set up certificate-based auth with OpenSSH. Conceptually it's quite simple:
* set up a CA using ssh-keygen
* modify SSH server config to trust the CA (can be done globally or per-user)
* sign a certificate with the CA keys using ssh-keygen
* put the cert and private key somewhere where the ssh command can find it... or use paramiko!
* log in to remote system

bitprophet:
Thanks @jasonrig! Greatly appreciated. As noted, it's in the next feature release bucket so I'll walk through here when I get to that.

dimrozakis:
Hey @jasonrig , just returned from holidays, I'm thrilled to see you jumped in and updated the PR, much appreciated!

@bitprophet is there any estimation of when the next feature release will come out?

bitprophet:
@dimrozakis I'm really bad at estimates so I need to stop giving them! I will note that I tend to work in sprints lately, so the next time I personally need to push out something feature-y I will also go down the milestone list and try to knock out as much as I can.

typpo:
Adding my support to see this PR merged after two years.  We use this kind of signed cert to manage server access across our organization without having to set up individual ssh keys on each server.

Facebook engineering recommends this sort of setup to control authorization on large scale production systems: https://code.facebook.com/posts/365787980419535/scalable-and-secure-access-with-ssh/  I expect this use case will become more common!

jasonrig:
Rebased on master, tests still pass ^.^
I can squash the commits if the code review is favourable.

Mletter1:
I look forward to testing this out. Will this be integrated with the clients connect function or is it already integrated through transport? what would the example client connection look like using certs.

jasonrig:
Hi @Mletter1 
This should work in almost exactly the same way as the current key-based authentication, since it simply overrides the RSAKey class with some extra certificate-specific details. In other words, if you currently use an RSA key for authentication, the RSA certificate method should be a drop-in replacement.

My use-case has been specifically for RSA certificates, and only for Paramiko acting as a client. As mentioned in the PR, implementing other types of certificates (DSA, ECDSA, etc.) should be trivial, but I just have not had the time. As for server logic, I don't feel I have experience enough to implement this without a closer reading of the spec and Paramiko code.

radssh:
It looks like keys accessible via `paramiko.Agent` that are OpenSSH certs will work on earlier versions of paramiko, which may provide a temporary workaround until this finds its way into a release.

wlhunt:
Greetings, thanks for the help. mletter1 and I have tried a client certificate authentication with your code and can't get it to work. The server is ok because a cert auth works from a command line ssh, typed manually. Paramiko publickey auth works fine if we place the public key in the server's authorized_keys file. Here's what we are doing:

`pKey = paramiko.RSAKey.from_private_key_file( args.keyFile )`
`cert = paramiko.RSACert(privkey_filename=args.keyFile, cert_filename=args.certFile)`
`cl = paramiko.SSHClient()`
`# below gives "Authentication failed" exception`
`cl.connect(myServer, username=args.userName, pkey=cert )`
`# below also gives "Authentication failed" exception`
`#cl.connect(myServer, username=args.userName, pkey=pKey )`

The certificate "id_rsa-cert.pub" is sitting alongside the private key, "id_rsa", in the same directory.

The exception thrown is:
  File "build/bdist.linux-x86_64/egg/paramiko/client.py", line 621, in _auth
paramiko.ssh_exception.AuthenticationException: Authentication failed.
# The log entry is: 
INFO:paramiko.transport:Authentication (publickey) failed.

It's my impression that paramiko is not presenting the cert to the server. Are we setting this up wrong or not calling the right connect()?

Thanks again,
Warren

wlhunt:
Hi again. I checked all my keys, my certificate, my calling options, my server setup, my python version, my paramiko version (this pre-merged version) and it worked! My apologies for the confusion!

To reiterate, calling connect() with pkey=certificate works:

`cert = paramiko.RSACert(privkey_filename=args.keyFile, cert_filename=args.certFile)`
`cl = paramiko.SSHClient()`
`cl.connect(myServer, username=args.userName, pkey=cert )`

Thanks for your help and contribution to paramiko. Looking forward to your work being merged to master.
Warren

saifhhasan:
@jasonrig : why are we delaying on this ? Are we not confident enough of pulling this in. Many people and companies can benefit from this feature. 2 years of delay is a HUGE and I can see from the conversation that many engineering hours were also poured into this.
Can we move forward with this ? At Facebook we are using paramiko for remote host testing and we need certificate support. I patched this into our internal pull of paramiko and it worked for us.

@jasonrig : One enhancement request
- Can we automate trying of SSH certificate if available on host at a default location instead of manually specifying it's location

jasonrig:
Hi @saifhhasan,
I'm not part of the paramiko project, just a fellow contributor that also uses this feature for our projects as well :) I cannot speed up the review processes that @bitprophet and others go through.

I can try to work on your enhancement request when I get some time, but I suppose it would be ideal that this PR is merged in some form and then at least we have an accepted base from which we can move forward with extra features. As you can see, I targeted only a "minimal viable product" here that suited my needs (i.e. client-side support for RSA certificates), but it is far from a complete certificate solution. I would be really pleased to see others improve what I have started!

ploxiln:
This looks good to me - clean and stand-alone.

saifhhasan:
@jasonrig Sorry I mistagged you in my earlier comment. Updated it accordingly. I appreciate your effort and diff which has been helping me get going by patching it in our local version of paramiko.

@bitprophet : Can we get some push on this one ?

incognick:
@jasonrig I tried integrating this with Fabric which uses paramiko and paramiko would not attempt to use the RSACert.  

I had to add `RSACert` [here](https://github.com/paramiko/paramiko/blob/master/paramiko/client.py#L557)

Also had to implement the `from_private_key_file" function in rsacert as: 
```
    @classmethod
    def from_private_key_file(cls, filename, password=None):
        return RSACert(privkey_filename=filename, cert_filename=filename+'-cert.pub', password=password)
```
Now, obviously, the cert_filename can't stay that way. Not sure how to get this properly integrated into your PR. 

However, the above code works and I can run fabric just fine using my cert file.  Thanks for your help.

jasonrig:
@incognick I'm happy to take advice on whether, for example, `from_private_key_file` should be implemented with a sensible guess for the certificate file name as the public key. Though I'm a bit hesitant to weave RSACert into other paramiko components if only to keep the scope of the PR small enough to not delay any review further than it has already been.

I would say subsequent improvements beyond this core enhancement should be submitted as additional PRs down the track. However I do not take issue with anyone forking my repo and basing a new branch on this PR with their desired improvements, and submitting their own changes as a new PR. So if you see any room for improvement, please go ahead and implement it!

bitprophet:
I'm +1 on merging as-is and figuring out integration with other cert-unaware codebases like Fabric afterwards. Moved this to the 2.3 milestone since it's thematic with some other auth-related features there like smartcard support (#827). (Actually makes me wonder whether that should be made aware of this...thoughts welcome, tho always possible to add later too.)

bitprophet:
Successfully tested out OpenSSH-level certificates, so I understand that level of things. Also skimmed the RFC. So next up is actual review, manual testing & merging \o/

bitprophet:
As-I-go commentary (I will address most/all while merging):

- Not 100% sure why this inherits from RSAKey itself, seems like very little actual overlap?
  - Plus the constructor renames/reorders a bunch of the init args (unless this is an artifact of PR age but I don't think so) so it's not even really usable as a semi-drop-in replacement
  - I'm thinking I'll try to do a 2nd-draft version of this, then, with a distinct new class that simply encapsulates an `RSAKey`. We'll see. Need to manually test & understand the test suite changes first.
- There seem to be unused encryption param attributes, eg `self.d/p/q`. (`p/q` could be part of DSS but that's not complete; none of the keys in the extension RFC seem to mention a `d` at all; it's not referenced anywhere else in the code either.)
- Not a fault of the OP, but, starting to try to get away from thoughtless vanilla `SSHException` use, so we should make some subclasses for where those are used here, if possible.
  - Though spots that 100% mimic RSAKey behavior can probably stay.
  - Similarly, the not-implemented exceptions for the later methods should be `NotImplementedError`, not `Exception` (nobody should ever raise vanilla `Exception`, honestly! How would you catch it?!)

bitprophet:
@ploxiln I edited that a bunch after you emoji'd it...FYI :D

ploxiln:
I see :)

It's good that you're investigating and finding the relationship with RSAKey is a bit awkward, and there are some currently unused attributes, sooner rather than later. I clearly didn't do that :)

It might be a good idea to open a pull request with your in-progress cleaned-up merges, or just point to your branch, so you might get some eyes and help on it.

bitprophet:
Was planning to at least link the branch once it's pushed, yea. Also trying very hard not to overstep bounds here re: overall key class organization stuff...>_>

----

Another random side note: wondering if we want to add 'full' client side support for this in the same way OpenSSH client does; e.g. its `-i` and `CertificateFile`. Especially re: file paths, I could see folks being confused, because! Right now we take a path explicitly assuming it's "the cert" - but `-i` does a tiny bit of magic:

> If no certificates have been explicitly specified by the `CertificateFile` directive, `ssh` will also try to load certificate information from the filename obtained by appending `-cert.pub` to identity filenames.

That said - not super worried because we can add that later if folks feel it's missing, and at the moment, one is either explicitly instantiating `RSACert` (and thus must be aware of where both key and cert file paths are) or relying on `Transport._key_types` (which will be dealing with already-not-file-related `Message` objects).

bitprophet:
Actually, just realized that would solve the possible confusion with `RSACert` constructor - users forgetting to give `cert_filepath` - since we could have it implicitly set `cert_filepath` to `filepath` + `-cert.pub` if it (and `cert_file_obj`) are `None`. Worth thinking about.

Still very unhappy with this fragile constructor anyways, so, maybe I'll meet myself halfway and rip it up into multiple constructors.

bitprophet:
Nah. Much of the annoyance I'm having stems from the original classes too (insofar as they mutate their responsibilities depending on whether they were given private key data or not). This stuff's all due a bigger shakeup later so I will defer to then. Still need to add a bunch more tests around corner cases though.

radssh:
You might get a decent compromise in the bang-for-the-buck category by being intentionally ignorant about the contents of the certificate blob. On the client side, while the components that make up the certificate are mildly interesting, for use in client authentication, that is left up to the server side to make sense of; the client offers the certificate as a blob, then if the certificate could be used for authentication, it proves possession of the corresponding private key. The client does not even need to unravel the blob itself.

This is why the current Paramiko code works just dandy with certificates loaded in a ssh-agent, as long as it can get the blob contents, and sign the composited message, it's all the same on the server side. Getting the details of the certificate content, or even being able to generate new certificates would be nice to have features, but probably pales in comparison to just being able to blindly loading a blob and being able to offer it as an alternative publickey authentication attempt, if available.

bitprophet:
In expanding the test suite (and by extension, some functionality/exceptions) I ran into one case that has no immediately obvious answer: what to do if someone instantiates the class with a _public-only_ key alongside the cert? (Provided that it is the "correct" public key; mismatching public key data is covered elsewhere.)

It's a little tempting to except here too, because it implies user error:

- for one thing, certificate data includes the public numbers anyways, so it's at best a bit silly;
- the primary use case for no-private-key-material certificate objects is that of authenticating the remote end, which only requires the certificate blob and does not require any additional key material (again in part because of the previous bullet point).
  - In most cases where Paramiko is using a cert object for this purpose, it's getting it straight out of a Message object, and so once again, this is not a situation where the constructor should really be expecting additional key kwargs.
- This is one of the MANY reasons I want to eventually split up this class hierarchy's constructors into multi-constructors...there are just too many of these "probably never gonna happen, but when it does, it's going to waste some poor fool's entire afternoon" possibilities.

All that said, it would not technically _hurt_ to be given the pubkey material a 'second time', so it's remotely possible someone is dealing with grandfathered-in kwarg dicts or whatever, and would be unhappy about an exception. Which is why I'm torn - it does not violate correctness.

bitprophet:
@radssh That's a good point, and also applies to the comment I posted after yours (didn't see yours til just now.) The public key part, which is embedded in the cert, is all that's necessary for auth as a client; my understanding above is a bit off (just re-read Transport to remind myself; probably adding some comments therein so future-me doesn't make this mistake again), private key is not required when one has a cert.

bitprophet:
Again, I suspect eventually breaking our classes up would also make it harder for high-functioning morons like myself to get tied in knots. Handing "the authentication process" discrete certificate and private key objects, for example, would make it more obvious that the cert largely just replaces the process of distributing a `known_hosts` file to the remote end (yes, alongside a bunch of other mostly-authorization related stuff that we currently don't care about unpacking.)

radssh:
> private key is not required when one has a cert

not quite - to be of much use at all, the client needs the private portion to do the signing (prove possession of (unlocked) private portion).

Having a cert without the private portion of the key is almost like having the public key (plus a bunch of other publicly viewable/extractable items). Not much you can do with it alone, other than regurgitate it. The cert does contain a form of embedded public key contents, but it requires some minor unraveling to get to (RSA you can get (e,n), DSS (p,q,g,y), but as discrete values, not as a ready-made Message that would fit with Paramiko), so to transform cert -> pubkey can be done, but would need a case by case handling. Generally, since you don't need to open the envelope to pass it along, just need to sign the sealed envelope; why bother with those details.

Point is, unless you really "need" the pubkey, you can skip the unraveling completely, and just handle the cert blob, as long as you can submit the whole enchilada (Message with blob, and other bits) + (Private key signed Message) to the remote side to authenticate. It's the remote side that (should) verify not only the signature of the auth request done by "your" private key, but also the unraveled blob, which itself is a signed Message by some CA, presumably identifiable by the server as trusted, according to the CA public key on file on the server side.

The worst that can happen on the client side, is that the server declines authentication using the cert, for whatever reason (untrusted CA, malformed signature, validity interval expired, principal name does not match, etc.), and the client has to try another auth method (plain public key) or just cope with the failed authentication. Yes, it would be nice to be able to elegantly detect inconsistencies in advance, but that could be construed as scope creep 😄

bitprophet:
Once again, I edited my comments once I had that same realization (re: private key.) 🙃

bitprophet:
More waffling - realizing that my initial read on the `msg` and `data` params is also wrong; in normal `RSAKey` usage, they may only be public key material, and in `RSACert` they are expected to be cert data. So they don't quite map along the same lines as `filename` and `file_obj`, which are expected to be private key related. (Have I mentioned yet how frustrating this API is to understand or extend? I have? Okay.)

Upshot is that I'm probably just gonna put the constructor args back the way I found them and file it under "let's just reorganize the entire friggin thing in a .0 release".

bitprophet:
(At least the docstring is getting a massive workout from all this...and the test suite...)

bitprophet:
Noticed another oops, the `key` param doesn't appear to be used at all. In `RSAKey` it's a way of handing in an already-created key object (i.e. some Cryptography primitive) so presumably we should honor it here too.

bitprophet:
Not done but pushing what I have so far today as the 531-int branch. Think this is the correct diff view (had to construct it manually because Github wasn't wanting to show me the original branch despite it still existing on Jason's end...):

    https://github.com/jasonrig/paramiko/compare/rsacert...paramiko:531-int#files_bucket

Can't promise exactly when I will pick this back up again but it shouldn't be more than a few days to a week tops. I intend to finish fleshing the tests out, the hopefully small bits of extra implementation needed to satisfy them (mostly around raising more footgun errors; and I'll be happy to skimp on some of 'em for now) and then merging.

bitprophet:
While I was supporting my dayjob, @radssh implemented what I think I'll consider a second draft of this (in #1042), with a different approach that has benefits over the one here:

- It's type-agnostic because it just treats the cert blob as what it is, an already-formatted SSH protocol message, and so there's no need to break it down only to reconstruct it again later. Thus it should support all common cert types out of the box.
  - This has the drawback of not being useful for server-side support, for which I suspect this PR's approach may eventually be required.
- It doesn't have any of the drawbacks of the direct-subclass approach (all the constructor, NotImplementedError method, etc, annoyances) because it just tacks an extra method onto the existing classes.
- It's less code, with fewer edge/corner cases.

Obviously I'm not super tickled at the idea of dropping or deferring something both the OP and myself put a bunch of effort into, but given that using certs to authenticate as a client feels like the much more common use case, #1042 serves that better and with less code and so it's more tempting to merge first.

bitprophet:
Dropping this from the release milestone and I encourage all followers of this ticket to try out the #1042 functionality when it's released. Keeping this alive for now & hoping anyone who finds the #1042 code lacking can report details on their use cases here. Right now I'm expecting that will chiefly be a) server-side support and b) host-certificate validation support on the client side.

bitprophet:
On second thought (and partly because #1042 got expanded during testing & does more now!), going to close this. Any remaining missing functionality can be reported as its own ticket/s.

PR #558, opened 2015-07-16
Typo causes failure if Putty Pageant is running (rebased)

remram44:
This is just #488 rebased on current master (conflict in changelog.rst)

> Pageant connections fail due to a typo in the setter for the security agent descriptor. Fixing the typo takes care of the problem. Note that the Pageant connection is not covered in the test-cases and probably needs to be checked by hand from time-to-time due to being Windows specific.


bitprophet:
Actually want this based on the appropriate bugfix release and not master, but thanks anyway!

bitprophet:
(See #488 - it is actually getting merged finally)

Issue #530, opened 2015-05-21
'The message [...] is not extractable!' on calls to getfo


ryanahall:
I'm currently running into this error when calling getfo on files that exist on the remote server. It seems to fail during CMD_STAT requests when the file is already opened for reading, but the same CMD_STAT will work if I manually call it without opening the file first. I imagine this is probably related to the configuration of the sftp server I'm connecting to, but I figured I should file an issue ayway.

Here's two manual calls of CMD_STAT on the same server/file, one with it already opened and the other not. getfo() uses the first approach of opening and then calling stat() seen here: https://github.com/paramiko/paramiko/blob/master/paramiko/sftp_client.py#L688-L689

```
In [43]: with sftpc.open('[removed]', 'rb') as fr:
   ....:     sftpc._request(CMD_STAT, '[removed]')
   ....:
---------------------------------------------------------------------------
IOError                                   Traceback (most recent call last)
<ipython-input-43-b71244708fe4> in <module>()
      1 with sftpc.open('[removed]', 'rb') as fr:
----> 2     sftpc._request(CMD_STAT, '[removed]')
      3

/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/paramiko/sftp_client.pyc in _request(self, t, *arg)
    727     def _request(self, t, *arg):
    728         num = self._async_request(type(None), t, *arg)
--> 729         return self._read_response(num)
    730
    731     def _async_request(self, fileobj, t, *arg):

/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/paramiko/sftp_client.pyc in _read_response(self, waitfor)
    774                 # synchronous
    775                 if t == CMD_STATUS:
--> 776                     self._convert_status(msg)
    777                 return t, msg
    778             if fileobj is not type(None):

/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/paramiko/sftp_client.pyc in _convert_status(self, msg)
    800         elif code == SFTP_NO_SUCH_FILE:
    801             # clever idea from john a. meinel: map the error codes to errno
--> 802             raise IOError(errno.ENOENT, text)
    803         elif code == SFTP_PERMISSION_DENIED:
    804             raise IOError(errno.EACCES, text)

IOError: [Errno 2] The message [[removed]] is not extractable!

In [44]: sftpc._request(CMD_STAT, '[removed]')
Out[44]:
(105,
[...]

Issue #194, opened 2013-08-02
"Message is not extractable" errors from offspec SFTP servers (possibly all IBM)


muraleee:
*[Maintainer's note: a) this seems to be specific to IBM SFTP servers or others which place limits on how many times one can `stat` a file; b) there are two potential fixes for this, #562 and #579]*

I have the following code with initalization of credentials removed.
Printing of directory listing works, however "get" fails with the following exception

paramiko [Errno 2] <....filename removed...> The message  is not extractable!

I worked around by making a  copy of getfo and removing prefetch call in it. 

-----------------------------------------------------------------------------------

    sftp = paramiko.SFTPClient.from_transport(t)

    dirlist = sftp.listdir('.')
    print "Dirlist:", dirlist
    
    sftp.chdir('Inbox')
    dirlist = sftp.listdir('.')
    print "Dirlist:", dirlist
    
    for fil in dirlist:
        fl = file(localpath + "/" + fil, 'wb')
        sftp.get(file, localpath + "/"+file)
        print fi

bitprophet:
See #530 which *feels like* the same issue, and indicates it may be a problem with open file handles on the server side.

torkil:
I second that #530 feels like the same issue, and I think it's a case of paramiko not working well with IBM Sterling, most likely due to a bug/config issue. This may be related http://www-01.ibm.com/support/docview.wss?uid=swg1IT02065

Like others in this and related threads I have been able to get around it by modifying the getfo-call. The files I care about is only "extractable" once so right now I am out of files to fetch, but I will return to this later in the week.

bufke:
I had to comment out two lines to work around this. I assume this will break the callback feature.

```
class HorribleIBM_SFTPClient(paramiko.SFTPClient):
    def getfo(self, remotepath, fl, callback=None):
        with self.open(remotepath, 'rb') as fr:
            #file_size = self.stat(remotepath).st_size
            #fr.prefetch()
            print('here!!!')
            size = 0
            while True:
                data = fr.read(32768)
                fl.write(data)
                size += len(data)
                if callback is not None:
                    callback(size, file_size)
                if len(data) == 0:
                    break
        return size
```

bitprophet:
@bufke Did you get a chance to try the changes from #562 or #579? They seem to be less-disruptive modifications achieving a similar end.

[...]

PR #562, opened 2015-07-23
Call stat before opening file for reading

torkil:
Calling stat after opening a file for reading will in some server
configurations(?) lead to IOError: The message 'filename' is
not extractable.

### DIFF
--- a/paramiko/sftp_client.py
+++ b/paramiko/sftp_client.py
@@ -685,9 +685,10 @@ class SFTPClient(BaseSFTP, ClosingContextManager):
 
         .. versionadded:: 1.10
         """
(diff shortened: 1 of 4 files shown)


bitprophet:
Do you have more info on what "some server configurations" means exactly - i.e., is there a reproducible way we can test this out?

The change seems reasonable at a glance either way, though, so thanks!

torkil:
The erroneous server configuration is something which have been pointed out in issues I think are related see, #194 #530 

I also think this is due, at least in some way, to http://www-01.ibm.com/support/docview.wss?uid=swg1IT02065

Issue #547, opened 2015-06-29
IndexError during private key reads


vogxn:
Running some python tests that execute commands against EC2 machines. Following error is encountered periodically when reading the private key needed to ssh to the machines. The key is an RSA key of size=1024

Following is the stacktrace:
```log
    def _read_private_key(self, tag, f, password=None):
        lines = f.readlines()
        start = 0
        while (start < len(lines)) and (lines[start].strip() != '-----BEGIN ' + tag + ' PRIVATE KEY-----'):
            start += 1
        if start >= len(lines):
            raise SSHException('not a valid ' + tag + ' private key file')
        # parse any headers first
        headers = {}
        start += 1
        while start < len(lines):
            l = lines[start].split(': ')
            if len(l) == 1:
                break
            headers[l[0].lower()] = l[1].strip()
            start += 1
        # find end
        end = start
>       while (lines[end].strip() != '-----END ' + tag + ' PRIVATE KEY-----') and (end < len(lines)):
E       IndexError: list index out of range

build/bdist.linux-x86_64/egg/paramiko/pkey.py:276: IndexError
```

vogxn:
Here's a more complete stack trace that I failed to attach previously

```log
lib/cluster.py:305: in execute
    out = self.shell.run(split_cmd)
build/bdist.linux-x86_64/egg/spur/ssh.py:158: in run
    ???
build/bdist.linux-x86_64/egg/spur/ssh.py:168: in spawn
    ???
build/bdist.linux-x86_64/egg/spur/ssh.py:243: in _get_ssh_transport
    ???
build/bdist.linux-x86_64/egg/spur/ssh.py:261: in _connect_ssh
    ???
build/bdist.linux-x86_64/egg/paramiko/client.py:307: in connect
    ???
build/bdist.linux-x86_64/egg/paramiko/client.py:444: in _auth
    ???
build/bdist.linux-x86_64/egg/paramiko/pkey.py:183: in from_private_key_file
    ???
build/bdist.linux-x86_64/egg/paramiko/rsakey.py:55: in __init__
    ???
build/bdist.linux-x86_64/egg/paramiko/rsakey.py:165: in _from_private_key_file
    ???
build/bdist.linux-x86_64/egg/paramiko/pkey.py:255: in _read_private_key_file
    ???
[...]

PR #565, opened 2015-07-29
compare end of key file before reading it

vogxn:
have been running this change for a week and haven't encountered the `IndexError` again.


bitprophet:
This is re #547, presumably? thanks!

bitprophet:
Seems like a reasonable approach, if you can write a quick test that fails w/o this change and succeeds with it in place, that'd be awesome. Thanks!

PR #572, opened 2015-08-10
Tweak BER encoding to be DER for SSH keys

coreywright:
Do not add sign padding to a data value's content length in long
definite form so it can be decoded as DER.

Paramiko, like OpenSSH, consumes private SSH keys in BER, but unlike
OpenSSH, and against Postel's Law (ie decode liberally and encode
conservatively), it produces keys in BER, not DER (a stricter subset
of BER).  This breaks tools that consume private SSH keys and only
implement DER decoding, tools that are compatible with OpenSSH output.

The difference between Paramiko's BER and OpenSSH's DER (as
implemented by OpenSSL), for the purposes of SSH RSA and DSA private
keys, is the number of subsequent length octets in the long definite
form.  Paramiko includes sign padding which sometimes introduces an
extra, unnecessary octet.  This octet is unnecessary because content
length is not encoded as two's complement and therefor doesn't require
a leading sign bit, but as it is always non-negative and only ever
produces a leading octet of zeros, ie zeros in the most significant
bits, it doesn't alter the content length value and therefor
incidentally legal in BER (see ITU-T X.690 (11/2008), section
8.1.3.5).  While being extra and unnecessary for BER, the octet
prevents private SSH keys output by Paramiko from being decoded as DER
because for DER "length encoding shall be... encoded in the minimum
number of octets" (see ITU-T X.690 (11/2008), section 10.1).


coreywright:
this is prompted by an issue [1] caused by an openstack nova change [2].

the new output vectors (ie `(RSA|DSS)_PRIVATE_OUT`) were generated with a script [3] to insure they were openssh equivalent and DER (and incidentally equal to the input vectors in tests/test_(rsa|dss).key, presumably generated by ssh-keygen).

i considered removing the `(RSA|DSS)_PRIVATE_OUT` variables and instead reading at test-time the private key text from `tests/test_(rsa|dss).key` as they are now equal and could be considered redundant, but i noticed that the same is true for `tests/test_ecdsa.key` and `ECDSA_PRIVATE_OUT` so i interpreted the redundancy as intentional and left the test logic using the (updated) variables.

[1] https://github.com/mitchellh/packer/issues/2526
[2] https://github.com/openstack/nova/commit/3f3f9bf22efd2fb209d2a2fe0246f4857cd2d21a (aka http://git.openstack.org/cgit/openstack/nova/commit/?id=3f3f9bf22efd2fb209d2a2fe0246f4857cd2d21a)
[3] https://gist.github.com/coreywright/c47c522e69f0bace3e00

bitprophet:
Thanks for the detailed input!

I'm wondering if this might be obviated by the handful of places where the BER class is no longer used in https://github.com/paramiko/paramiko/pull/394/files, which is slated to merge relatively soon. Don't have time to dig myself right now but it's something that sprung to mind.

coreywright:
@bitprophet You are correct that #394 obviates this.

This is most easily seen in that the changes to the test RSA and DSA private key outputs are the same between the two PRs as they both cause Paramiko to produce them in DER format (ie https://github.com/paramiko/paramiko/pull/394/files#diff-523fff3457d80985c9d2429eb3aecb6eL43 and https://github.com/paramiko/paramiko/pull/572/files#diff-523fff3457d80985c9d2429eb3aecb6eL43 or `git diff 9e67e71 be4abcc tests/test_pkey.py`).

Thanks for reviewing this PR and making me aware of #394!

I'm looking forward to a release of Paramiko with #394 merged.

coreywright:
Okay, the intention of this PR is fulfilled with #394: to output private SSH keys in the more widely accepted DER format.

But #394 does not address the bug in the BER class that this PR addresses: BER still incorrectly includes sign padding in content length octets (as documented in this PR's commit message).

As my only concern was DER formatting of private SSH keys, which is indirectly fixed with #394, I'm leaving this PR closed for someone else to reopen (though I presume with the adoption of pyasn1 in #394 that the proper course of action is to remove Paramiko's BER class and use whatever pyasn1 provides for BER, as BER [and DER] are encoding rules for ASN.1 and presumably pyasn1 provides an implementation of such).

coreywright:
as #394 is still not merged (understandably because of its size) and this remains an issue, i'm reopening this.

ploxiln:
#394 was merged, and also this looks related to #568 although subtly different

PR #575, opened 2015-08-15
Use correct variable name

CptLemming:
Attempting to access a global variable named descriptor. Issue causes fatal error on Windows.


bitprophet:
See #488

Issue #577, opened 2015-08-24
SFTP get stalls


stephencpope:
I believe I have identified and fixed the problem of SFTP gets hanging (probably the cause of at least issues  #560 and #515 ). I was experiencing this problem when retrieving large files (15G +) over long haul WAN links (low bandwidth and large latency) most reliably.

I diagnosed this using python-2.7.9 on CentOS 5.8 using paramiko 1.15.2 although the fix should more or less apply for any recent version; until recently I had been using paramiko 1.9.0 with python-2.5.1 without encountering this error.

Here is a simple program to demonstrate the problem:

    #!/usr/bin/env python
    
    from __future__ import with_statement
    
    import sys
    import os
    import time
    import threading
    import paramiko
    
    lock = threading.Lock()
    progress = None
    
    def callback(sofar, total):
        global lock, progress
        with lock:
            progress = (sofar, total)
    
    if len(sys.argv) != 5:
        print >>sys.stderr, 'usage: %s host username remote local' % sys.argv[0]
        sys.exit(1)
    
    host = sys.argv[1]
    username = sys.argv[2]
    remote = sys.argv[3]
    local = sys.argv[4]
    
    transport = paramiko.Transport((host,22))
    path = os.path.join(os.environ['HOME'], '.ssh', 'id_dsa')
    key = paramiko.DSSKey.from_private_key_file(path)
    transport.connect(username=username,pkey=key)
    sftp = paramiko.SFTPClient.from_transport(transport)
    
    last = 0
    
    t = threading.Thread(target=sftp.get, args=(remote,local, callback))
    t.setDaemon(True)
    t.start()
    
    while True:
        time.sleep(5)
        with lock:
            if progress is not None:
                if progress[0] == progress[1]:
                    break
                if progress[0] == last:
                    print >>sys.stderr, 'stalled %d of %d' % progress
                    sys.exit(1)
                print '%d of %d' % progress
                last = progress[0]

With 1.15 HEAD an example run (you'll need to use your own appropriate remote host, user, and file as mine are all internal to a corporate WAN) looks like this:

    % ./test-sftp-stall remote-host remote-user remote-file local-file 
    20348928 of 15472925029
    23199744 of 15472925029
    23855104 of 15472925029
    stalled 23855104 of 15472925029

[...]

PR #578, opened 2015-08-24
fix sftp stalls

stephencpope:
Proposed fix for #577. Implemented as a branch off 1.15 HEAD as I was not interested in the extra credit of determining exactly when the issue arose earlier in the history. However, it would apply in principal beginning at the point where sftp prefetch was implemented (sometime after 1.9.0).

### DIFF
--- a/paramiko/sftp_client.py
+++ b/paramiko/sftp_client.py
@@ -747,10 +747,10 @@ class SFTPClient(BaseSFTP, ClosingContextManager):
                     raise Exception('unknown type for %r type %r' % (item, type(item)))
             num = self.request_number
             self._expecting[num] = fileobj
-            self._send_packet(t, msg)
             self.request_number += 1
         finally:
             self._lock.release()
+        self._send_packet(t, msg)
         return num
 
     def _read_response(self, waitfor=None):
@@ -761,15 +761,19 @@ class SFTPClient(BaseSFTP, ClosingContextManager):
                 raise SSHException('Server connection dropped: %s' % str(e))
             msg = Message(data)
             num = msg.get_int()
-            if num not in self._expecting:
-                # might be response for a file that was closed before responses came back
-                self._log(DEBUG, 'Unexpected response #%d' % (num,))
-                if waitfor is None:
-                    # just doing a single check
-                    break
-                continue
-            fileobj = self._expecting[num]
-            del self._expecting[num]
+            self._lock.acquire()
+            try:
+                if num not in self._expecting:
+                    # might be response for a file that was closed before responses came back
+                    self._log(DEBUG, 'Unexpected response #%d' % (num,))
+                    if waitfor is None:
+                        # just doing a single check
+                        break
+                    continue
+                fileobj = self._expecting[num]
+                del self._expecting[num]
+            finally:
+                self._lock.release()
             if num == waitfor:
                 # synchronous
                 if t == CMD_STATUS:
--- a/paramiko/sftp_file.py
+++ b/paramiko/sftp_file.py
@@ -478,7 +478,7 @@ class SFTPFile (BufferedFile):
         while True:
             with self._prefetch_lock:
                 # spin if in race with _prefetch_thread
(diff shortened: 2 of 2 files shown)


stephencpope:
Bah! I don't do no stinkin' Python 3. My submission is failing the various Python 3 variations (no dict.has_key() method). I am sure some simple syntactic change can fix that.

stephencpope:
So I fixed the has_key problem. Python 3.2 passes. Python 3.3 and Python 3.4 both timeout, apparently hanging up in the SSHClient tests, which do not involve SFTPClient in any way; I suspect the problem is not caused by me. I got and built Python 3.4.3, ran the tests manually myself, and they worked fine. Don't know what to say ...

bitprophet:
This is most likely covered by some older tickets as well, need to dig those up sometime to make sure use case coverage is better understood. (I'm paranoid about fixes breaking other areas of code...:D)

stephencpope:
we've been using the fix here for lots of very large downloads (files as big as 35G thanks to recent market turmoil) as well as plenty of small ones with no adverse effects. Not that that will give you warm fuzzies!

bitprophet:
It's still helpful though! So thanks :) and mostly I just want to make sure we link related tickets for closure or noting that "hey we're merging this and so um watch out for side effects".

stephencpope:
So one thing to look out for: it seemed to me that the SFTP code wasn't safe to use for both getting and putting simultaneously. While my fixes address some locking issues, I didn't even attempt to address this since it is not a use case of mine.

Tatsh:
Status on this? I'm getting 'Garbage packet' exceptions when retrieving some large files (like anything above 50 MiB it seems, but sometimes things randomly work).

Other times the connection is extremely slow. It might be the server (as always can be the case), but lftp is able to retrieve such files very easily at around 2 MiB/s.

Is this a useful fix for that? I am using Python 3 and it appears this needs patching for 3.4?

shuhaowu:
Any status on this? I think I just ran into this problem, although I am not 100% positive. I used pdb to trace it to hang on: https://github.com/paramiko/paramiko/blob/1.16/paramiko/sftp_file.py#L477.

PDB shows:

```
(Pdb) p self._prefetch_lock.locked()
True
```

When I step over that line it hangs.

When I straced one of the processes, I get

```
poll([{fd=3, events=POLLIN}], 1, 100)   = 0 (Timeout)
poll([{fd=3, events=POLLIN}], 1, 100)   = 0 (Timeout)
poll([{fd=3, events=POLLIN}], 1, 100)   = 0 (Timeout)
poll([{fd=3, events=POLLIN}], 1, 100)   = 0 (Timeout)
...
```

fd=3 is the socket to the SFTP server.

Tatsh:
I apply this patch on the version of Paramiko I'm using now (in my virtualenv). It works really well. Using Python 3.4. My project really depends on it. So I hope it does get merged at some point.

shuhaowu:
Applied this patch on top of 1.16 today. Seems to have fixed this issue.

Is it possible we get some project member to look into merging this? I can help solve the merge conflict and re PR if OP is not going to do that.

bitprophet:
Noting for the record that I wasn't able to recreate this (using the description in #577) even across a (not slow...) WAN connection, but thanks to @stephencpope's excellent analysis (thank you!) and reading the code, I'm reasonably sure I grok what it's doing.

More importantly, I'm not able to detect any serious regression in my integration branch :)

stephencpope:
Thank you!

bitprophet:
@stephencpope it's remotely possible this patchset is causing #730. I'll be trying to confirm/deny that tomorrow; if you happen to have time before then & want to try reproducing it (it's probably intermittent and mostly affecting Python 3.5, hooray) please comment on #730 with any findings.

bitprophet:
As seen on #730 I was wholly unable to recreate the occurrence, so, hopefully a false alarm. *glares suspiciously at all computers everywhere*

Tatsh:
I am glad because I have wanted this merged for a long time.

PR #579, opened 2015-08-26
Update sftp_client.py and sftp_file.py so that STAT on a file is only…

mwarnes:
… called once and the value passed to any required subroutines.

This is to allow support for the IBM Sterling Integrator SFTP Server when files are defined with an extractability count of 1.

Files with an extractability count of 1 will become unavailable once the GET request has started and any subsequent STAT commands to obtain the file size will fail with a file not found error.


bitprophet:
Thanks!

This feels like it probably duplicates #562? Just doing a triage pass now but we should figure that out before merging either.

bitprophet:
Going with #562, see notes in #194.

Issue #580, opened 2015-09-03
Add support for RFC 6668


ericwb:
RFC 6668 recommends new hash algorithms to replace the aging md5, sha1.

From https://tools.ietf.org/html/rfc6668

```
  hmac-sha2-256     RECOMMENDED   HMAC-SHA2-256
                                   (digest length = 32 bytes,
                                    key length    = 32 bytes)

  hmac-sha2-512     OPTIONAL      HMAC-SHA2-512
                                   (digest length = 64 bytes,
                                    key length    = 64 bytes)
```

bitprophet:
See PR

PR #581, opened 2015-09-03
Add support of RFC 6668

ericwb:
Adds hmac-sha2-256 and hmac-sha2-512 as supported and recommended
MACs.

fixes: #580

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -26,7 +26,7 @@ import sys
 import threading
 import time
 import weakref
-from hashlib import md5, sha1
+from hashlib import md5, sha1, sha256, sha512
 
 import paramiko
 from paramiko import util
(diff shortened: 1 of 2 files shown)


bitprophet:
Doesn't this partially duplicate #356?

ericwb:
True, there is some overlap, but looks like #356 was mostly to address the key exchange.  It lacks support for sha512 which means our support of RFC 6668 is incomplete.

bitprophet:
Cool, thanks for clarifying!

ilmerovingio:
+1

bitprophet:
TFW you use just slightly incorrect language re: Github commit hooks. See referencing commit above, this is in now, had to manually pull changes in due to overhauls re #356. Thanks again.

PR #586, opened 2015-09-10
Added additional attribute preserve to SFTPClient.get method, which p…

deceq:
…reserves remote file access_time and modification_time after downloading to local path


deceq:
i'll fix tests later

deceq:
now closed, then will be reopened

PR #599, opened 2015-10-20
- Added hmac-sha2-512 and hmac-sha2-256

andreycizov:


ploxiln:
this was implemented separately, see #581

PR #606, opened 2015-11-03
Handle special keys

Metalgrid:
Correctly send special keys (*arrow keys, insert, delete, end, home, page up, page down, etc*), which send an escape sequence and a keycode from a single key press (e.g. *up-arrow* sends `^[[A`). This fixes *vim* running in an interactive terminal session, but requires to put `sys.stdin` in non-blocking mode so that regular keys are sent without filling a buffer.

### DIFF
--- a/demos/interactive.py
+++ b/demos/interactive.py
@@ -16,7 +16,6 @@
 # along with Paramiko; if not, write to the Free Software Foundation, Inc.,
 # 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA.
 
-
 import socket
 import sys
 from paramiko.py3compat import u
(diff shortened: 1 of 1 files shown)


bitprophet:
Thanks for this. I like the idea of making nontrivial demo sessions work better, though I wonder whether this could increase chances of the base case breaking for some users?

Guessing not (I can't think of many situations offhand where a nonblocking stdin would make somebody's terminal wig out) but something I'm always wary of.

Maybe some additional comments noting what the `fcntl` calls are doing & how best to remove them if they're problematic, would be the best middle road.

Metalgrid:
I removed the non-blocking code and switched to `os.read()` on `sys.stdin`'s file descriptor instead. So far it works just as good without using non-blocking mode.

flobz:
I can confirm that it fixed #1494.
Can we merge please ?

PR #610, opened 2015-11-05
added ecdsa to demo.py

mchlt:
I added ecdsa key authentication to demo.py
To avoid adding another almost duplicate `elif auth == 'e'` block I added a dict to contain the different key authentication parameters.

This is a precursor to me adding nistp384 and nistp521 to ecdsa key authentication, but the change is independent of that so created a separate PR.

PS, apologies for the two previous PRs which I closed again. I needed to learn to branch... _facepalm_


mchlt:
This is in #611

PR #611, opened 2015-11-05
Add ecdsa-sha2-nistp384 and ecdsa-sha2-nistp521 key authentication

mchlt:
Added ecdsa-sha2-nistp384 and ecdsa-sha2-nistp521 key authentication support.
Didn't have to change too much. The ecdsa module already supported them, so just needed a little work to get the right ssh parameters into `ECDSAKey` and having it use ecdsa's features for identifying which key size is being used.
Added bits parameter to ecdsa `generate` function to make it compatible with RSAKey and DSSKey. Because of this, I could easily add ecdsa to `demo_keygen.py`. 

Tested client login against all key sizes, added ecdsa t demo.py.
Generating ecdsa 256, 384 and 521 size keys works with demo_keygen.py
Existing tests for ecdsa-sha2-nistp256 all still work fine without modification.

### DIFF
--- a/demos/demo.py
+++ b/demos/demo.py
@@ -60,32 +60,29 @@ def agent_auth(transport, username):
 
 def manual_auth(username, hostname):
     default_auth = 'p'
-    auth = input('Auth by (p)assword, (r)sa key, or (d)ss key? [%s] ' % default_auth)
+    key_auth = {
+      'r': {'name':'RSA key', 'keyfile':'id_rsa', 'class':paramiko.RSAKey},
+      'd': {'name':'DSS key', 'keyfile':'id_dsa', 'class':paramiko.DSSKey},
+      'e': {'name':'ECDSA key','keyfile':'id_ecdsa','class':paramiko.ECDSAKey}
+    }
+    auth = input('Auth by (p)assword, (r)sa, (d)ss, or (e)cdsa key? [%s] ' % default_auth)
     if len(auth) == 0:
         auth = default_auth
 
-    if auth == 'r':
-        default_path = os.path.join(os.environ['HOME'], '.ssh', 'id_rsa')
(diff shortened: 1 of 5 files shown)


bitprophet:
Thanks for this. Seems like it includes the changes from #610 too? If so, you can close #610, I don't actually mind "multi-change" PRs if the 2nd change is just in the demo folder :)

mchlt:
I've added a fix to transport.py where in the case of GSS, the key exchange algorithm list was replaced further on in the code, which has caused a later update which added `diffie-hellman-group-exchange-sha256`to be forgotten in that location.  Moved the GSS preferred kex list to the top for consistency

bitprophet:
We're planning to merge #394 into master soon and that'll make it hard to merge this as-is. Pretty sure it'd be possible to take the general approach in here and apply it to the updated `ecdsakey.py` afterwards; I'm going to bump this to a 2.1 milestone with that in mind. (If @mchlt finds time to make this #394-friendly before 2.0 releases, I can make it part of 2.0, doesn't matter a ton to me either way.)

CrazyCasta:
I'm going to have a version of my pull request done soon. Should I link that here or what?

CrazyCasta:
Ok, I've updated my PR, now #731. A quick overview of how mine differs from mchlt's:

* I have some wrappers around my definitions of the various curves, should make it easier to add more in the future if we ever want that.
* Added tests to test all the new stuff.
* I didn't bother with the demo stuff. I just want this in by 2.0 so I can stop using my own branch of paramiko.
* I'm not familiar enough with GSSAPI/Kerberos to know what I'm doing wrt making the gss_ keys, so I just left that out for now.

bitprophet:
Close to merging #731. I'd ideally like the demos to get updated too but it can wait for a bugfix or similar update once someone else has the time :)

Issue #602, opened 2015-10-28
Cannot parse OpenSSH 6.5 formatted private keys


kabel:
The key format described at https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.key is not currently supported causing key authentication problems when that format is used. The current parser erroneously assumes that the key file is unencrypted.

Related to #387

mchlt:
Looking into this issue. 
Some additional info about the key format here:
http://www.tedunangst.com/flak/post/new-openssh-key-format-and-bcrypt-pbkdf

mchlt:
OK, i now have some working code that correctly reads the base64 blobs from a new OPENSSH type key.
Unfortunately, OpenSSH have chosen a rather exotic key derivation function to derive the key from the passphrase: bcrypt_pbkdf
The only Python implementation of that that I could find is in `py_bcrypt` (https://github.com/grnet/python-bcrypt).  I'll use that for now to see if I can actually do the decryption.

Wondering what's paramiko's stance on introducing new module dependencies?  I'm thinking maybe do the import in a `try..except` and then throwing an exception only if someone tries to load an encrypted new format key?

bitprophet:
@mchtl We do the "try/except/complain" for other optional deps right now, such as `python-gssapi` for GSSAPI support ([docs](http://www.paramiko.org/installing.html#optional-dependencies-for-gss-api-sspi-kerberos)). So we don't want to do it willy-nilly, but there's precedent.

That said, "a new privkey format used by OpenSSH" feels more mainstream a feature than Kerberos/GSSAPI support, so I lean towards "if we can get this implemented before I finish the other work for 2.0, let's put `python-bcrypt` in that release as a hard dependency".

If you haven't, you might also look to see if https://github.com/pyca/cryptography has it ([this](https://cryptography.io/en/latest/hazmat/primitives/key-derivation-functions/#cryptography.hazmat.primitives.kdf.pbkdf2.PBKDF2HMAC) mentions pbkdf2 but I'm not crypto-fluent enough to tell if it's truly the same as `python-bcrypt`'s C-ext) or is willing to add it, since paramiko 2.0 will be relying on `cryptography` anyways.

bitprophet:
Also, thanks a lot for looking into this! :heart:

bitprophet:
Continuing discussion in here since not 100% sure what'll happen to #618 (it might want closing & replacing with another PR):

[...]

PR #618, opened 2015-11-10
Support for new OpenSSH 6.5+ private key format 

mchlt:
Added support for new OpenSSH 6.5+ private key format. See issue #602.
This code only adds the ability to READ that key format, not write it.

I could only test the keys on one box as I found that my Solaris, Raspberry pi and Oracle Linux boxes all still have older versions of SSH. So only tested this on Ubuntu 15 against OpenSSH_6.9p1.

Figuring out the actual format and which blob was which key was quite the puzzle, it's not really documented completely anywhere.

As discussed previously, this does add a dependency on py-bcrypt 0.4
The cryptography module does not have bcrypt, nor the required bcrypt_pbkdf function (which is similar, but not quite the same as pbkdf2...)


cjac:
Have you, in your wanderings through this code, found a way to identify the Public Key Algorithm of a given private key file?  Or even the content of that file would work.

XVilka:
Should be rebased, I think. Any chance to get this merged?

XVilka:
Should be closed since superseded by https://github.com/paramiko/paramiko/pull/1343

bitprophet:
Rolling into #1343 as noted.

PR #616, opened 2015-11-08
Fix connection error-handling code on Python 3

nchammas:
Convert the dictionary view returned by `errors.keys()` to a list. Thanks to @edk0 for pointing this out.

Fixes #615.

PR #624, opened 2015-11-19
TypeError: 'dict_keys' object is not subscriptable

rkubik:
paramiko/ssh_exception.py
    Class: NoValidConnectionsError
    Line: 166

My errors object contains the following before the TypeError is raised:

    {('10.0.2.2', 2222): ConnectionRefusedError(111, 'Connection refused')}

In Python 3 dict.keys() returns an iteratable but not indexable object.

Ref: http://stackoverflow.com/questions/26394748/nltk-python-error-typeerror-dict-keys-object-is-not-subscriptable

### DIFF
--- a/paramiko/ssh_exception.py
+++ b/paramiko/ssh_exception.py
@@ -162,7 +162,7 @@ class NoValidConnectionsError(socket.error):
         :param dict errors:
             The errors dict to store, as described by class docstring.
         """
-        addrs = errors.keys()
+        addrs = list(errors.keys())
         body = ', '.join([x[0] for x in addrs[:-1]])
         tail = addrs[-1][0]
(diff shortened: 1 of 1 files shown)


bitprophet:
Thanks! Adding to bugfix release.

nchammas:
I think this is a dup of #616.

bitprophet:
Yup, good catch, @nchammas - sorry about that, the volume of tickets is such that I can't even keep all the dupes cached in my head anymore :cry: Thanks & sorry, @rkubik !

nchammas:
Yeah, I think GitHub could help better here if they automatically checked for duplicates like Stack Overflow does when you start typing up a question.

PR #632, opened 2015-12-03
Make sure the callback function is called only once

ab9-er:
Since the callback is called before checking the size of data chunk, the callback function is called twice.

### DIFF
--- a/paramiko/sftp_client.py
+++ b/paramiko/sftp_client.py
@@ -695,10 +695,10 @@ class SFTPClient(BaseSFTP, ClosingContextManager):
                 data = fr.read(32768)
                 fl.write(data)
                 size += len(data)
-                if callback is not None:
-                    callback(size, file_size)
                 if len(data) == 0:
                     break
+                if callback is not None:
+                    callback(size, file_size)
         return size
 
     def get(self, remotepath, localpath, callback=None):


bitprophet:
Can you elaborate on why this is problematic in your case? I'm assuming your callback is doing something other than the typical "progress bar/percentage/etc" type output, but offhand I'm not sure why having the callback called 2x in a row with the same `size` value would actually break things.

Seems like an innocent enough change but it also feels like one of those spots where somebody else may be relying on this only-sort-of-buggy behavior, somehow.

ab9-er:
Consider this log,
~~~
[INFO] - [03/12/2015 06:37:18 PM] : 43% transfer completed
[INFO] - [03/12/2015 06:37:18 PM] : 87% transfer completed
[INFO] - [03/12/2015 06:37:18 PM] : 100% transfer completed
[INFO] - [03/12/2015 06:37:18 PM] : File transfer from SFTP to local TEMP complete
[INFO] - [03/12/2015 06:37:18 PM] : 100% transfer completed
[INFO] - [03/12/2015 06:37:18 PM] : File transfer from SFTP to local TEMP complete
~~~

In my trivial case, the logger is called twice. In case, if some other function is called it could result in errorneous behavior.

bitprophet:
Backported to 1.15.x, added test, refactored (I think you probably meant to update `putfo` too, but it got missed :D now that won't be a problem anymore...), added changelog, should be all set now. Thanks!

ab9-er:
👍

PR #635, opened 2015-12-09
Allow auth retries when client presents a bad key

sruggier:
Without this, a user has to go out of their way to prevent their SSH
client from sending the server that key, even if it's just a valid key
that paramiko doesn't support yet.


sruggier:
I discovered this issue by interacting with a server that uses paramiko while my agent included an ed25519 key.

bitprophet:
Linking this to #387 since it feels like part of the same set of crummy behaviors. Thanks!

Issue #637, opened 2015-12-10
Paramiko does not work with pycryptodome


patrys:
[pycryptodome](https://github.com/Legrandin/pycryptodome) is a self-contained fork of pycrypto that also happens to be the only crypto package that can be installed on Windows using nothing but `pip`.

It's mostly compatible save for some APIs that pycryptodome's author [considers insecure](https://github.com/Legrandin/pycryptodome/issues/6):

> Paramiko verifies RSA signatures in an insecure way and it does that by means of some old, questionable parts of the PyCrypto API (specifically, the method `verify` of an RSA key object).
>
> PyCryptodome is (fortunately or unfortunately, depending on how you look at it) not 100% compatible to PyCrypto: I explicitly removed or modified those few APIs that are inherently dangerous (like the `verify` method).

It would be great to be able to use Paramiko with pycryptodome even just for convenience it provides.

bitprophet:
First I've heard of cryptodome, interesting. Thanks for the info!

We're actually planning to move away from PyCrypto anyway in 2.0 (see #394) so I don't know that this will get much love. If someone provides a minimally-disruptive patch against, say, 1.13 and up (at least 1.15 and up) I might merge it so people on the 1.x line have the option of using cryptodome. Would probably need to not make huge waves to be worth the risk tradeoff, though, FWIW.

johnthagen:
+1 for ``pycryptodome`` support, or also check out ``cryptography``: https://cryptography.io/en/latest/

Though it's likely ``cryptography`` would require more porting work.

Legrandin:
FWIW, I have created a minimal patch for this in #646.

bitprophet:
Let's roll this into #646 - either that'll get merged or the 2.0 switch off of pycrypto will "solve" it for enough users.

PR #646, opened 2015-12-18
Add support for PyCryptodome (and PyPy)

Legrandin:
This patch makes paramiko work with PyCryptodome in addition to PyCrypto.
With the former, paramiko can run under PyPy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
   - "3.3"
   - "3.4"
   - "3.5"
+  - "pypy"
 install:
   # Self-install for setup.py-driven deps
   - pip install -e .
--- a/README.rst
+++ b/README.rst
@@ -44,7 +44,7 @@ Requirements
 
 - `Python <http://www.python.org/>`_ 2.6, 2.7, or 3.3+ (3.2 should also work,
   but it is not recommended)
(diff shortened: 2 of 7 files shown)


bitprophet:
Thanks for this. I have gut-feeling-only concerns about some of the changes (mostly the ones in `rsakey.py`) and as mentioned in #637 I'm hoping to largely obviate the issue by merging cryptography.io support for 2.0. (That's another minor/moderate concern - how well those two PRs interact, should we merge both in either order.)

So for now I'll leave this open & attach to a new "post-2.0 1.x feature release" milestone so that when I re-examine the state of 1.x vs 2.x, it's on deck. Thanks again!

alex:
@bitprophet if you plan to merge the cryptography.io patch, I'd say there's no reason to land this, it's 100% code that my PR removes/replaces.

bitprophet:
@alex Good to have that confirmed (if you see the earlier notes / the other ticket, I was kinda hoping it would be the case)! Thanks :)

blackwind:
It's great that cryptography.io is coming, but the PR for it has been open for a year and a half and counting, and in the interim, using paramiko on Windows is FAR more difficult than it needs to be. Any chance you'd reconsider merging this as a stopgap?

Issue #651, opened 2015-12-31
Trouble connecting with windows putty to demo_server.py


jamercee:
Attempts to connect from windows putty to an instance of `demo_server.py` results in:

    C:\> python demo_server.py
    Read key: 60733844cb5186657fdedaa22b5a57d5
    Listening for connection ...
    Got a connection!
    *** No channel.

From the `demo_server.log` file:

    ERR [20151230-20:38:57.563] thr=1   paramiko.transport: Exception: Client asked to handle paket userauth-request
    ERR [20151230-20:38:57.565] thr=1   paramiko.transport: Traceback (most recent call last):
    ERR [20151230-20:38:57.565] thr=1   paramiko.transport:   File "C:\Python27\lib\site-packages\paramiko\transport.py", line 1761, in run
    ERR [20151230-20:38:57.565] thr=1   paramiko.transport:     self.auth_handler._handler_table[ptype](self.auth_handler, m)
    ERR [20151230-20:38:57.565] thr=1   paramiko.transport:   File "C:\Python27\lib\site-packages\paramiko\auth_handler.py", line 499, in _parse_userauth_request
    ERR [20151230-20:38:57.565] thr=1   paramiko.transport:     %MSG_NAMES[ptype])
    ERR [20151230-20:38:57.565] thr=1   paramiko.transport: SSHException: Client asked to handle paket userauth-request

If you disable GSSAPI authentication from putty, the connection is successful.

Just wondering if this is a known issue ?

Versions:

    Windows 8.1 Pro
    putty 0.64
    Python 2.7.3, 32-bit
    pyasn1 0.1.9
    paramiko 1.16.0

Identical behavior from Kitty 0.64.1

PR #652, opened 2015-12-31
Gracefully fail `gss-with-mic` to allow fallback to next authentication

jamercee:
While researching issue https://github.com/paramiko/paramiko/issues/651, I discovered what I think is a bug in `_parse_userauth_request()`. If the authentication method is **gssapi-with-mic** and the ptype is not `MSG_USERAUTH_GSSAPI_TOKEN`, then the function raises an exception. I believe the correct response is to fail the request to allow the client to fallback to alternate methods.

I've applied and tested the patch associated with this pull request and it now works as expected.

I've also attached a copy of the Putty log to this message if you'd like to see the specifics of how it attempts authentication.

[putty.txt](https://github.com/paramiko/paramiko/files/75568/putty.txt)

### DIFF
--- a/paramiko/auth_handler.py
+++ b/paramiko/auth_handler.py
@@ -356,7 +356,7 @@ class AuthHandler (object):
             m.add_string(p[0])
             m.add_boolean(p[1])
         self.transport._send_message(m)
- 
+
     def _parse_userauth_request(self, m):
         if not self.transport.server_mode:
             # er, uh... what?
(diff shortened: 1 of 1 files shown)


bitprophet:
FTR the current behavior was added in 3e1f9f09b1da0397f82e4ee9e1886f5271705e29 (via #4 / by @SebastianDeiss). If Sebastian is active I wouldn't mind a +1 from him on this since he's the original author of this part of the GSSAPI support.

Sans that input, I'd probably merge this since it's in an uncommonly-used part of the codebase and "let the client try other auth types" does seem like a more useful approach than excepting. Thanks @jamercee :)

SebastianDeiss:
:+1: 

Yes, that's definitely a bug!
Of course the client can always abort the current authentication by sending a `SSH_MSG_USERAUTH_REQUEST` packet, so raising an exception is not a good idea. :-)

Responding with `AUTH_FAILED` to allow fallback to other auth methods is the correct behavior.
Thanks for the patch @jamercee.

bitprophet:
Thanks @SebastianDeiss !

bitprophet:
Grumble grumble github cherry pick grumble. This has been ported to 1.15.x and above. Thanks @jamercee :)

PR #668, opened 2016-01-28
NoValidConnectionsError: error.keys() is not subscriptable

nehz:
error.keys() is a dict keys view object in py3


nehz:
Duplicate of #616

PR #688, opened 2016-02-12
Removed support for hmac-md5 and truncated hmac-sha1

alex:
These are being removed from OpenSSH as well: https://github.com/openssh/openssh-portable/commit/714e367226ded4dc3897078be48b961637350b05


alex:
Failing test appears to be a known one.

bitprophet:
Dropping this in 3.0 mostly because I've been telling folks 2.0 will remain API/behavior compatible. As noted elsewhere 3.0 will still go out in the nearish future :) thanks!

PR #691, opened 2016-02-18
Fixed "Bad file descriptor", "Broken pipe" and another kind of error…

qualab:
Fixed "Bad file descriptor", "Broken pipe" and another kind of errors on multithreading usage of PosixPipe when we've got close() and set() at the same time. In such case check for self._closed is invalid and not synchronized with close() method called from separated thread.

### DIFF
--- a/paramiko/pipe.py
+++ b/paramiko/pipe.py
@@ -28,6 +28,7 @@ will trigger as readable in `select <select.select>`.
 import sys
 import os
 import socket
+import threading
 from paramiko.py3compat import b
 
 
@@ -45,12 +46,17 @@ class PosixPipe (object):
         self._set = False
         self._forever = False
         self._closed = False
+        self._lock = threading.Lock()
     
     def close(self):
-        os.close(self._rfd)
-        os.close(self._wfd)
-        # used for unit tests:
-        self._closed = True
+        self._lock.acquire()
+        try:
+            os.close(self._rfd)
+            os.close(self._wfd)
+            # used for unit tests:
+            self._closed = True
+        finally:
+            self._lock.release()
     
     def fileno(self):
         return self._rfd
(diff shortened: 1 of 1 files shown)


qualab:
Seems tests.test_transport.TransportTest should have greater timeout or it is unstable.

bitprophet:
See #691 for more info.

whbos:
We've been running into the same errors. Will this code be committed?

tsellers-r7:
We are seeing similar.  Is there something we can do to help get this fix or a version of it into production?

bitprophet:
Adding to a release milestone.

I'll admit that changes like this make me nervous, the project has a history of back & forths on this type of code, basically bug whack-a-mole where it fixes one set of issues but creates (or regresses, often) into another.

That said, this basically just looks like "rub some locking on it" so it's better than similar changes that are just shuffling `close` calls around and suchlike.

If y'all want to make this an even easier merge, and can come up with a new test for the test suite that catches the bug this fixes, that'd be awesome. Not required, however.

qualab:
Actually there is a very hard way to reproduce this. System must run many threads which are asynchronously do via SSH short but recent actions using paramiko. There is recent bug but in case of system mix close() and set() recent enough for race on socket descriptor (start set() due close() running). I found that only using strace, there was one of the most hard bugs I ever found.
But it used in tests of Parallels Automation (platform for Parallels Plesk, now renamed to Odin) with that fix and never fire after that without any slow down of test run.

rafaelthca:
I've ran into the same issue. I am the developer of https://github.com/OmniDB/OmniDB, a python web based database management tool.

OmniDB instantiates several threads to perform database operations concurrently. These threads access the database directly or via SSH tunnels (created with paramiko). I was experiencing the same issue (bad descriptor) and @qualab's PR fixed it.

I'll try to come up with a test code that generates this issue.

stromnet:
I'm having the same issues. I have tested both this fix and the one proposed in #1158, both works.

I'll keep my details in #1157 as that one is open.

PR #699, opened 2016-03-01
Prefer SHA1 over MD5 MACs

jjohnson42:
When running with FIPS restricted OpenSSL, md5 will error.  A number of
endpoints do sha1, but not sha2.  Reordering the preference allows the client
to connect from such clients.

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -112,10 +112,10 @@ class Transport (threading.Thread, ClosingContextManager):
     _preferred_macs = (
         'hmac-sha2-256',
         'hmac-sha2-512',
-        'hmac-md5',
         'hmac-sha1-96',
-        'hmac-md5-96',
         'hmac-sha1',
+        'hmac-md5',
+        'hmac-md5-96',
     )
     _preferred_keys = (
         'ssh-rsa',


bitprophet:
Thanks for the patch! This is tough; changing the order like this is backwards incompatible, meaning if I went for it, we'd want to put it in 2.0 or 3.0. And in 2.0 or 3.0 I'd like to merge #688 which removes MD5 HMAC entirely (since MD5 is now considered weak encryption).

I think I'm going to close on those counts - sorry!

jjohnson42:
No worries.  If you have a little more time to elaborate, interested in hearing about a backwards incompatible scenario.  One reason I did it this way was because I thought this would be nice and backwards compatible (still supporting the same hashes, just a slightly different preference of selecting it).  It should help me for the future.

bitprophet:
Sure - the main problem is existing users are used to a given agreed-upon set of algorithms between Paramiko and any given server of theirs; when we change the preferred order, that agreement can change, which is less than ideal generally, and can sometimes cause specific issues preventing operation entirely. (Usually due to misconfigurations or bugs on the server end, but still.)

I distinctly remember at least a couple tickets surrounding this, but naturally I can't find them now :(

Out of curiosity I looked at how OpenSSH handles this (though we only follow them for behavioral compatibility, not "how to release software" :)) and they do seem to *sometimes* remove or change ciphers in minor releases - but noted as "possibly backwards incompatible".

Which just comes back to the greater question of when to put in backwards incompatible changes - I've always tried to make sure that if someone was on any 1.x version of $lib, they can safely upgrade to any other 1.x without anything surprising changing underneath them. Other maintainers obviously differ.

----

I do want to note here that 3.0 is planned to happen within the next month or so (with 2.0 coming today but just being a crypto backend swap, no behavioral changes planned), so the core issue in your case - forcing use of SHA1 over MD5 - would still be solved once #688 is merged/released at that time.

PR #700, opened 2016-03-01
Adding key_search_path to SSHClient

ashafer01:
The intent here is to add /root/.ssh to the key_search_path when a program utilizing this class is being run with sudo.


ploxiln:
If that's a problem for you, run can run sudo with the `-H` option so the home folder is set to root's home folder. (You can also specify the specific key e.g. `/root/.ssh/id_rsa` as easily as specifying the search path.)

PR #685, opened 2016-02-10
Fixes NoValidConnectionsError on python3

winstonewert:
Fixes #678

PR #702, opened 2016-03-09
python3 fixup

mariodsantana:
In python3, dict.keys() returns a view instead of a list - which is iterable
    but not indexable.  Fixed by generating a list from the iterable.


mariodsantana:
Nevermind, looks like this will be fixed by pull request #685, sorry!

PR #712, opened 2016-03-24
fixes #387 and unblocks my work on gerritbot

### LINKED ISSUE #387: Key handling is terribad

cjac:
I tried to find a way to recognize ed25519 keys as well, but it looks like this pubkey algo is not even supported in gnutls yet.  So I don't feel so bad.  This does not handle putty-formatted keys or such things.  I've got some more changes ready, but this one meets my needs.

cjac@hype0:/usr/src/deb/openssh-7.2p2$ for keyfile in ~/.ssh/id_*.test ; do certtool -k --infile ${keyfile} | grep Algo ; done
	Public Key Algorithm: DSA
	Public Key Algorithm: EC
import error: Error in parsing.
	Public Key Algorithm: RSA
cjac@hype0:/usr/src/deb/openssh-7.2p2$ head -1 ~/.ssh/id_*.test
==> /home/cjac/.ssh/id_dsa.test <==
-----BEGIN DSA PRIVATE KEY-----

==> /home/cjac/.ssh/id_ecdsa.test <==
-----BEGIN EC PRIVATE KEY-----

==> /home/cjac/.ssh/id_ed25519.test <==
-----BEGIN OPENSSH PRIVATE KEY-----

==> /home/cjac/.ssh/id_rsa.test <==
-----BEGIN RSA PRIVATE KEY-----

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -23,6 +23,7 @@ SSH client & key policies
 from binascii import hexlify
 import getpass
 import os
+import re
 import socket
 import warnings
 from errno import ECONNREFUSED, EHOSTUNREACH
@@ -44,6 +45,7 @@ from paramiko.util import retry_on_signal, ClosingContextManager
 
 
 class SSHClient (ClosingContextManager):
+
     """
     A high-level representation of a session with an SSH server.  This class
     wraps `.Transport`, `.Channel`, and `.SFTPClient` to take care of most
@@ -142,7 +144,8 @@ class SSHClient (ClosingContextManager):
         with open(filename, 'w') as f:
             for hostname, keys in self._host_keys.items():
                 for keytype, key in keys.items():
-                    f.write('%s %s %s\n' % (hostname, keytype, key.get_base64()))
+                    f.write('%s %s %s\n' %
+                            (hostname, keytype, key.get_base64()))
 
     def get_host_keys(self):
         """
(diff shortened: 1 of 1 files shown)


mchlt:
Seems this is already covered by #618 "Support for new OpenSSH 6.5+ private key format" ?

cjac:
I don't think so.  Let me try a pip install from your branch...

bitprophet:
GH doesn't scan links in subject lines so just spamming #387 here so the link appears there. Thanks! Will try to poke this before I dive into #387 on my own.

PR #716, opened 2016-03-31
Change raw_input() to input() for Python 3.x

mkowalski:
When using Python 3.x and connecting via `paramiko.SSHClient().connect(host, port, user, pkey)` to the server which forces two-factor authentication, function `auth_interactive_dumb` from `transport.py` is being invoked, where `raw_input()` is called. Of course it does not work in Python 3.x, therefore I'm changing it to `input()` and adding additional check for backwards compatibility with Python 2.x

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -1395,7 +1395,10 @@ class Transport (threading.Thread, ClosingContextManager):
                     print(instructions.strip())
                 for prompt,show_input in prompt_list:
                     print(prompt.strip(),end=' ')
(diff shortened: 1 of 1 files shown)


mkowalski:
Please note the CI build error is not related to the change

bitprophet:
Weird, I guess github doesn't honor 'closes' if it's not 1st on a line? Anyway, cherry picked this back to 1.16. thanks! (ps - for future reference, most things like this have hooks in `paramiko.py3compat` :))

PR #725, opened 2016-04-18
Add support for ECDSA key sizes 384 and 521 alongside the existing 256.

CrazyCasta:
Previously only 256-bit was handled and in certain cases (private key
reading) 384- and 521-bit keys were treated as 256-bit keys causing
silent errors.

Tests have been added to specifically test the 384 and 521 keysizes. As
RFC 5656 defines 256, 384, and 521 as the required keysizes this seems a
good set to test. Also, this will cover the branches at ecdsakey.py:51.
It also tests all the OIDs in the ECDSA curve set at ecdsakey.py:87. Test
keys were renamed and test_client.py was modified as a result.


bitprophet:
This seems like a duplicate of #611.

PR #727, opened 2016-04-19
Add support for PyCryptodome (and PyPy)

Legrandin:
Shamelessly resubmitting this pull request (was #646).
Maybe it is acceptable at least for a 1.x branch (not 2.x)?

===

This patch makes paramiko work with PyCryptodome instead of PyCrypto.
With the former, paramiko can run under PyPy (and way more easily on Windows!).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
   - "3.3"
   - "3.4"
   - "3.5"
+  - "pypy"
 install:
   # Self-install for setup.py-driven deps
   - pip install -e .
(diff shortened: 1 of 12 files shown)


bitprophet:
The Cryptography.io merge is Actually Happening For Real™ and a 2.0 using it will be out within the next day or two. See #394's more recent activity :)

I'd be willing to hear discussion in the case where Cryptography.io was somehow harder to install than PyCryptodome for some users, but I'd guess that's not the case given the attention PyCA has put into their packaging!

Regardless, thanks for putting in the effort on this & apologies for the earlier foot-dragging on #394.

PR #728, opened 2016-04-23
Enabling caching of pip downloads and wheels

alex:


bitprophet:
Thanks! Cherry-picked back to 1.15 just for shits n giggles.

PR #729, opened 2016-04-23
Simplify setup.py by always relying on setuptools

alex:
It is installed everywhere now, there is no practical way to use python without it.


dstufft:
+1, this is recommended practice now adays.

bitprophet:
God bless code more than a decade old!

bitprophet:
Backported to 1.15.x

PR #731, opened 2016-04-25
Add support for ECDSA key sizes 384 and 521 alongside the existing 256.

CrazyCasta:
Previously only 256-bit was handled and in certain cases (private key
reading) 384- and 521-bit keys were treated as 256-bit keys causing
silent errors.

Tests have been added to specifically test the 384 and 521 keysizes. As
RFC 5656 defines 256, 384, and 521 as the required keysizes this seems a
good set to test. Also, this will cover the branches at ecdsakey.py:55.
Test keys were renamed and test_client.py was modified as a result.

This also fixes two bugs in ecdsakey.py. First, when calculating bytes
needed to store a key, the assumption was made that the key size (in
bits) was divisible by 8 (see line 137). This has been fixed by rounding
up (wasn't an issue as only 256-bit keys were used before). Another bug
was that the key padding in asbytes was being done backwards (was
padding on current_length - needed_length bytes).


bitprophet:
Thanks for this!

Is it feasible for you to split out the bugfixes to another PR (based on the 1.15 branch)? It'd be nice to make those available to folks not on the upcoming 2.x line.

CrazyCasta:
The bugs that I refer to don't exist prior to the switch to cryptography.

bitprophet:
Ah, that wasn't clear from your description, sorry. Works for me then, will get on merging this post-haste.

bitprophet:
Confirmed merge OK, tests pass locally, did a quick edit/format pass.

Need to write up a changelog entry, which I will do once I square bitprophet/releases#45.

paramiko/ecdsakey.py
```diff
@@ -57,41 +116,53 @@ def __init__(self, msg=None, data=None, filename=None, password=None,
...
+            key_bytes = (curve.key_size + 7) // 8
             numbers = ec.EllipticCurvePublicNumbers(
-                x=inflate_long(pointinfo[1:1 + curve.key_size // 8], always_positive=True),
-                y=inflate_long(pointinfo[1 + curve.key_size // 8:], always_positive=True),
+                x=inflate_long(pointinfo[1:1 + key_bytes],
+                               always_positive=True),
+                y=inflate_long(pointinfo[1 + key_bytes:],
+                               always_positive=True),
                 curve=curve
             )
```
alex: If your'e willing to depend on a recent pyca/cryptography, this mess can be replace with `ec.EllipcticCurvePublicNumbers.from_encoded_point(curve, pointinfo)`. I can send a PR to do that separately if you'd like.

alex: (Where recent means "released 6 months ago")

bitprophet: @alex Given we have no public releases relying on pyca/cryptography yet (& assuming it's always trended towards "easier to install" over time :D) I'm totally fine with modernizing the requires line in setup.py to anything stable.

alex: PR sent #733

bitprophet:
@alex, @CrazyCasta - I just merged #733 into master, then merged that new master into a copy of this here PR, and 4 of the pkey tests are failing (all ones using ecdsakey.py of course).

My merge commit looked pretty innocuous (just accounting for how alex replaced a small chunk w/ new crypto.io-using code) so I don't *think* it was that, but not sure.

See https://travis-ci.org/paramiko/paramiko/builds/126447338 - those are the same 4 errors I get locally.

Presumably the easiest way forward is for @CrazyCasta to update this branch vs latest master so the tests pass. But if @alex figures it out first I can apply whatever patch he generates & merge my integration branch to master, which should get us over the hump.

(Minor quirk: my local copy of master is head of upstream, but it's all just changelog crap that I don't want hitting the website yet.)

alex:
@bitprophet the problem is https://github.com/paramiko/paramiko/compare/731-int#diff-00344f8a080e03112e93210cdfd37302R134, that should be `s/ec.SECP256R1()/self.ecdsa_curve.curve_class`.

reaperhulk:
@alex We want an instance so it should be `self.ecdsa_curve.curve_class()` right?

alex:
Uhh, yeah.

On Thu, Apr 28, 2016 at 3:45 PM, Paul Kehrer <notifications@github.com>
wrote:

> @alex <https://github.com/alex> We want an instance so it should be
> self.ecdsa_curve.curve_class() right?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/paramiko/paramiko/pull/731#issuecomment-215540908>
>



-- 
"I disapprove of what you say, but I will defend to the death your right to
say it." -- Evelyn Beatrice Hall (summarizing Voltaire)
"The people's good is the highest law." -- Cicero
GPG Key fingerprint: D1B3 ADC0 E023 8CA6

bitprophet:
Cool, that seems to do the trick, thanks guys!

alex:
👍

bitprophet:
Merged to my local master, will go up once I finish the changeloggy junk.

PR #733, opened 2016-04-28
Use a more recent cryptography and call a function

alex:


alex:
Lowering coverage by deleting code, my specialty.

bitprophet:
Thanks! Also, I really need to replace coveralls with codecov, ISTR it being a lot more useful (Invoke is using it on its repo).

bitprophet:
Note to self, will need to re-merge #731 after merging this probably.

PR #754, opened 2016-05-21
Implement RFC 4252 public key query (Huawei CloudEngine series support)

AlexanderChernikov:
[RFC 4252](https://tools.ietf.org/html/rfc4252#section-7) section 7 describes 2 ways of performing "pubkey" auth.

1. sending publickey with signed data immediately.
2.  _querying whether authentication using the "publickey" method would be acceptable._ for the particular key prior to actually try to send signed data.

paramiko implements the former, while openssh implements the latter.
Huawei VRP software (powers all [CloudEngine series](http://e.huawei.com/en/products/enterprise-networking/switches/data-center-switches) switches) works fine with openssh but fails with paramiko:
>
DEBUG:paramiko.transport:Trying SSH agent key 56104577c38ecfca2c38a885a68a9648
DEBUG:paramiko.transport:userauth is OK
INFO:paramiko.transport:Disconnect (code 2): The connection is closed by SSH Server

While vendor implementation behavior doesn't seem to be valid (so I'll submit a bug to them), it may be worth adding support for this key query scenario. Again, openssh uses it by default for non-encrypted keys (see [sshconnect2.c:userauth_pubkey()](https://github.com/openssh/openssh-portable/blob/master/sshconnect2.c#L1438)).

This patch makes the change optional (and off by default) by providing check_keys parameter to both .connect() methond and SSHClient constructor.

### DIFF
--- a/paramiko/auth_handler.py
+++ b/paramiko/auth_handler.py
@@ -58,6 +58,7 @@ class AuthHandler (object):
         self.banner = None
         self.password = None
         self.private_key = None
+        self.check_key = False
         self.interactive_handler = None
         self.submethods = None
         # for server mode:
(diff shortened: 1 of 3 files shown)


AlexanderChernikov:
The situation is a bit more interesting: openssh is able to login to the box without performing given pubkey query, so the real problem lies elsewhere. Despite the fact that this patch fixes login issue w/ Huawei SW, this looks like not the best way to fix the problem. Will investigate further and reopen.

Issue #758, opened 2016-05-31
Argument error on connect: int too long to convert


swohlerLL:
Attempting to perform a simple connection with IP address on local network, username and password yields the following error:
`ArgumentError: argument 2: <class 'OverflowError'>: int too long to convert`

I'm running this in the iPython interpreter, on a 64-bit machine with Anaconda as the base python environment:
```
Python 3.5.1 |Anaconda 2.5.0 (64-bit)| (default, Jan 29 2016, 15:01:46) [MSC v.1900 64 bit (AMD64)]
IPython 4.0.3 -- An enhanced Interactive Python.
```

The sequence of commands:
```
ssh = paramiko.SSHClient()
ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
ssh.connect(host, username=username, password=password)
```

The full stack trace (some identifying info commented out with ***):
```
C:\Users\***\Anaconda3\lib\site-packages\paramiko\client.py in connect(self, hostname, port, username, password, pkey, key_filename, timeout, allow_agent, look_for_keys, compress, sock, gss_auth, gss_kex, gss_deleg_creds, gss_host, banner_timeout)
    378             gss_host = hostname
    379         self._auth(username, password, pkey, key_filenames, allow_agent,
--> 380                    look_for_keys, gss_auth, gss_kex, gss_deleg_creds, gss_host)
    381
    382     def close(self):

C:\Users\***\Anaconda3\lib\site-packages\paramiko\client.py in _auth(self, username, password, pkey, key_filenames, allow_agent, look_for_keys, gss_auth, gss_kex, gss_deleg_creds, gss_host)
    528         if not two_factor and allow_agent:
    529             if self._agent is None:
--> 530                 self._agent = Agent()
    531
    532             for key in self._agent.get_keys():

C:\Users\***\Anaconda3\lib\site-packages\paramiko\agent.py in __init__(self)
    361             # no agent support
    362             return
--> 363         self._connect(conn)
    364
    365     def close(self):

C:\Users\***\Anaconda3\lib\site-packages\paramiko\agent.py in _connect(self, conn)
     64     def _connect(self, conn):
     65         self._conn = conn
---> 66         ptype, result = self._send_message(cSSH2_AGENTC_REQUEST_IDENTITIES)
     67         if ptype != SSH2_AGENT_IDENTITIES_ANSWER:
     68             raise SSHException('could not get keys from ssh-agent')

[...]

PR #763, opened 2016-06-04
fix 64 bit issue with user SID

birm:
force user SID to be cast as long, brought up in #758
I haven't tested this on 32 bit environments, so I'd do that before merging.


birm:
I was looking at the issue in #758

birm:
I'll close and patch to jaraco.windows if I find anything to fix for this

Issue #741, opened 2016-05-09
PKey._write_private_key uses old identifier from cryptography


virlos:
Cryptography changed BestEncryption to BestAvailableEncryption in https://github.com/pyca/cryptography/commit/199dc276cd1b45a799b511090b37237df49d68a3#diff-f567fb1e66bd7ded34d4717d684339b7R188.

This results in a failure to write the key to a file with paramiko 2.0.0:

> > > import paramiko
> > > from StringIO import StringIO as S
> > > key = paramiko.RSAKey.generate(bits=1024)
> > > key.write_private_key(S(), password='hello')
> > > Traceback (most recent call last):
> > >   File "<stdin>", line 1, in <module>
> > >   File "/usr/local/lib/python2.7/dist-packages/paramiko/rsakey.py", line 142, in write_private_key
> > >     password=password
> > >   File "/usr/local/lib/python2.7/dist-packages/paramiko/pkey.py", line 345, in _write_private_key
> > >     encryption = serialization.BestEncryption(password)
> > > AttributeError: 'module' object has no attribute 'BestEncryption'
> > > from cryptography.hazmat.primitives import serialization as s
> > > dir(s)
> > > ['BestAvailableEncryption', 'Encoding', 'Enum', 'KeySerializationEncryption', 'NoEncryption', 'PrivateFormat', 'PublicFormat', 'UnsupportedAlgorithm', '__builtins__', '**doc**', '**file**', '**name**', '**package**', '_load_ssh_dss_public_key', '_load_ssh_ecdsa_public_key', '_load_ssh_rsa_public_key', '_read_next_mpint', '_read_next_string', 'abc', 'absolute_import', 'base64', 'division', 'dsa', 'ec', 'load_der_private_key', 'load_der_public_key', 'load_pem_private_key', 'load_pem_public_key', 'load_ssh_public_key', 'print_function', 'rsa', 'six', 'struct', 'utils']
> > > paramiko.__version__
> > > '2.0.0'
> > > import cryptography
> > > cryptography.**version**
> > > '1.3.2'

bitprophet:
Thanks, not sure how this got overlooked unless it's related to 1e7849ba910365e6008fd276b271e90a718fd618 (but the commit you linked is apparently present in crypto 0.8 and up, so even our old requirement should have been API consistent...)

bitprophet:
[...]

PR #772, opened 2016-07-08
Fix two issues with writing private key with passphrase

cjh1:
See #568

### DIFF
--- a/paramiko/pkey.py
+++ b/paramiko/pkey.py
@@ -342,7 +342,7 @@ class PKey(object):
         if password is None:
             encryption = serialization.NoEncryption()
         else:
-            encryption = serialization.BestEncryption(password)
+            encryption = serialization.BestAvailableEncryption(password.encode())
 
         f.write(key.private_bytes(
             serialization.Encoding.PEM,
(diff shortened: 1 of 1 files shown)


paramiko/pkey.py
```diff
@@ -342,7 +342,7 @@ def _write_private_key(self, f, key, format, password=None):
         if password is None:
             encryption = serialization.NoEncryption()
         else:
-            encryption = serialization.BestEncryption(password)
+            encryption = serialization.BestAvailableEncryption(password.encode())
```
alexpilotti: The method does not specify if the "password" parameter is bytes or string. If it's not documented elsewhere, wouldn't it be better to expect bytes in input? This affects python3.

alexpilotti:
This fixes issue #741

paramiko/pkey.py
cjh1: All the public methods in this class that take a password are expecting strings, so I think this password parameter should also be a string. Should I add a doc string to indicate this?

bitprophet:
Rolling into #741.

cjh1:
So have these fixes been merged?

ploxiln:
yes, they'll be in the next 2.y.z releases

bitprophet:
Been laid low today but yea I'm planning to release all this stuff as soon as I can. May toss in some low hanging fruit from the bugfix milestone too, but there's already so much ready to go out I don't absolutely have to.

cjh1:
Excellent! Many Thanks

PR #773, opened 2016-07-12
allow ~ expansion inside proxycommand

qqo:
Hi paramiko,

this supertrivial patch adds support for ssh_config like this:

```
Host test
    Hostname        %h.domain.com
    ProxyCommand    ssh -F ~/.ssh/test_config bastion nc %h %p
```

so instead of looking for literal `~/` path it'll be nicely expanded to `/home/myuser/`.

### DIFF
--- a/paramiko/config.py
+++ b/paramiko/config.py
@@ -209,6 +209,7 @@ class SSHConfig (object):
                         ],
                         'proxycommand':
                         [
+                            ('~', homedir),
                             ('%h', config['hostname']),
                             ('%p', port),
                             ('%r', remoteuser)
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -248,6 +248,19 @@ Host *
                 val
             )
 
+    def test_11_proxycommand_tilde_expansion(self):
+        """
+        Tilde (~) should be expanded inside ProxyCommand
+        """
+        config = paramiko.util.parse_ssh_config(StringIO("""
+Host test
+    ProxyCommand    ssh -F ~/.ssh/test_config bastion nc %h %p
+"""))
+        self.assertEqual(
+            'ssh -F %s/.ssh/test_config bastion nc test 22' % os.path.expanduser('~'),
+            host_config('test', config)['proxycommand']
+        )
+
     def test_11_host_config_test_negation(self):
         test_config_file = """
 Host www13.* !*.example.com


bitprophet:
Tentative :+1:, but do you know offhand if this is how OpenSSH behaves too? We usually try to mimic their behavior. I'd guess 'yes' but no time right now to go check. Thanks!

bitprophet:
Also, tests would be great :D

qqo:
Hey Jeff,

I use openssh-client 7.2p2-5 as of now, and been using "~" in ssh_config since some early 6.x and it was expanded to the homedir, yes. So I guess this patch adds what is in openssh already, but if you need me to do more serious check on that, please point me in the right direction and I'll try to do it.

Added test for that, too, checked with `./test.py test_11_proxycommand_tilde_expansion`. Dunno why coverage bot is not happy with it though, maybe I forgot to register it somewhere?

radssh:
As to how OpenSSH handles this, the ssh_config man page section on ProxyCommand states:

`The command string extends to the end of the line, and is executed with the user’s shell.  In the command string, ‘%h’ will be substituted by the host name to connect and ‘%p’ by the port.
`
(%r was added to a later version of OpenSSH then the one I grabbed the man page from). So the ~ expansion is actually done by the user's shell, not ssh.

A possible alternative to this patch might be to change the ProxyCommand Popen call to include "shell=True" to theoretically catch a broader range of potential shell-isms, not just ~ expansion. Balance that advantage against the Python docs strongly discouraging the usage of shell=True on potentially untrusted command line input... Food for thought.

bitprophet:
@radssh Yea, excellent point on the `shell=True`, it's the obvious/easy fix & also checks the "do what OpenSSH does" box, but also not amazing security practices.

A counterpoint is that in many other libraries I use and/or maintain that do shell things, `shell=True` is typically still used by default, as the convenience largely outweighs the security concerns in this instance.

Re: this specific ticket, @qqo's change is pretty darn small (just reuses previously-obtained homedir value) so I'd probably opt for that first, and punt the `shell=True` discussion until later. (Tho I'd bet there are other open tickets that discuss it or would be helped by it, already; no time to search right now unfortunately.)

@qqo - thanks! no need to dig further. And don't worry about coveralls, it's not very helpful, I still need to try replacing it with codecov sometime :(

I have assigned this to the next bugfix release so it'll get love next time I release. Thanks!

radssh:
Yeah, was not lobbying too hard for that option, just adding info to the issue on the off chance that more reports trickle in about "ProxyCommand doesn't work with <blah>" each gets handled as a one-off. If its been this long before the lack of ~ support got reported, it's probably not ever going to become a big deal, but you never know.

bitprophet:
Cherry-picked this into a new 1.18 feature release branch, to release soon.

PR #784, opened 2016-07-22
Exclude _winapi from coverage

jaraco:
Consider this technique or similar (I haven't tested this) to exclude the _winapi.py from coverage, as it's covered by jaraco.windows instead.


bitprophet:
Good catch, makes sense to me. Thanks!

Issue #789, opened 2016-07-30
ProxyCommand objects need the socket-like-obj _closed fix too


bitprophet:
See https://github.com/paramiko/paramiko/issues/774#issuecomment-236374022 ; also commutative to #520.

bitprophet:
Interestingly, `ProxyCommand` doesn't appear to expose *any* 'closed' attributes whatsoever - no `closed`, no `_closed`, etc.

It only implements `ClosingContextManager` which calls `self.close`, but no state is ever tracked for whether it's closed or not.

Since it wraps a `subprocess.Popen` I think the "most correct" analogue to "is the socket closed" would be `if process.returncode is not None`?

bitprophet:
@nvgoldin If possible, please try cherry-picking or just applying https://github.com/paramiko/paramiko/commit/228ed87e2f4b7314e4cfb67ee462550c5f20edef to your local Paramiko. It's pretty basic so I think it'll work, but I haven't actually tested it myself yet. Hope to later.

nvgoldin:
The exception is gone, thanks!!
But, there is another problem which I am not sure is directly related to this(tell me if to open another issue). Seems like the proxy command process isn't killed properly and it leaves zombie processes, running the following for several loops:
```python
import paramiko
import time
import logging
import os
ssh_host='localhost'
proxy_cmd='ssh -o StrictHostKeyChecking=no -W localhost:22 localhost'
logging.basicConfig(level=logging.DEBUG)
while True:
    logging.debug('running PID %s', os.getpid())
    ssh = paramiko.SSHClient()
    ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
    ssh_proxy = paramiko.ProxyCommand(proxy_cmd)
    ssh.connect(sock=ssh_proxy, hostname=ssh_host)
    sftp = ssh.open_sftp()
    sftp.close()
    ssh.close()
    ssh_proxy.close()
    logging.debug('going to sleep')
    time.sleep(3)
```
Results in:  (`20125` is the program pid):
```
>ps  xao pid,ppid,pgid,sid,comm | grep 20125

20126 20125 20125 15530 ssh <defunct>
20169 20125 20125 15530 ssh <defunct>
20228 20125 20125 15530 ssh <defunct>
20271 20125 20125 15530 ssh <defunct>
20330 20125 20125 15530 ssh <defunct>
20373 20125 20125 15530 ssh <defunct>
20416 20125 20125 15530 ssh <defunct>
[...]

PR #795, opened 2016-08-03
add _closed field to proxy command

xmonader:


bitprophet:
Thanks for the report! Unfortunately this seems to be a duplicate of issue #789 . Please make sure to use the search feature in the future :)

PR #800, opened 2016-08-23
Fixes cryptography compatibility

alexpilotti:
Writing and reading password protected private keys fails with
cryptography >= 0.8.

This patch fixes the issue by updating changed class names and
adding support for AES-256-CBC.

Closes bug #741


alexpilotti:
There's already a fix proposed for this issue at #772 with the same solution, so nevermind :)

bitprophet:
Thanks for the updates here & elsewhere @alexpilotti - #741 is still in a release milestone so I'll get to it (& #772) eventually!

PR #809, opened 2016-08-30
Change  BestEncryption to BestAvailableEncryption

ibuler:
cryptography.hazmat.primitives.serialization has no class name BestEncryption

### DIFF
--- a/paramiko/pkey.py
+++ b/paramiko/pkey.py
@@ -342,7 +342,7 @@ class PKey(object):
         if password is None:
             encryption = serialization.NoEncryption()
         else:
-            encryption = serialization.BestEncryption(password)
+            encryption = serialization.BestAvailableEncryption(password)
 
         f.write(key.private_bytes(
             serialization.Encoding.PEM,
--- a/paramiko/pkey.py
+++ b/paramiko/pkey.py
@@ -342,8 +342,10 @@ class PKey(object):
         if password is None:
             encryption = serialization.NoEncryption()
         else:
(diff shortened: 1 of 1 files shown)


bitprophet:
Unfortunately this will break things for anybody using Cryptography 1.4! :(

I would suggest using a try/except to 'cascade' this so it first tries `BestAvailableEncryption`, then falls back to `BestEncryption`. That should mean it works on either version of Cryptography. Thanks!

bitprophet:
Note to self: we'll probably also want to tweak our `.travis.yml` so it has test matrix cells for different Cryptography versions...

ibuler:
Add try except , but I see Cryptography have no version use BestEncryption after 1.0

bitprophet:
Putting in 2.1 but it might be appropriate for a 2.0.x bugfix, depending. Will have to think about it when I get to it.

ploxiln:
I think `BestEncryption` was always a typo. In cryptography-0.8 it was `BestAvailableEncryption` and in cryptography-0.7 neither existed as far as I can tell.

ploxiln:
Found link in another issue :) https://github.com/pyca/cryptography/commit/199dc276cd1b45a799b511090b37237df49d68a3#diff-f567fb1e66bd7ded34d4717d684339b7R188

`BestEncryption` was never in a release, it was renamed in the same pull request it was introduced in

bitprophet:
Huh, weird. And I no longer recall what gave me the impression re: 1.4.x.

bitprophet:
Rolling into #741 / #912.

PR #811, opened 2016-08-31
Change ProxyCommand.close to use Popen.close/poll

nvgoldin:
This fixes the 'zombie process' left when using ProxyCommand. Popen.poll() is needed to update the return code.

From the [docs](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.poll):

> Popen.poll()
>    Check if child process has terminated. Set and return returncode attribute.

@bitprophet - I based this on https://github.com/paramiko/paramiko/commit/228ed87e2f4b7314e4cfb67ee462550c5f20edef, fixes https://github.com/paramiko/paramiko/issues/789

### DIFF
--- a/paramiko/proxy.py
+++ b/paramiko/proxy.py
@@ -105,7 +105,8 @@ class ProxyCommand(ClosingContextManager):
             raise ProxyCommandFailure(' '.join(self.cmd), e.strerror)
 
     def close(self):
-        os.kill(self.process.pid, signal.SIGTERM)
+        self.process.kill()
+        self.process.poll()
 
     @property
     def closed(self):


bitprophet:
See comments on #789, I was unable to reproduce the zombies @nvgoldin mentions there, but I'm wondering if it's Linux specific or if other factors are involved. Moving that discussion here since #789 itself is merging.

nbartos:
This problem shouldn't be Linux specific (although I don't have a non-Linux system to test).  In UNIX whenever a called process exits, the parent has to wait() for it to clear the zombie.  From the wait man page (https://linux.die.net/man/2/wait):

```
Conforming To

SVr4, 4.3BSD, POSIX.1-2001.

Notes

A child that terminates, but has not been waited for becomes a "zombie". The kernel maintains a minimal set of information about the zombie process (PID, termination status, resource usage information) in order to allow the parent to later perform a wait to obtain information about the child. As long as a zombie is not removed from the system via a wait, it will consume a slot in the kernel process table, and if this table fills, it will not be possible to create further processes. If a parent process terminates, then its "zombie" children (if any) are adopted by init(8), which automatically performs a wait to remove the zombies.
```

I'd really like to get this fixed one way or another, as it's a big problem in an application I have.  I opened up https://github.com/paramiko/paramiko/pull/962 before I found this one.  Either one should be fine, as long as one gets merged.

paramiko/proxy.py
```diff
@@ -105,7 +105,8 @@ def recv(self, size):
             raise ProxyCommandFailure(' '.join(self.cmd), e.strerror)
 
     def close(self):
-        os.kill(self.process.pid, signal.SIGTERM)
+        self.process.kill()
+        self.process.poll()
```
nbartos: This should be "self.process.wait()".  In my tests, poll() doesn't' remove the zombie (which makes sense if you read the man page (2) for wait()).

paramiko/proxy.py
```diff
@@ -105,7 +105,8 @@ def recv(self, size):
             raise ProxyCommandFailure(' '.join(self.cmd), e.strerror)
 
     def close(self):
-        os.kill(self.process.pid, signal.SIGTERM)
+        self.process.kill()
```
nbartos: I found it's good to wrap this kill in a try block (see https://github.com/paramiko/paramiko/pull/962), as otherwise the program calling paramiko will traceback if something else killed the process.

paramiko/proxy.py
ploxiln: I think `poll()` does clean up zombie processes, it just doesn't wait for the process if it hasn't exited yet. It's like libc `wait()` with `WNOHANG`. So in this case it was "too soon", it needed to check if the process did exit, and if not, try again.

avoidik:
For the reference solution for this problem was provided with #962

PR #820, opened 2016-09-25
Prevent SFTP get() from emptying files (fixes #724)

### LINKED ISSUE #724: sftp_client.get() creates empty file when sftp_client is closed
```
>>> import paramiko
>>> client = paramiko.SSHClient()
>>> client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
>>> client.connect("192.168.50.3", username="vagrant", password="vagrant", port=22)
>>> sftp = client.open_sftp()
>>> sftp.get(".bashrc", "downloaded/.bashrc")
```
Checking that file indeed was downloaded (another terminal tab):
```
$ ls -la downloaded/
total 8
drwxr-xr-x   3 dmitry.tokarev  US\Domain Users  102 Apr 18 11:25 .
drwxr-xr-x  29 dmitry.tokarev  US\Domain Users  986 Apr 18 11:21 ..
-rw-r--r--   1 dmitry.tokarev  US\Domain Users  231 Apr 18 11:25 .bashrc
```
Now close sftp and try to `get()` the file again:
```
>>> sftp.get(".bashrc", "downloaded/.bashrc_new")
  File "/Library/Python/2.7/site-packages/paramiko/sftp_client.py", line 721, in get
    size = self.getfo(remotepath, fl, callback)
  File "/Library/Python/2.7/site-packages/paramiko/sftp_client.py", line 689, in getfo
    file_size = self.stat(remotepath).st_size
  File "/Library/Python/2.7/site-packages/paramiko/sftp_client.py", line 413, in stat
    t, msg = self._request(CMD_STAT, path)
  File "/Library/Python/2.7/site-packages/paramiko/sftp_client.py", line 729, in _request
    num = self._async_request(type(None), t, *arg)
  File "/Library/Python/2.7/site-packages/paramiko/sftp_client.py", line 751, in _async_request
    self._send_packet(t, msg)
  File "/Library/Python/2.7/site-packages/paramiko/sftp.py", line 170, in _send_packet
    self._write_all(out)
  File "/Library/Python/2.7/site-packages/paramiko/sftp.py", line 133, in _write_all
    n = self.sock.send(out)
  File "/Library/Python/2.7/site-packages/paramiko/channel.py", line 698, in send
    return self._send(s, m)
  File "/Library/Python/2.7/site-packages/paramiko/channel.py", line 1058, in _send
    raise socket.error('Socket is closed')
socket.error: Socket is closed
>>> 
```
As expected we got *Socket is closed* is closed error.
Now let's check what happened with local filesystem:
```
$ ls -la downloaded/
total 8
drwxr-xr-x   4 dmitry.tokarev  US\Domain Users  136 Apr 18 11:29 .
drwxr-xr-x  29 dmitry.tokarev  US\Domain Users  986 Apr 18 11:21 ..
-rw-r--r--   1 dmitry.tokarev  US\Domain Users  231 Apr 18 11:25 .bashrc
-rw-r--r--   1 dmitry.tokarev  US\Domain Users    0 Apr 18 11:29 .bashrc_new
```
We got a new empty file `.bashrc_new`.
I understand that paramiko creates a local file to write to before opening a remote file, but i would expect that if sftp client is closed it would not create the empty file or it would remove it.

I get same results if i close "main" ssh client:

operatingops:
Fixes #724

The problem is that `getfo()` (the method that gets data from the SFTP server and writes it out) in `sftp_client.py` takes a file object, not a path, so `get()` (a convenience method that takes a path) opens the local file before it calls `getfo()`. Python's `open()` overwrites no matter what, even if it has no data to write. When `getfo()` errors the local file is overwritten with nothing. I added a test to cover this. I looked through the code's other uses of `open()` and didn't see any with this problem.

The bug was introduced when this line was removed: `file_size = self.stat(remotepath).st_size`. It wasn't used in the method, but if the connection was closed it caused an error before the `open()` call, which prevented the local file from being opened and then emptied. Adding it or something similar would be an interesting one-line fix, but I don't think it's the right way for two reasons:
1. I'd have to explain it in comments.
2. There's still a race condition where a connection dies because of network problems between the `stat()` call and the `open()` call. Although unlikely, it'd be almost impossible to reproduce.

I think the fix with the lowest impact is to modify `get()` to backup the local file and restore it if we get an error. Some details:
- I chose to backup and restore instead of downloading into a temp file so that if there are problems on the filesystem (permissions, etc.) it will fail before downloading, like it did before.
- This may not be atomic in all cases, especially on Windows. The original method wasn't atomic and I thought it was out of scope to address that while fixing #724. Making filesystem operations truly atomic in Python isn't straightforward.

After reading the contributing docs I used branch 1.16 since that's the oldest one with recent merges that has the bug. I tested the same change in master, though, just to make sure it'd work there. I'm a bit new to some of the details of the contributing process you use and if I used the wrong branch or anything just let me know and I'll fix it up.

I ran the tests with paramiko installed in editable mode in a virtualenv running Python 2 and a pyvenv running Python 3.

What do you think? Let me know if you think anything should be different and I'll rebase it in.

### DIFF
--- a/paramiko/sftp_client.py
+++ b/paramiko/sftp_client.py
@@ -717,11 +717,20 @@ class SFTPClient(BaseSFTP, ClosingContextManager):
         .. versionchanged:: 1.7.4
             Added the ``callback`` param
         """
-        with open(localpath, 'wb') as fl:
-            size = self.getfo(remotepath, fl, callback)
-        s = os.stat(localpath)
-        if s.st_size != size:
-            raise IOError('size mismatch in get!  %d != %d' % (s.st_size, size))
+        backuppath = '{0}.paramiko_bak'.format(localpath)
+        if os.path.isfile(localpath):
+            os.rename(localpath, backuppath)
+        try:
+            with open(localpath, 'wb') as fl:
+                size = self.getfo(remotepath, fl, callback)
+            s = os.stat(localpath)
+            if s.st_size != size:
+                raise IOError('size mismatch in get!  %d != %d' % (s.st_size, size))
+            if os.path.isfile(backuppath):
+                os.remove(backuppath)
+        except:
+            os.rename(backuppath, localpath)
+            raise
 
     ###  internals...
 
--- a/sites/www/changelog.rst
+++ b/sites/www/changelog.rst
@@ -2,6 +2,9 @@
 Changelog
 =========
 
+* :bug:`724 (1.16+)` (via :issue:`724`) Prevent SFTP get() method from emptying
+  files when called on closed connections. Thanks to ``@dmitrytokarev`` for the
+  report and Adam (``@operatingops``) for the patch.
 * :support:`819 backported (>=1.15,<2.0)` Document how lacking ``gmp`` headers
   at install time can cause a significant performance hit if you build PyCrypto
   from source. (Most system-distributed packages already have this enabled.)
--- a/tests/test_sftp.py
+++ b/tests/test_sftp.py
@@ -595,6 +595,7 @@ class SFTPTest (unittest.TestCase):
         """
         verify that get/put work.
         """
+        global sftp
         warnings.filterwarnings('ignore', 'tempnam.*')
 
         fd, localname = mkstemp()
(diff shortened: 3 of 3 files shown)


operatingops:
Weird, one of the tests failed in Travis but worked on my laptop. Can't look at this right now but will investigate and see what's up. Also will look at why coverage dropped

operatingops:
Turns out I wasn't using the test framework right and had missed a syntax requirement of Python 2.6. Rebased in the fix, which has also made the coverage increase instead of decrease.

singhjaideep:
FYI `file_size = self.stat(remotepath).st_size` does not work, if the SFTP server doesn't support "df -hi" command. Paramiko will return IOError.

bitprophet:
Not sure why this didn't show up in my feed originally...thanks for the submission!

The `stat` call removal was definitely on purpose, don't have the issue numbers handy but it was causing some (impossible to work around) edge case errors for some users. So putting it back is unlikely to be the right call; I think making a backup probably is the right approach, have to think about it a bit.

operatingops:
No worries @bitprophet - if there's any other investigation or testing you'd like to see just let me know

PR #824, opened 2016-10-06
open() function does not support setting file permissions

JohnVillalovos:
The open() function does not support setting file permissions[1]. The
code was setting the buffer to 0o600. Remove the attempted file
permission value.

[1] https://docs.python.org/2/library/functions.html#open

Change-Id: I582e226f8bc76e877c7da686f36e2efa33f0d24f

### DIFF
--- a/paramiko/pkey.py
+++ b/paramiko/pkey.py
@@ -333,8 +333,7 @@ class PKey(object):
 
         :raises IOError: if there was an error writing the file.
         """
-        with open(filename, 'w', o600) as f:
-            # grrr... the mode doesn't always take hold
+        with open(filename, 'w') as f:
             os.chmod(filename, o600)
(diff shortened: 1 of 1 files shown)


bitprophet:
Bizarre, as far as I can tell that should never have been correct, e.g. I spot checked Python 2.2 (Paramiko dates back that far) and it's the same signature as now, unsurprisingly.

Seems this has been in since the beginning (01bf5477a04cbb34974aae92f6c5965572800b63) and nobody ever noticed because it doesn't cause exceptions (that, and this method isn't called terrifically often by most users).

Thanks for the catch!

JohnVillalovos:
You're welcome. Glad to help!

bitprophet:
Cherry-picked to 1.17.

PR #827, opened 2016-10-13
Smartcard pkcs11 support

starboarder2001:
This adds pkcs11 support to enable using paramiko with smartcards.  I have a forked version of Ansible that uses this feature and its working great.

I have tested on python 2.7 and python 3.6.2

# Multithreading Example:

``` python
import paramiko
from multiprocessing import Queue
from threading import Thread

pkcs11provider="/usr/local/lib/opensc-pkcs11.so"
smartcard_pin="123456"

def do_it(q):
    session = q.get()
    ssh = paramiko.SSHClient()
    ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
    ssh.connect("HOSTNAME", username="USERNAME", pkcs11_session=session)
    stdin, stdout, stderr = ssh.exec_command("uname -a")
    for line in stdout:
        print(line)

q = Queue()
session = paramiko.pkcs11.open_session(pkcs11provider, smartcard_pin)
mythread = Thread(target=do_it, args=(q,))
q.put(session)
mythread.start()
mythread.join()
paramiko.pkcs11.close_session(session)
```
# Basic Example:

``` python
import paramiko

pkcs11provider="/usr/local/lib/opensc-pkcs11.so"
smartcard_pin="123456"

ssh = paramiko.SSHClient()
ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
session = paramiko.pkcs11.open_session(pkcs11provider, smartcard_pin)
ssh.connect("HOSTNAME", username="USERNAME", pkcs11_session=session)
paramiko.pkcs11.close_session(session)
stdin, stdout, stderr = ssh.exec_command("uname -a")
for line in stdout:
    print(line)
```


Ovomaltine85:
Hi, great work! I'm actually waiting for this feature.
Can you tell me when it becomes part of a new release of paramiko?
Thank you!

SethMichaelLarson:
You should probably add an entry to the change log. :)

bitprophet:
@starboarder2001 Yea, something in the `tests/` folder would be grand (hopefully there are existing tests in roughly the same area, re: `Client.connect` parameters), and docs, I think for now it'd be sufficient to update the docstring for `Client.connect` so it mentions that the new params exist and what they're for. (Typically, anytime you modify a function's API, you should be modifying the nearby docstring as well :))

Also what @SethMichaelLarson said, tho changelog is always something I can add myself (I am picky about language...)

Thanks!

bitprophet:
Also, you'll want to make sure the merge conflicts are fixed, we pep8'd the crap out of the codebase recently which may be what caused those.

starboarder2001:
I think its in pretty good shape. I updated the changelog, docstrings for Client.connect, and made sure the code passed all the pep8 checks.

- David

starboarder2001:
Trying to help out the best I can on this one, is there anything I can do to help this along?

ploxiln:
There was a big round of merging and releasing about a week ago (as you probably noticed), and it reduced the backlog noticeably. The maintainer will probably do another around in a few weeks. I think there is a decent chance this PR will eventually get in, if you keep it up-to-date :)

I think it would help a bit to squash down to just one commit (or a few). This will require a careful interactive rebase, re-resolving some merge conflicts, and a force-push. Since the "flake8 re-style" is behind us, I don't expect this much tedium being needed again.

paramiko/auth_handler.py
```diff
@@ -229,6 +271,74 @@ def _parse_service_request(self, m):
         # dunno this one
         self._disconnect_service_not_available()
 
+    def _pkcs11_get_public_key(self):
+        public_key = ""
+        if self.pkcs11session is None:
+            if which("pkcs15-tool") is None:
```
ploxiln: I think you should remove the `which()` function you added. Just catch the `subprocess.Popen()` exception for this case (an `OSError` with `.errno` equal to https://docs.python.org/2/library/errno.html#errno.ENOENT). Then you can re-throw your more specific `Exception("pkcs15-tool not found in PATH")` (I wouldn't include the extended instructions in the exception text).

paramiko/client.py
```diff
@@ -279,6 +282,12 @@ def connect(
             The targets name in the kerberos database. default: hostname
         :param float banner_timeout: an optional timeout (in seconds) to wait
             for the SSH banner to be presented.
+        :param str pkcs11pin: If using PKCS11, this will be the pin of your
+            token or smartcard.
+        :param str pkcs11provider: If using PKCS11, this will be the provider
+            for the PKCS11 interface. Example: /usr/local/lib/opensc-pkcs11.so.
+        :param str pkcs11session: If using PKCS11 in a multithreaded
+            application you can share the session between threads.
```
ploxiln: Can this be reduced to just one additional argument to `SSHClient.connect()`? Perhaps the user could be required to call `pkcs11_open_session()` themselves, then pass just the session to `connect()`?

ploxiln:
In some cases, where you raise a `Exception` (with explanation text), it might be better to use a more specific type from `paramiko.ssh_exception` e.g. `AuthenticationException`. (Each case would require some consideration, which I haven't done yet :)

paramiko/client.py
starboarder2001: For most cases the pkcs11provider is all thats needed, very few users will need to use the pkcs11session argument.  pkcs11session is optional and only needed when used in a multithreaded application.

paramiko/pkcs11.py
```diff
@@ -0,0 +1,115 @@
...
+        p = subprocess.Popen(["pkcs15-tool", "--read-ssh-key", "01"],
+                             stdout=subprocess.PIPE,
+                             stderr=subprocess.PIPE,
+                             stdin=subprocess.PIPE)
+        out, err = p.communicate()
+        if out is not None:
+            public_key = out
+    except OSError as error:
+        if error.errno == errno.ENOENT:
+            raise SSHException("Cannot find pkcs15-tool in PATH.")
```
ploxiln: If not translating to this exception, you still need to re-raise the original exception, e.g. just

```python
else:
    raise
```

starboarder2001: Thanks, good point! I'll do more review and testing tomorrow.

ploxiln:
https://github.com/paramiko/paramiko/pull/827/commits/73ef6e1ec1cdf5fe67a56817c79bae565364aed8 contains a bunch of changes from master. Merging with master hides this from the diff in the "Files Changed" tab, but does not fix the problem.

(When you squashed, you needed to ensure you did not include commits you did not author.)

starboarder2001:
I think the last rebase worked ok.

ploxiln:
yup, looks good, thanks

paramiko/auth_handler.py
```diff
@@ -243,13 +350,34 @@ def _parse_service_accept(self, m):
...
+                    blob = self._get_session_blob(
+                        self.private_key, 'ssh-connection', self.username)
+                    sig = self.private_key.sign_ssh_data(blob)
+                    m.add_string(sig)
+                else:
+                    # Smartcard PKCS11 Private Key
+                    keym = Message()
+                    keym.add_string('ssh-rsa')
+                    pkcs11_tmp_pub = load_ssh_public_key(
+                        self._pkcs11_get_public_key(),
```
ploxiln: Is the pkcs11 public key guaranteed to always be rsa?

starboarder2001: Thats a good point.  I don't think its a guarantee.

paramiko/auth_handler.py
```diff
@@ -243,13 +353,43 @@ def _parse_service_accept(self, m):
...
+                        key_name = 'ssh-rsa'
+                    elif isinstance(pkcs11_tmp_pub, DSAPublicKey):
+                        key_name = 'ssh-dss'
+                    elif isinstance(pkcs11_tmp_pub, EllipticCurvePublicKey):
+                        key_name = 'ssh-ecdsa'
+                    else:
+                        raise Exception("Unable to determine pkcs11 key type")
+                    keym.add_string(key_name)
+                    keym.add_mpint(pkcs11_tmp_pub.public_numbers().e)
+                    keym.add_mpint(pkcs11_tmp_pub.public_numbers().n)
```
ploxiln: If key types besides ssh-rsa are possible, it is a bit more complicated to support them - they do not have "e" and "n", they have different sets of defining values.

Paramiko's `RSAKey` has an `asbytes()` method which does exactly this. The other key types also have that method, to pack the appropriate Message for that type. So the way to do this would be similar to how `HostKeyEntry.from_line()` does it: https://github.com/paramiko/paramiko/blob/2.2.1/paramiko/hostkeys.py#L325

starboarder2001: Thanks for the hints.  I cleaned up the code quite a bit to use Paramiko's RSAKey (and other algorithms). I believe it will work now with algorithms other than rsa, but I really need to implement a few unit tests for this to verify its functioning correctly.

starboarder2001:
I cleaned up the support for multiple key algorithms. I could reuse the existing _get_session_blob method instead of creating one specific for pkcs11.

starboarder2001:
Hows it currently looking, closer to where it needs to be?

paramiko/client.py
```diff
@@ -518,6 +529,17 @@ def _auth(self, username, password, pkey, key_filenames, allow_agent,
         allowed_types = set()
         two_factor_types = set(['keyboard-interactive', 'password'])
 
+        # PKCS11 / Smartcard authentication
+        if username is not None and pkcs11pin is not None and \
+           pkcs11provider is not None:
+            try:
+                allowed_types = set(self._transport.auth_pkcs11(username,
+                                    pkcs11pin, pkcs11provider, pkcs11session))
+                if not two_factor:
```
ploxiln: The pattern in the other cases has a bit more:

```python
two_factor = (allowed_types & two_factor_types)
if not two_factor:
```

starboarder2001: Yep. Small one, but could have caused issues.  The latest push should have resolved it.

paramiko/auth_handler.py
```diff
@@ -20,8 +20,13 @@
 `.AuthHandler`
 """
 
+import threading
+import os
+from ctypes import (cdll, Structure, c_ulong, c_void_p, byref, c_char,
+                    c_int)
```
ploxiln: minor style nit: see below, this is usually indented instead of wrapped:

```python
from ctypes import (
    cdll, Structure, c_ulong, c_void_p, byref, c_char, c_int
)
```

paramiko/auth_handler.py
```diff
@@ -243,13 +343,51 @@ def _parse_service_accept(self, m):
...
+                        if keytype == 'ssh-rsa':
+                            pub_key = RSAKey(data=decodebytes(pub_key))
+                        elif keytype == 'ssh-dss':
+                            pub_key = DSSKey(data=decodebytes(pub_key))
+                        elif keytype in ECDSAKey\
+                            .supported_key_format_identifiers():
+                            pub_key = ECDSAKey(data=decodebytes(pub_key),
+                                               validate_point=False)
+                        elif keytype == 'ssh-ed25519':
+                            pub_key = Ed25519Key(data=decodebytes(pub_key))
```
ploxiln: I think this "if keytype == blah: key = blah(data)" pattern can be put in a helper function somewhere such that hostkeys.py could share use of it.

ploxiln:
I think this is looking pretty good. (The test failure on pypy is unfortunately a familiar spurious one, not your fault.)

You might have noticed that paramiko merging activity is done in spurts, every few months. So it's hard to tell when this might get in.

bitprophet:
Digging into this now, with intent to apply my own tweaks & merge it soon; thoughts as I go:

- Tests would be nice but probably a pain if they're even possible at all with current design.
  - In these situations I often at least try to verify functionality myself, but not sure I have the personal bandwidth to experiment with e.g. a YubiKey and the current state of apparent OpenSC support for macOS Sierra. We'll see though.
- Seeing a handful of minor-ish issues I can fix myself, such as the key identification bits having an apparent no-op line around `SSHException` (there's no `raise`!)
- Speaking of, I agree with @ploxiln about refactoring the "which key class to use" bits. This is actually related to #387 but I think it's feasible to at least prevent further spread of the pattern. Main question is where such code should live:
  - Can't go in `util` since `util` is used in the key files themselves...
  - Could go into `hostkeys.py` or `auth_handler.py`, though because the functionality is generic, it doesn't "really" want to live in either place, arguably.
  - Gut says just make a new file. Don't want to make even-more-files cuz we already have a ton (and someone looking for "key stuff" already has a lot of confusing places to look) but it may still be the best option.
    - Also useful for #387 re: starting to tease out this functionality to easier-to-test standalone classes; _and_ useful in the sense that much of this wants to eventually migrate to the Cryptography library or other external code.

ploxiln:
Yeah, new file might be the best of the awkward choices. Idea: "keytypes.py"

bitprophet:
Going with `authentication.py` for now, hoping we can eventually put some high level classes/funcs in here to handle all the wacky connection-time auth handling logic, and whatever else. Trying to very slowly make the codebase marginally more responsibility-based, `¯\_(ツ)_/¯`

Side note: I note that in `hostkeys.py` where we instantiate `ECDSAKey` (which was copied for this PR) we're passing an extra kwarg vs the other classes which are all the same: `validate_points=False`. This seems to have been done for #298; **however**, we migrated the ECDSA handling to Cryptography later on, which doesn't even seem to care about `validate_points` at all - it's completely unused.

And the other spot I found that can use this new module of mine - host key verification in `transport.py` - doesn't even care about that param, period, and may never have.

So, noting for the future that I am removing that particular kwarg from these other calls to the class constructor; it'll stay in the class itself for backwards compat but is now unused in our own code.

bitprophet:
Another difference in the code being consolidated: `auth_handler.py` raises `SSHException` if the remote host has an unknown key type; but `hostkeys.py` simply logs that error and returns `None`.

IIRC this is on purpose, e.g. so one can use `known_hosts` files containing "bad" (unknown-type) lines for their "good" contents, so I will keep the same overall behavior for now by having `hostkeys.py` trap the exception.

bitprophet:
_brushing off clouds of dust_ OK that refactoring is done, all tests (and fabric tests) pass, so the only place there ought to be errors is in the new auth_handler code branch around the new pkcs11 functionality. And I at least fixed the exception raising bit :D 

Going through the rest of the diff now and making any necessary style/etc tweaks, will report back w/ any qualitative changes.

bitprophet:
Also randomly wondering "surely other folks have done PKCS11 with Python", is there even a reason we're not simply deferring to something like https://github.com/LudovicRousseau/PyKCS11 or http://python-pkcs11.readthedocs.io/en/latest/?

Though the former's a C-ext and the latter uses Cython; so this PR's "you need to have a built OpenSC plus its `pkcs11-tool` lying around and we call it via `ctypes`" approach is at least a different set of tradeoffs...

That does bring me to another observation, there's literally no docs here: no "psst, you need to get OpenSC, here's how" or anything. Also no link to how the author learned how to implement the fiddly C/math bits; is it based on OpenSC's docs (I haven't looked) or directly on [what seems to be the PKCS#11 spec](http://docs.oasis-open.org/pkcs11/pkcs11-ug/v2.40/cn02/pkcs11-ug-v2.40-cn02.html) or other?

I'm going to turn my attention to the other 2.3.x milestone members for now and will check back for answers on the above later :) thanks! (I am still 99% likely to continue merging this, FWIW.)

bitprophet:
Re: docs: note that I'm OK writing final prose myself (& I'm not expecting it to be large, just a few paragraphs) but some basic pointers like "well here's what I've been doing on my end re: obtaining OpenSC or whatnot, and here's what my server has running such that it's using this client support" would be appreciated!

paramiko/auth_handler.py
bitprophet: Got this in my branch.

paramiko/auth_handler.py
bitprophet: Got this in my branch.

paramiko/client.py
bitprophet: I second @ploxiln's question - I think what he's getting at is it's unclear what the relationship is between these arguments & the data stored in the session - which, w/o doing a deep analysis, seems to largely be a 3-tuple whose members include the pin and provider anyways.

So if we can simplify the API on our end to "you should have called `pkcs11_open_session()` yourself, please hand me the resulting session obj/tuple" that cleans up a lot of intermediate code in an already-very-crowded space.

bitprophet:
Another final comment so I don't lose it: I'd really, really love the project to stop spawning additional, frustratingly useless `raise SSHException("actual message")` calls, effective immediately. It'd be great if we could identify the real classes of errors within this PR (in the 7 spots it does `raise SSHException`) and make a few (hopefully <7...) new subclasses of `SSHException` instead.

(This isn't a dig at the OP, following local style is always the right way to go by default! But new functionality like this is a great time to start enforcing changes to that style :))

starboarder2001:
I did look at some of the other PKCS11 python libraries, they seemed quite heavy handed for what needed to be done and I also didn't want to add any unnecessary dependencies.  I wouldn't mind taking that approach, but I think it would be beneficial to avoid adding dependencies.

  It currently depends on OpenSC as it uses pkcs15-tool to read the public ssh key, someone who wanted to put in more effort to implement this natively could remove this dependency.  OpenSC is the only option for a open source pkcs11 middleware, so I don't see an issue with specifying OpenSC as the only supported pkcs11 middleware.

  I could document that the OpenSC middleware is required and that pkcs15-tool should be in the system PATH.  Something I didn't consider is that the pkcs15-tool key id is hard coded to 01 (also hard coded in objid_str = "1"), its possible someone could have a use case where there are multiple keys.  I could add pkcs11tokenid, and make the default "01" since its likely what 90% of users will use.  I think it will work for most users as is, but its something that will eventually come up.

  I will take note of the SSHExceptions preference and I can go back and subclass if needed.

starboarder2001:
* Added specific pkcs11 exceptions instead of using SSHException everywhere
* Added options to use a different keyid or slotid in pkcs11_open_session.
* Simplified API by removing pkcs11pin and pkcs11provider from connect arguments, now its just pkcs11session.

I tested in my fork of Ansible and with a few test scripts, so far so good.  Feel free to just direct issues back to me if they come up.  I would appreciate if anyone wants to do testing on ubikey, all my testing has been with a smartcard.

paramiko/pkcs11.py
```diff
@@ -0,0 +1,160 @@
+from ctypes import (c_void_p, c_ulong, c_int, c_char_p, cast, addressof,
+                    sizeof, byref, cdll, Structure)
+import subprocess
+import os
+import errno
+from paramiko.ssh_exception import AuthenticationException
+
+
+class PKCS11Exception (Exception):
```
ploxiln: I think we probably still want this to sub-class SSHException

ploxiln:
The "just pkcs11session" refactor looks good :)

paramiko/pkcs11.py
starboarder2001: I was debating if I should.  Its technically separate from the SSH2 protocol, so I decided to keep the pkcs11 specific exceptions to not sub-class the SSHException.  For the PKCS11 authentication related exceptions (wrong pin, etc), I kept PKCS11AuthenticationException a subclass of AuthenticationException.

ploxiln: That's a fair argument, which may prevail, but the counter-argument is that paramiko.SSHException is used for anything paramiko specific, not just SSH protocol specific. Various failures to read or parse rsa keys etc also raise "SSHException".

starboarder2001: Good point about other key type exceptions raising "SSHException".  I'll change it to sub-class SSHException.

starboarder2001:
* I added a few unit tests, not much, but enough to determine it was broken with python3
* Fixed pkcs11 to work with python3.  Tested using python3 in my test environment to confirm.

bitprophet:
Cool, thanks! I spent my last OSS day brushing up on generic Yubikey use so hopefully I'll be able to test this out with my YubiKey locally today.

I've also got a branch based on the previous version of this PR (and it predates my work in master re: certificates, which clashed) so hopefully reconciling all of these won't take up too much time.

bitprophet:
Cool, confirmed I can use eg [this walkthrough](https://developers.yubico.com/PIV/Guides/SSH_with_PIV_and_PKCS11.html) to generate an RSA key on my Yubikey, plop its corresponding PKCS11-compatible cert in there too, and use it with `ssh -I`.

Then confirmed I can use the basic nonthreading example in the OP, in the same way, and it works fine 👍

What's left for me:

- [ ] Changelog checkup
- [ ] Doc checkup, e.g. putting in some of the reference links provided by @starboarder2001 
- [ ] Probably a tiny bit more refactoring/cleanup

Also thought about checking other key types, but even if this "ECCP256/384" format Yubi supports is covered under our ECDSA key class, it's kinda moot because OpenSSH itself doesn't support it via PKCS11. Arguably it'd be neat if we had a feature OpenSSH's client _lacked_ but I think it really just means almost nobody's gonna be looking to do that...

bitprophet:
Right, my test changes just needed more love re: the cert stuff merged in the meantime. (In the process of figuring this out I broke some ground on #387 so that's fun...)

Almost done now?!

bitprophet:
Grump, looking over the implementation details more closely and they're not quite up to par for the code quality I'd like in here. @starboarder2001 could you please do the following? I started in on this myself but there's a lot of it and I'm afraid I don't have the time right now (& I'd rather not just-merge it and kick the can down the road...)

- Checkout & work off of my in-progress branch: https://github.com/paramiko/paramiko/tree/final-827-int - it's got a bunch of refactoring, cleanup & docs already
- Replace old-style `%s` string formatting with `.format()`.
- String continuations like `Exception("blah \ <newline> more blah")` should just be one single line and slap `# noqa` at the end to satisfy flake8.
- Change all `pkcs11_xxx` module member names to just be `xxx` and to be imported/used as `pkcs11.xxx`.
  - E.g. instead of `from paramiko.pkcs11 import pkcs11_open_session; pkcs11_open_session(...)` it should be `from paramiko import pkcs11; pkcs11.open_session(...)`.
  - The exception classes are exempt from this requirement, since they need to be more obviously named when encountered out of context :) they can stay as-is.
- The arguments _to_ the methods in `pkcs11.py` should also have `pkcs11_` stricken from their names & use underscores where appropriate too.
  - E.g. not `pkcs11_open_session(pkcs11provider, pkcs11pin, ...)` but `open_session(provider, pin, ...)`. Namespacing exists! 👍
- The parameter(s) in classes outside that module should be named `pkcs11_xxx` instead of `pkcs11xxx`, I think this is now just the session arg in Client.
- The methods in `pkcs11.py` should all get a 1st line of actual prose and not just the param bits.

Thanks! Once that is all squared away let me know and I can do a final pass & merge. I have already done most of the extra doc work I wanted to do (see e.g. how it is in the API docs now & has a module header w/ an overview.)

starboarder2001:
I think its all squared away this time.

bitprophet:
Hm, it doesn't seem like you actually based the new work on my branch, so it undoes a bunch of the refactoring & other modifications I made in said branch :( Really what I was asking as for you to checkout that branch of mine (`final-827-int`) & make more modifications directly on top of it.

That may involve you having to make a new PR, but if so, that's totally fine - I'd rather be able to just click merge than have to resolve all these conflicts myself =/

starboarder2001:
I can use the same PR.  I'll see if I can get to it after work sometime this week, shouldn't be much trouble.

- David

bitprophet:
Yea, again don't care too much about how it looks at the Github level, so long as when I pull some branch from you to check it over, it includes `final-827-int` such that I don't have merge conflicts or  removals of the stuff I committed. Thanks! (And no rush on my part, it'll be a bit before I get to working on Paramiko 2.4.)

starboarder2001:
I ended up just creating another pull request...#1073

Issue #837, opened 2016-10-30
Pypy support


medecau:
I have noticed that pypy is mentioned in `tox.ini` but not `.travis.yml`.

bitprophet:
Rolling this into #838

PR #838, opened 2016-10-30
Fix tests

medecau:
This pull-request:
- adds test environment for python 3.5
- removes redundant pip command from tox.ini

---

Tox automates the creation of virtualenvs for each configured interpreter. Generates a package distribution according to setup.py. And finally installs that package in each environment before running the tests.

https://tox.readthedocs.io/en/latest/#what-is-tox

It is one of tox's features to also test dependencies are in place when running tests.

### DIFF
--- a/tox-requirements.txt
+++ b/tox-requirements.txt
@@ -1,3 +1,0 @@
-# Not sure why tox can't just read setup.py?
-cryptography >= 1.1
-pyasn1 >= 0.1.7
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,5 @@
 [tox]
-envlist = py26,py27,py33,py34,pypy
+envlist = py{26,27,33,34,35,py}
 
 [testenv]
-commands = pip install -q -r tox-requirements.txt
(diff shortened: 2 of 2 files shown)


bitprophet:
Thanks for this & the other ticket!

We don't fully support the tox.ini right now; I just haven't really wanted to remove it after somebody added it 🙁  Unsure if I want to keep fielding PRs to it or remove it (or spend the time to figure out if I want to instead embrace tox after all...) - will have to think about it.

----

Re: PyPy (#837) - I've never explicitly tested Paramiko on PyPy, but now that we are using cryptography.io on the backend, I wonder how well it would work. (One of the crypto principals is also a PyPy developer.)

If you want to update travis.yml so it adds PyPy cells, then at least we can see how Travis+PyPy runs the PR's branch and get a sense :) if it's happy I may then just merge the whole PR.

bitprophet:
Cherry-picked the superceding #866 back to 1.17. Still includes Python 3.5 testing on Travis; changelog credit included. thanks!

PR #863, opened 2016-12-14
Avoid PacketizerTest.test_closed_3 to fail on platforms where errno.ETIME is not defined

### LINKED ISSUE #862: Fix portability of test execution PacketizerTest.test_closed_3
There are some platforms where Python doesn't defined the errno.ETIME, it results the execution of test PacketizerTest.test_closed_3 to raise an AttributeError because of the call

`os.sterror(errno.ETIME)` in file [tests/test_packetizer.py#L119](https://github.com/paramiko/paramiko/blob/master/tests/test_packetizer.py#L119)

As FreeBSD maintainer, I included a local fix into our [ports](https://github.com/freebsd/freebsd-ports/blob/master/security/py-paramiko/files/patch-tests_test__packetizer.py) and I would like to upstream this fix because it could also impact others platforms. I will submit a PR for that.

Ansible developers did similar fix [here](https://github.com/ansible/ansible/pull/7081/files)

I think it's better to not enforce the value of the message because Linux, OSX and al differs on ETIME definition.

After fixing all tests are completed successfully using make test :
```
===>  Testing for py35-paramiko-2.0.3
===>   py35-paramiko-2.0.3 depends on file: /usr/local/bin/python3.5 - found
............................................................................................................................................................................. 0s 0s ......... 1s 0s ......... 0s ......... 0s ......... ........ .........  .........  ...
----------------------------------------------------------------------
Ran 174 tests in 17.235s

OK
```

sbz:
This changes define the proper Timer expired error message instead of raising
AttributeError when errno.ETIME is not available on the platform. fixes #862

### DIFF
--- a/tests/test_packetizer.py
+++ b/tests/test_packetizer.py
@@ -114,9 +114,13 @@ class PacketizerTest (unittest.TestCase):
         import signal
 
         class TimeoutError(Exception):
-            pass
+            def __init__(self, error_message):
+                if hasattr(errno, 'ETIME'):
+                    self.message = os.sterror(errno.ETIME)
+                else:
+                    self.messaage = error_message
 
-        def timeout(seconds=1, error_message=os.strerror(errno.ETIME)):
+        def timeout(seconds=1, error_message='Timer expired'):
             def decorator(func):
                 def _handle_timeout(signum, frame):
                     raise TimeoutError(error_message)


bitprophet:
Thanks, looks good, adding to bugfix milestone!

bitprophet:
Cherry-picked back to 1.17 on up.

PR #866, opened 2016-12-18
test with pypy on Travis CI

ploxiln:
and remove tox config
as suggested in https://github.com/paramiko/paramiko/pull/838#issuecomment-257656945

(we'll see how it goes ...)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ python:
   - "3.3"
   - "3.4"
   - "3.5"
+  - "pypy-5.4.1"
 install:
   # Self-install for setup.py-driven deps
   - pip install -e .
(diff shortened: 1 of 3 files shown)


ploxiln:
still trying to find a version of pypy3 that will work ;)

ploxiln:
I gave up on finding a new-enough version of pypy3 available on Travis CI. Anyway, this passes tests now with pypy2.7-5.4.1 added to the testing matrix.

bitprophet:
Yup, last time I looked at a PyPy related issue (was only a number of weeks ago, probably over on pyinvoke/invoke), it needed PyPy3+ to truly be considered fixed, and Travis didn't have that yet. So we're kinda waiting on that before we can be super happy.

Sounds like you got tests working on a good-enough PyPy version anyways tho. thanks!

bitprophet:
Cherry-picked back to 1.17+.

PR #868, opened 2016-12-22
2.1 - Add timeout to authentication step

timsavage:
Ran into a problem with an SSH server that became non-responsive during authentication. The current behaviour of paramiko is to block while authenitcation is in progress leading to the client app hanging on an auth response. 

Change is to add an `auth_timeout` option and apply the timeout in the `wait_for_response` method of  `AuthHandler`.

Includes a test case for this additional code.

### DIFF
--- a/paramiko/auth_handler.py
+++ b/paramiko/auth_handler.py
@@ -21,6 +21,7 @@
 """
 
 import weakref
+import time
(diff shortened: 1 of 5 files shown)


timsavage:
Will re-submit from a feature branch.

PR #869, opened 2016-12-23
Add timeout to authentication step

timsavage:
Ran into a problem with an SSH server that became non-responsive during authentication. The current behaviour of paramiko is to block while authentication is in progress leading to the client app hanging on an auth response.

Change is to add an auth_timeout option and apply the timeout in the wait_for_response method of AuthHandler.

Includes test cases for this additional code.

### DIFF
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -23,6 +23,7 @@ Some unit tests for authenticating over a Transport.
 import sys
 import threading
 import unittest
+from time import sleep
 
 from paramiko import Transport, ServerInterface, RSAKey, DSSKey, \
     BadAuthenticationType, InteractiveQuery, \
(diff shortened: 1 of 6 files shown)


bitprophet:
Thanks! Added to milestone as a reminder to review later.

tests/test_auth.py
```diff
@@ -232,3 +236,24 @@ def test_8_auth_gets_disconnected(self):
...
+            self.tc.connect()
+            try:
+                remain = self.tc.auth_password('unresponsive-server', 'hello')
+            except:
+                etype, evalue, etb = sys.exc_info()
+                self.assertTrue(issubclass(etype, AuthenticationException))
+                self.assertTrue('Authentication timeout' in str(evalue))
+        finally:
+            # Restore value
+            self.tc.auth_timeout = auth_timeout
```
ploxiln: `setUp()` is run for every test method, so it's a new client for every test, so this shouldn't be needed

PR #872, opened 2017-01-08
Added support for parsing 'Include' directive in ssh config file.

FDT2k:
Just added support for parsing the Include directive in ssh config file.

### DIFF
--- a/paramiko/config.py
+++ b/paramiko/config.py
@@ -85,7 +85,7 @@ class SSHConfig (object):
                 # support for Include directive in ssh_config
                 path = value
                 # according to SSH documentation if the path is relative, we look forward in ~/.ssh
-                if os.path.isabs(path) and path[0] != '~':
+                if os.path.isabs(path) == False and path[0] != '~':
                     path = '~/.ssh/'+path
                 #expand the user home path
                 path = os.path.expanduser(path)
@@ -96,7 +96,7 @@ class SSHConfig (object):
                 for filename in glob.iglob(path):
                     if os.path.isfile(filename):
                         if filename in parsed_files:
-                            raise Exception("Include loop detected in ssh config file: %s" % path)
+                            raise Exception("Include loop detected in ssh config file: %s" % filename)
                         with open(filename) as fd:
                             parsed_files.append(filename)
                             self.parse(fd, parsed_files)
--- a/paramiko/config.py
+++ b/paramiko/config.py
@@ -26,6 +26,7 @@ import os
 import re
 import shlex
 import socket
+import glob
 
 SSH_PORT = 22
 
@@ -49,7 +50,7 @@ class SSHConfig (object):
         """
         self._config = []
 
-    def parse(self, file_obj):
+    def parse(self, file_obj,parsed_files=None):
         """
         Read an OpenSSH config from the given file object.
 
@@ -68,7 +69,7 @@ class SSHConfig (object):
                 raise Exception("Unparsable line %s" % line)
             key = match.group(1).lower()
             value = match.group(2)
-            
+
             if key == 'host':
                 self._config.append(host)
                 host = {
@@ -80,6 +81,27 @@ class SSHConfig (object):
                 # at the end (for compatibility with issue #415). After 3.x, it
                 # will simply not get stripped, leaving a nice explicit marker.
                 host['config'][key] = None
+            elif key == 'include':
+                # support for Include directive in ssh_config
+
+                path = value
+                # according to SSH documentation if the path is relative, we look forward in ~/.ssh
+                if path[0] != '/' and path[0]!='~':
+                    path = '~/.ssh/'+path
+                #expand the user home path
+                path = os.path.expanduser(path)
+
+                if parsed_files == None:
+                    parsed_files = []
+
+                #parse every included file
+                for filename in glob.iglob(path):
+                    if os.path.isfile(filename):
+                        if path in parsed_files:
+                            raise Exception("Include loop detected in ssh config file: %s" %path)
+                        with open(filename) as fd:
+                            parsed_files.append(path);
+                            self.parse(fd,parsed_files)
             else:
                 if value.startswith('"') and value.endswith('"'):
                     value = value[1:-1]
--- a/paramiko/config.py
+++ b/paramiko/config.py
@@ -50,7 +50,7 @@ class SSHConfig (object):
         """
         self._config = []
 
-    def parse(self, file_obj,parsed_files=None):
+    def parse(self, file_obj, parsed_files=None):
         """
         Read an OpenSSH config from the given file object.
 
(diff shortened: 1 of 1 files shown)


paramiko/config.py
```diff
@@ -49,7 +50,7 @@ def __init__(self):
         """
         self._config = []
 
-    def parse(self, file_obj):
+    def parse(self, file_obj,parsed_files=None):
```
ploxiln: style nit: space after comma

paramiko/config.py
```diff
@@ -80,6 +81,27 @@ def parse(self, file_obj):
                 # at the end (for compatibility with issue #415). After 3.x, it
                 # will simply not get stripped, leaving a nice explicit marker.
                 host['config'][key] = None
+            elif key == 'include':
+                # support for Include directive in ssh_config
+
+                path = value
+                # according to SSH documentation if the path is relative, we look forward in ~/.ssh
+                if path[0] != '/' and path[0]!='~':
```
ploxiln: style nit: spaces around `!=`

paramiko/config.py
```diff
@@ -80,6 +81,27 @@ def parse(self, file_obj):
...
+                    parsed_files = []
+
+                #parse every included file
+                for filename in glob.iglob(path):
+                    if os.path.isfile(filename):
+                        if path in parsed_files:
+                            raise Exception("Include loop detected in ssh config file: %s" %path)
+                        with open(filename) as fd:
+                            parsed_files.append(path);
+                            self.parse(fd,parsed_files)
```
ploxiln: more minor style nits - space after comma, no semicolon at end of the previous line, spaces around `%` three lines up

paramiko/config.py
ploxiln: you may want to use `os.path.isabs()` for the first part of this condition, to perhaps enable better compatibility with windows

ploxiln:
Overall looks good to me.

ploxiln:
Worth documenting here that this is a new feature in ssh 7.3p1, and this PR does seem to reproduce it pretty closely.

```
 Include
         Include the specified configuration file(s).  Mul‐
         tiple pathnames may be specified and each pathname
         may contain glob(3) wildcards and, for user config‐
         urations, shell-like “~” references to user home
         directories.  Files without absolute paths are
         assumed to be in ~/.ssh if included in a user con‐
         figuration file or /etc/ssh if included from the
         system configuration file.  Include directive may
         appear inside a Match or Host block to perform con‐
         ditional inclusion.
```

paramiko/config.py
```diff
@@ -80,6 +81,26 @@ def parse(self, file_obj):
...
+                #expand the user home path
+                path = os.path.expanduser(path)
+                if parsed_files == None:
+                    parsed_files = []
+
+                #parse every included file
+                for filename in glob.iglob(path):
+                    if os.path.isfile(filename):
+                        if filename in parsed_files:
+                            raise Exception("Include loop detected in ssh config file: %s" % filename)
```
bitprophet: This should be some more specific exception class, using just `Exception` is almost never useful (it's impossible to catch without catching everything else too!); not sure what works best though. Maybe ValueError, or a new custom exception class (meh).

bitprophet:
This is a comparatively easy part of the code to test so I'd love it if tests could be added.

bitprophet:
Was gonna drop this as a line comment but not sure I trust github to display it even after the code is modified, and it got complex enough to be a permanent part of the discussion. It's re: the treatment of the file path stuff on [line 87](https://github.com/paramiko/paramiko/pull/872/files#diff-a58d9daf5b4cbdffb4dee739593a7813R87):

That's not how I'd interpreting the current OpenSSH docs...it's just stating that it allows normal shell-like globbing and user home directory references (e.g. `~/some-dir/some_file.conf` expands to be `/home/myusername/some-dir/some_file.conf` - _not_ `~/.ssh/some-dir/some_file.conf`).

What they mean by "assuming `~/.ssh`" is simply that the implicit relative location is `~/.ssh`, so `Include foo` means `Include ~/.ssh/foo`:

> Files without absolute paths are assumed to be in ~/.ssh if included in a user configuration file or /etc/ssh if included from the system configuration file.

Also, that brings up another wrinkle, namely the detection of whether the current file being parsed is a "user" or "system" configuration file (and they do not even mention how it is expected to behave if it is a runtime-specified file...). Need to think about that, I could see a few different ways to handle it:

- Look at the file path of the file object, if possible, and _if_ it appears to be a user or system file, interpret relative paths as documented
- Simply grab the `os.path.dirname` of that file path and use it as the relative root; this automatically works for regular user/system conf files, but is possibly more poorly defined for runtime files and especially files loaded via Include itself (!).
    - At the same time, this feels like a relatively intuitive behavior to me - that a conf file include would be relative to where the including file is located.
    - But generally, especially given this feature area's whole point, we want to ape OpenSSH's behavior as closely as possible...so maybe source diving is needed.

tescalada:
I think the relative locations handling here is correct.

My ssh config uses `Include config.d/*`
which then expands to parse out /Users/tristan/.ssh/config.d/*

I tried adding `Include ~/sshtest`
which correctly parsed the file in /Users/tristan/sshtest

tescalada:
I would be happy to help out by adding some unit testing here, but it does not seem as straight forward to add the tests as it is for the other config tests. I would either need to add mock as a requirement to the project, or drop some test config files into the repo to be able to parse finding and reading the actual files (although I would not be able to test ~ this way). 

Do you have a preference on those two options, or a better option I didn't think of?

brycefisher:
This PR makes me super happy!! I'd love to see this get merged in the nearish future and I'm happy to help contribute some tests if @FDT2k would be amenable to PRs against this PR :-)

It seems like the work that needs doing here is:

 * [ ] Add tests somehow... Would @ploxiln / @bitprophet or others object to introducing `mock` as a dependency into this project to help with testing things like tilde expansion?
 * [ ] Get consensus on how to handle relative file paths between. It appears the current implementation in this PR would expand `Include foo` -> `$HOME/.ssh/foo`. To me that seems right as is. @bitprophet should that be something else?
 * [ ] Decide whether or not check if the file is a system or user configuration. My opinion is that we ought to skip that check and revisit this design if it causes issues for users in the future. If we really wanted to do this check, then we could look at what OpenSSH does -- but again, I would just do the tilde expansion regardless to keep the implementation simple and portable.
 * [ ] Raise a more specific exception in the case of a file inclusion loop. I have no opinion on what a better one would be, but I agree more specific than `Exception` would be great. I'd be happy to provide `ValueError` instead.

haakonn:
This doesn't support globs, such as this:

`Include config.d/*.conf`

This leads to `Exception: Include loop detected in ssh config file: /home/username/.ssh/config.d/*.conf` (but there is no include loop).

PMickael:
How about a merge ? 🙏

humiaozuzu:
Any update on this PR?

tobald:
Rebased against current master, works fine.

@bitprophet 
> That's not how I'd interpreting the current OpenSSH docs...it's just stating that it allows normal
> shell-like globbing and user home directory references (e.g. ~/some-dir/some_file.conf expands
> to be /home/myusername/some-dir/some_file.conf - not ~/.ssh/some-dir/some_file.conf).

Original author @FDT2k implementation is correct : I compared file expansion with native openssh client on Debian stable.

> Also, that brings up another wrinkle, namely the detection of whether the current file being parsed
> is a "user" or "system" configuration file (and they do not even mention how it is expected to behave
> if it is a runtime-specified file...)

Good catch, please see #1194.

@haakonn 
> This doesn't support globs, such as this:
> Include config.d/*.conf

I cannot reproduce your issue : importing patterns with wildcards patterns work fine here. It seems you are trying to import a file multiple times. (see https://github.com/paramiko/paramiko/pull/872/commits/9d183fed7118ee7ef35eb7d643c6848b54a334ba#diff-a58d9daf5b4cbdffb4dee739593a7813R101)

danihodovic:
How about a merge ? :pray:

guenhter:
@FDT2k this is such a useful feature. Would you mind working in the one comment of the review.

jsiverskog:
i cleaned this up and added a test (#1892)

anarcat:
this PR should be closed in favor of #1892.

bitprophet:
Yup, rolling into #1892

PR #875, opened 2017-01-15
Added bytes as type to write

KalleDK:
Small change so that my IDE stops telling me, that this function doesn't understand bytes :)


bitprophet:
Thanks, cherry-picked to release branches on up.

PR #885, opened 2017-02-03
Added filepointer entries

KalleDK:
Allowing io.StringIO to be used.

Also cleaned up the save, so HostKey's save method is used, like the load was used.

I would also comment on the load before save... If I delete a key and try to save, wouldn't that load the key back and thereby not deleting it ?

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -102,7 +102,12 @@ class SSHClient (ClosingContextManager):
             except IOError:
                 pass
             return
-        self._system_host_keys.load(filename)
+
+        with open(filename, 'r') as fp:
+            self._system_host_keys.load_fp(fp)
+
+    def load_host_keys_fp(self, fp):
+        self._host_keys.load_fp(fp=fp)
 
     def load_host_keys(self, filename):
         """
@@ -121,7 +126,12 @@ class SSHClient (ClosingContextManager):
         :raises IOError: if the filename could not be read
         """
         self._host_keys_filename = filename
-        self._host_keys.load(filename)
+
+        with open(filename, 'r') as fp:
+            self.load_host_keys_fp(fp)
+
+    def save_host_keys_fp(self, fp):
+        self._host_keys.save_fp(fp=fp)
 
     def save_host_keys(self, filename):
         """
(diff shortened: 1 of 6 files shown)


KalleDK:
Hmmm the pypy took too long.. Can't see my changes should do that ?

paramiko/hostkeys.py
```diff
@@ -25,11 +25,11 @@
...
 
-try:
-    from collections import MutableMapping
-except ImportError:
+try:                                                    # pragma: no cover
+    from collections import MutableMapping              # pragma: no cover
+except ImportError:                                     # pragma: no cover
     # noinspection PyUnresolvedReferences
-    from UserDict import DictMixin as MutableMapping
+    from UserDict import DictMixin as MutableMapping    # pragma: no cover
```
ploxiln: I don't think this "code coverage ignore" is worth it - just take the code coverage report with a grain of salt, understand that it's just a tool, and 100% code coverage doesn't actually guarantee everything is tested anyway.

(Side note: all the coveralls PR comments are annoying. I've seen other projects that have coveralls instead put the score as an extra commit status right next to the CI result, which I prefer.)

KalleDK: It's worth it when I try to develop and uses hotkeys to jump to next covered part :) But they can be removed if you wishes.

ploxiln: To be clear, I have no authority here, just trying to offer my opinion (to perhaps help refine things or move them along), without being annoying :)

KalleDK:
@bitprophet will you take a look? I should pass for all version. I rewrote some tests, to be more consistent and remove the need to actually read or write a file.

paramiko/client.py
```diff
@@ -102,7 +102,12 @@ def load_system_host_keys(self, filename=None):
             except IOError:
                 pass
             return
-        self._system_host_keys.load(filename)
+
+        with open(filename, 'r') as fp:
+            self._system_host_keys.load_fp(fp)
```
ploxiln: Since `HostKeys.load()` does exactly this ("with open() ...") you might as well let it, rather than duplicate that (admittedly tiny amount of) code here. Same argument applies to `load_host_keys()` below.

paramiko/hostkeys.py
```diff
@@ -151,31 +175,31 @@ def __len__(self):
...
                 else:
                     raise KeyError(key)
 
             def __getitem__(self, key):
-                for e in self._entries:
-                    if e.key.get_name() == key:
-                        return e.key
+                for entry in self._entries:
+                    if entry.key.get_name() == key:
+                        return entry.key
```
ploxiln: These diff sections just renaming `e` to `entry` is not really needed in this PR IMHO.

PR #887, opened 2017-02-06
Python 2.6 was EOL 4 years ago

KalleDK:
https://www.python.org/dev/peps/pep-0361/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ cache:
   directories:
     - $HOME/.cache/pip
 python:
-  - "2.6"
   - "2.7"
   - "3.3"
   - "3.4"


KalleDK:
/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/cryptography/__init__.py:26: DeprecationWarning: Python 2.6 is no longer supported by the Python core team, please upgrade your Python. A future version of cryptography will drop support for Python 2.6

SethMichaelLarson:
If we're dropping Python 2.6 support you also have to remove all places in the code that support Python 2.6 features. Frankly just because Python 2.6 has reached EOL doesn't mean there isn't some value in supporting it.

ploxiln:
IIRC RHEL 6 / CentOS 6 has python-2.6 and will be supported for another couple of years. @bitprophet intends to support python-2.6 for quite a while yet: https://github.com/fabric/fabric/issues/1514#issuecomment-252317746

I personally think support for python-2.6 could have been dropped in paramiko-2.x (since 1.x is still maintained) and fabric-1.13 or so. But I don't think maintaining compatibility with 2.6 is too onerous, and I respect the maintainer's judgement.

KalleDK:
fabric/fabric#1514 (comment)

I'll just close the issue then ;)

bitprophet:
Well aware of Python 2.6's support status - but thanks! Chances are we'll drop it in 3.0.

mimi1vx:
for SUSE Linux Enterprise ( python 2.6) support end 31.3.2022

Issue #344, opened 2014-06-07
SFTPClient does not keep a reference to the Client


FiloSottile:
SFTPClient does not keep (properly) a reference to the parent, so when it falls out of scope and gets closed, the sftp object becomes unusable.

``` python
>>> import paramiko
>>>
>>> username = "xxx"
>>> key_filename = 'xxx'
>>> hostname = 'xxx'
>>> port = 22

>>> def connect():
...     client = paramiko.SSHClient()
...     client.load_system_host_keys()
...     client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
...     client.connect(hostname, port, username, key_filename=key_filename)
...     sftp = client.open_sftp()
...     return sftp
...
>>> sftp = connect()
>>> sftp.listdir()
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/usr/local/lib/python2.7/site-packages/paramiko/sftp_client.py", line 153, in listdir
    return [f.filename for f in self.listdir_attr(path)]
  File "/usr/local/lib/python2.7/site-packages/paramiko/sftp_client.py", line 174, in listdir_attr
    t, msg = self._request(CMD_OPENDIR, path)
  File "/usr/local/lib/python2.7/site-packages/paramiko/sftp_client.py", line 646, in _request
    num = self._async_request(type(None), t, *arg)
  File "/usr/local/lib/python2.7/site-packages/paramiko/sftp_client.py", line 668, in _async_request
    self._send_packet(t, msg)
  File "/usr/local/lib/python2.7/site-packages/paramiko/sftp.py", line 170, in _send_packet
    self._write_all(out)
  File "/usr/local/lib/python2.7/site-packages/paramiko/sftp.py", line 135, in _write_all
    raise EOFError()
EOFError
>>>
>>> client = paramiko.SSHClient()
>>> client.load_system_host_keys()
>>> client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
>>> client.connect(hostname, port, username, key_filename=key_filename)
>>> sftp = client.open_sftp()
>>> sftp.listdir()
[u'xxx']
```

Viggor:
It's because you must return the client variable, if you don't return it the garbage collect it immediatly when the scope change.

FiloSottile:
Yeah, I know why that happens. My point is that that's completely unexpected in a idiomatic Python environment.

SFTPClient should keep a reference to Client so that it doesn't get garbage collected while it's in use.

[...]

Issue #44, opened 2011-09-30
channels don't work unless a reference to SSHClient is explicitly kept


w31rd0:
Using an instance of Channel received from somewhere without a reference to SSHClient being explicitly kept (to prevent it from being garbage collected) sometimes gives unexpected results. For instance, a call to exec_command() raises an exception about the channel being closed, while recv_exit_status() just returns -1.

Here's a synthetic test case.

``` python
import paramiko

def exec_remote_command(command, host, username=None, password=None):
    channel = _exec_remote_command(command, host, username=username, password=password)
    return channel.recv_exit_status()

def _exec_remote_command(command, host, username=None, password=None):
    client = paramiko.SSHClient()
    client.set_missing_host_key_policy(paramiko.MissingHostKeyPolicy())
    client.connect(host, username=username, password=password)
    channel = client.get_transport().open_session()
    channel.exec_command(command)
    return channel

if __name__ == '__main__':
    print exec_remote_command('ls', '127.0.0.1') # prints -1
```

PR #891, opened 2017-02-09
Add back-reference from Transport to the SSHClient that created it

omeranson:
In some cases, the SSH client is created, the command is executed, the
streams are extracted, and the explicit reference to SSHClient is then
discarded (since it was e.g. created in a function that only returns the
streams). In this case, the SHSClient may be garbage collected, and the
connection's state is undefined.

This fix adds a reference from Transport to the SSHClient that created
it. The streams have a reference to the Channel, which references the
Transport. Now that the Transport references the SSHClient, it won't be
garbage collected until it is closed.

Closes-Bug: #44
Related-Bug: #344

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -336,6 +336,7 @@ class SSHClient (ClosingContextManager):
         if banner_timeout is not None:
             t.banner_timeout = banner_timeout
         t.start_client(timeout=timeout)
+        t.set_sshclient(self)
         ResourceManager.register(self, t)
 
         server_key = t.get_remote_server_key()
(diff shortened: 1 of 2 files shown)


omeranson:
Hi, @bitprophet , I'd be happy if you could take a look.

Thanks.

bitprophet:
Can't help but think this is a code smell implying that Client and Transport need some shaking up, but I'm not about to demand that for such an old codebase 😆  thanks for the patch!

bitprophet:
Cherry-picked back to 1.17.x on up. Thanks!

PR #895, opened 2017-02-15
Fix bug in handling multiple interactive queries

seveas:
If repeated interaction is needed, check_auth_interactive_response returns an InteractiveQuery object. Unfortunately a bug caused it not to be recognized, causing autheintication to fail. This fixes said bug by properly looking at the type of the returned object.


bitprophet:
Well that's certainly a 12-year-old bug. Good catch! Suspect uncaught for so long due to server mode being the uncommon use case, and then this style of auth being itself not too often used (at least, not until recent years, assuming this maps to key+TOTP and suchlike).

bitprophet:
Cherry-picked to 1.17 on up, just in case. Thanks!

Issue #794, opened 2016-08-02
Connection fails when host key is ecdsa-sha2-nistp384 or ecdsa-sha2-nistp521


ncoult:
The code seems to have just left out a few lines to support these host key types.

ncoult:
Patch here that fixes it:

https://gist.github.com/ncoult/4e33ad3b15580f791fdd75348b902134

mcclurem:
Any chance this could get some merge-love? I've had to start patching my own installs for this bug :-(

bitprophet:
I've been letting releases slip lately but almost done with the side project that's taken up my time, so I will be trying to give Paramiko a lot of love soon. Sorry for the delays!

alexreg:
Glad to see this is supported. Looks like the docs need updating though. (Grep for "keytype" in <http://docs.paramiko.org/en/2.4/api/hostkeys.html>.)

PR #899, opened 2017-02-20
client: prefer keytype from known_hosts, more ecdsa server hostkey support

### LINKED ISSUE #865: only rsa host key checked
I'm testing with Fabric==1.13.1, paramiko==2.1.1, cryptography==1.7.1

The default ssh server setup on ubuntu 16.04 (and probably most recent versions of ubuntu and debian) is to generate one of each kind of host key. The default `sshd_config` contains this:

```
Protocol 2
# HostKeys for protocol version 2
HostKey /etc/ssh/ssh_host_rsa_key
HostKey /etc/ssh/ssh_host_dsa_key
HostKey /etc/ssh/ssh_host_ecdsa_key
HostKey /etc/ssh/ssh_host_ed25519_key
```

When using the openssh client (6.9p1 on MacOS 10.11), it chooses the ecdsa key to put in `known_hosts`, and on later invocations it checks all the host keys the server presents to find the one of the appropriate type to check.

But paramiko seems to only look at the rsa one, sees that it's not the same type as what's in known_hosts, and so ignores it. Thus paramiko never notices when the host key changes.

```
$ ssh testdeploy03.ec2.st-av.net
...
@    WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!     @
...
Someone could be eavesdropping on you right now (man-in-the-middle attack)!
It is also possible that a host key has just been changed.
The fingerprint for the ECDSA key sent by the remote host is
SHA256:aB6l8o7WhWdJ02tPavyPBGTFgqe1VaqwuLVfJttiQ/c.
...
Offending ECDSA key in /Users/pierce/.ssh/known_hosts:19
ECDSA host key for testdeploy03.ec2.st-av.net has changed and you have requested strict checking.
Host key verification failed.
$ fab -H testdeploy03.ec2.st-av.net basictask
[testdeploy03.ec2.st-av.net] Executing task 'basictask'
[testdeploy03.ec2.st-av.net] run: hostname && echo hi there
[testdeploy03.ec2.st-av.net] out: testdeploy03.ec2.st-av.net
[testdeploy03.ec2.st-av.net] out: hi there
...
Disconnecting from testdeploy03.ec2.st-av.net... done.
```

But if I put an incorrect rsa key in known_hosts, paramiko correctly notices that the host key is wrong. Or, if I put an incorrect ecdsa key in known_hosts, and change `sshd_config` to only offer the ecdsa key, paramiko correctly notices that the host key is wrong.

```
$ fab -H testdeploy03.ec2.st-av.net basictask
[testdeploy03.ec2.st-av.net] Executing task 'basictask'
[testdeploy03.ec2.st-av.net] run: hostname && echo hi there

Fatal error: Host key for testdeploy03.ec2.st-av.net did not match pre-existing key! Server's key was changed recently, or possible man-in-the-middle attack.

Underlying exception:
    ('testdeploy03.ec2.st-av.net', <paramiko.ecdsakey.ECDSAKey object at 0x10c805310>, <paramiko.ecdsakey.ECDSAKey object at 0x10c7fb7d0>)

Aborting.
```

This also works to hide the problem pointed out in #794 - if the ecdsa-sha2-nistp384/521 key is not the only host key, if there is also a dsa or rsa host key in sshd_config, then paramiko will use the rsa or dsa host key and will seem to work. (But if you had the ecdsa key in known_hosts, it's not being checked.)

(As a final detail, recent openssh server and client will refuse to actually use the dsa key, I think it's obsolete.)

ploxiln:
fixes #865 

Note that `Transport.connect()` already does something similar: if a hostkey is passed, it sets `_preferred_keys` to the type of that hostkey. (But SSHClient does not use Transport.connect(), it uses the alternative `start_client()` for more control.)

See commit messages for more details. Also, I'll be doing a bit more testing with different sets of key types available on the server, just to be sure...

### DIFF
--- a/paramiko/auth_handler.py
+++ b/paramiko/auth_handler.py
@@ -416,8 +416,8 @@ class AuthHandler (object):
             except SSHException as e:
                 self.transport._log(INFO, 'Auth rejected: public key: %s' % str(e))
                 key = None
-            except:
-                self.transport._log(INFO, 'Auth rejected: unsupported or mangled public key')
+            except Exception as e:
+                self.transport._log(INFO, 'Auth rejected: unsupported or mangled public key: %s' % str(e))
                 key = None
             if key is None:
                 self._disconnect_no_more_auth()
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -335,36 +335,38 @@ class SSHClient (ClosingContextManager):
             t.set_log_channel(self._log_channel)
         if banner_timeout is not None:
             t.banner_timeout = banner_timeout
-        t.start_client(timeout=timeout)
-        t.set_sshclient(self)
-        ResourceManager.register(self, t)
-
-        server_key = t.get_remote_server_key()
-        keytype = server_key.get_name()
 
         if port == SSH_PORT:
             server_hostkey_name = hostname
         else:
             server_hostkey_name = "[%s]:%d" % (hostname, port)
+        our_server_keys = None
 
         # If GSS-API Key Exchange is performed we are not required to check the
         # host key, because the host is authenticated via GSS-API / SSPI as
         # well as our client.
         if not self._transport.use_gss_kex:
-            our_server_key = self._system_host_keys.get(server_hostkey_name,
-                                                         {}).get(keytype, None)
-            if our_server_key is None:
-                our_server_key = self._host_keys.get(server_hostkey_name,
-                                                     {}).get(keytype, None)
-            if our_server_key is None:
-                # will raise exception if the key is rejected; let that fall out
-                self._policy.missing_host_key(self, server_hostkey_name,
-                                              server_key)
-                # if the callback returns, assume the key is ok
-                our_server_key = server_key
+            our_server_keys = self._system_host_keys.get(server_hostkey_name)
+            if our_server_keys is None:
+                our_server_keys = self._host_keys.get(server_hostkey_name)
+            if our_server_keys is not None:
+                t._preferred_keys = list(our_server_keys.keys())
 
-            if server_key != our_server_key:
-                raise BadHostKeyException(hostname, server_key, our_server_key)
+        t.start_client(timeout=timeout)
+        t.set_sshclient(self)
+        ResourceManager.register(self, t)
+
+        if not self._transport.use_gss_kex:
+            server_key = t.get_remote_server_key()
+            if our_server_keys is None:
+                # will raise exception if the key is rejected; let that fall out
+                self._policy.missing_host_key(self, server_hostkey_name, server_key)
+            else:
+                our_key = our_server_keys.get(server_key.get_name())
+                if our_key != server_key:
+                    if our_key is None:
+                        our_key = list(our_server_keys.values())[0]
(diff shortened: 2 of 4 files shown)


ploxiln:
looks like I re-discovered #794

ploxiln:
Made some updates:

  * log "HostKey agreed:"
  * include exception string in "unsupported or mangled public key" log message (inspired by #902)
  * fill `_key_info` from `ECDSAKey.supported_key_format_identifiers()` (inspired by #902)

Regarding `supported_key_format_identifiers()`, I kinda prefer just writing the key type strings out literally, both in `_key_info` and `_preferred_keys`. They're short, there's just 3 of them, they won't be changing for years, and the order in `_preferred_keys` matters. The next type to be implemented will be a new class for curve25519. Currently in this PR I don't write them literally, I call `supported_key_format_identifiers()` to cater to the existing style, but I'd be glad to change to literals if you like that.

Tested with known_hosts containing just:

  * valid ecdsa-sha2-nistp256
  * wrong ecdsa-sha2-nistp256
  * valid ssh-rsa
  * wrong ssh-rsa
  * valid ecdsa-sha2-nistp521

with the server configured with:

  * just ssh-rsa
  * just ecdsa-sha2-nistp256
  * just ecdsa-sha2-nistp521
  * all three enabled

And succeeds / fails where I expect 😁 

I think this approach is more correct than #900.

In this approach, the client prefers the key type(s) that it has in known_hosts. This enhances Fabric behavior with no changes to Fabric. This is also what the openssh client seems to do. Also, by coincidence, this is also what putty now does as of the [0.68 release](http://www.chiark.greenend.org.uk/~sgtatham/putty/changes.html) yesterday (since it only now supports ecdsa keys):

> Host key preference policy change: PuTTY prefers host key formats for which it already knows the key.

In #900, the user of the SSHClient must specify the correct key type. That would require changes to Fabric to be useful with it.

Further, note that currently, and in #900, if a different key type is negotiated from the server than is present in `known_hosts`, it is considered that they host key is *missing* when really it is a host key mismatch.

With this PR, if the server is not configured with a key type that is in `known_hosts`, it will fail at host key negotiation. But also, if that did succeed for some reason, it will be a host key mismatch.

djmattyg007:
This issue is hitting me quite hard. I'd really appreciate it if this could be merged and a new release created ASAP.

ploxiln:
If any parts of this could be considered "fixes" for existing functionality, I would be willing to open a branch with the desired parts backported to the desired older version, then rebase this on top of master after that whole process goes through. Or just having all this in the 2.2 release is also fine with me :)

ploxiln:
I'm going to open two new pull requests, as an alternative to this one. You should see the references to them show up below this comment later today :)

  * one will be for master, and include the ecdsa key specific stuff (support for nistp384 and nistp521, and prefer ecdsa keys by default)
  * the other will be for an older branch, with the SSHClient host key type selection fix

ploxiln:
alternative PRs opened ^^

bitprophet:
Closing in favor of the newer PRs, thanks!

PR #900, opened 2017-02-20
Select pref key

kasdoe:
Using Paramiko I've come across an impracticality.

Resent OpenSSH clients have a preference towards ECDSA keys, and therefore, such keys tend to end up in the known_hosts files. Paramiko on the other hand, tend to end up using RSA key types after initial negotiation., and as a result Paramiko is likely to fail host key verification even though a valid host key has previously been added to the known_hosts file by OpenSSH.

I've added functionality to the SSHClient.connect method to indicate preference towards a given type of key. Loading the know host keys prior to the call to connect enables you to identify a key type which is already in the known_hosts (if any) and instruct Paramiko to use this key type:

host = 'myserver.com'
ssh = paramiko.SSHClient()
ssh.load_host_keys(os.path.expanduser('~/.ssh/known_hosts'))

keys_of_all_known_hosts = ssh.get_host_keys()
pref_hkey_type = None
if host in keys_of_all_known_hosts:
    pref_hkey_type = keys_of_all_known_hosts[host].keys()[0]

ssh.connect(host, username='eddie', password='murphy', pref_key_type=pref_hkey_type)

During the process of developing this I've come across an issue in relation to setting of key_types through the SecurityOptions object. Getting the key_types from the SecurityOptions and setting the same values again directly afterwards would raise an exception. I have proposed a fix for this issue too.

I hope you'll find this useful, and please don't hesitate to contact me for further comments.

BR / Kasper Døring

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -226,7 +226,8 @@ class SSHClient (ClosingContextManager):
         gss_kex=False,
         gss_deleg_creds=True,
         gss_host=None,
-        banner_timeout=None
+        banner_timeout=None,
+        pref_key_type=None
     ):
         """
         Connect to an SSH server and authenticate to it.  The server's host key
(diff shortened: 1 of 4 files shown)


ploxiln:
see also #899

bitprophet:
and see also #387 and the rest of the keys label. Thanks for this, will take a closer look when I have time to try smashing through all this keys related stuff!

PR #904, opened 2017-02-22
add timeout to Transport.connect(), pass through to Transport.start_client()

mdellavo:
Transport.connect() was missing the ability to pass an auth timeout through to it's call of Transport.start_client().


bitprophet:
See also #869 though...they seem orthogonal (complimentary?) at a glance.

tgomd:
For me this would be a welcomed feature since the [pysftp](https://pysftp.readthedocs.io) uses transport.connect when connecting. If the server is unresponsive it will hang.

bskinn:

Thanks!

PR #906, opened 2017-02-23
compatibility import cleanup

ploxiln:
Remove try/except blocks which were for supporting old versions of python which paramiko no longer supports.

cStringIO was actually used unconditionally before the python3 port: 0b7d0cf0a23e4f16f8552ae05a66539119e2e920

### DIFF
--- a/paramiko/hostkeys.py
+++ b/paramiko/hostkeys.py
@@ -20,16 +20,11 @@
 import binascii
 import os
 
+from collections import MutableMapping
 from hashlib import sha1
 from hmac import HMAC
 
 from paramiko.py3compat import b, u, encodebytes, decodebytes
-
-try:
-    from collections import MutableMapping
-except ImportError:
-    # noinspection PyUnresolvedReferences
-    from UserDict import DictMixin as MutableMapping
 
 from paramiko.dsskey import DSSKey
 from paramiko.rsakey import RSAKey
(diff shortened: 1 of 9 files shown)


ploxiln:
Side note: it's a bit odd that the latest version of Fabric supports python2.5, while no supported version of paramiko does. It looks like the last version of paramiko that supported python2.5 was 1.12.4 from May of 2014.

ploxiln:
added in a bit more general import cleanup

bitprophet:
Awesome! Looks good overall. Can you rebase against 1.18? For the short term I don't want to leave 1.x folks completely hanging so I'm trying to backport most changes there.

The Fabric thing is largely laziness on my part / the fact that 2.0 is looming. Paramiko got a Py3 compat overhaul in 1.13 which is when it dropped 2.5 support; Fabric 1 can still arguably only work on Python 2 so a user with Paramiko 1.12 and Fabric 1 and Python 2.5 should still be able to operate.

ploxiln:
rebased on 1.18
(since this is not really features nor bug fixes, wasn't sure it made sense to do so)

I think it's not worth the effort and confusing git history to up-merge from 1.18 to 2.0, you'll probably want to start doing 1.17/1.18 together and 2.0/2.1/+ together (cherry-picking back to the older "bundle" of branches). Or just cherry-pick back for each, that's not uncommon.

bitprophet:
Yes, that's exactly what I've been doing so far, ever since 1.18/2.1 came out:

- get the code into 1.17 via rebase or cherry-pick
- merge up into 1.18
- get the code into 2.0 via cherry-pick
- merge up into 2.1 and master

It's just that, it's kind of a pain in the rear ;) especially since I had such a strong habit of simply "merge up up up!" from the past.

Anyway, I'll do the needful based on this, thank you!

Issue #683, opened 2016-02-08
log_to_file shall use filemode "append", not "write"


vlcinsky:
See [StackOverflow question](http://stackoverflow.com/questions/35268029/appending-the-log-file-using-logging)

Paramiko log files created by means of `paramiko.util.log_to_file("filename.log')` result always in overwriting the previous logs.

The reason is clear, the [`util.py` line 24](https://github.com/paramiko/paramiko/blob/master/paramiko/util.py#L24) is opening the file with "w" mode.

Consider changing to "a".

sunjeewa:
+1

vicosoft:
+1

PurplePinkie:
+1

bitprophet:
Fixed via #907

PurplePinkie:
very nice. tnx! ;-)

PR #907, opened 2017-02-23
log_to_file() should append instead of replace

ploxiln:
easy fix for #683 

in fairness, this was written in 2004 for demo/test code ...

### DIFF
--- a/paramiko/util.py
+++ b/paramiko/util.py
@@ -241,7 +241,7 @@ def log_to_file(filename, level=DEBUG):
     if len(l.handlers) > 0:
         return
     l.setLevel(level)
-    f = open(filename, 'w')
+    f = open(filename, 'a')
     lh = logging.StreamHandler(f)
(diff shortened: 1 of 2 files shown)


bitprophet:
Why doesn't github's "change base" feature _ever_ work the way I want it to 😭  I'll just cherry-pick these. Thanks!

ploxiln:
"change base" to an "older" branch will end up including commits that are on the newer branch already but not the older branch. The pull-request branch has to be updated as well. (I'm happy to do that by the way, I'll cherry-pick to 1.18 and push up here.)

bitprophet:
Already did it, but thanks anyway. And yea, I just...you know...wish github would do the rebasing for us. Otherwise I'm not entirely clear on how useful the feature is.

bitprophet:
Cherry-picked to 1.18, 2.0, 2.1, master.

ploxiln:
sounds good

PR #910, opened 2017-02-23
SSHClient: request host key type from known_hosts

### LINKED ISSUE #865: only rsa host key checked

ploxiln:
fixes #865 

one part of an alternative set of PRs to #899 

I only backported this to 2.1 because there were some changes around `t.start_client()` in 2.0 and 2.1 which make forward-merging have possibly tricky conflicts there. I can backport further if you would like, and won't mind the conflict resolution when forward-merging :)

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -341,37 +341,43 @@ class SSHClient (ClosingContextManager):
             t.set_log_channel(self._log_channel)
         if banner_timeout is not None:
             t.banner_timeout = banner_timeout
-        t.start_client(timeout=timeout)
-        t.set_sshclient(self)
-        ResourceManager.register(self, t)
-
-        server_key = t.get_remote_server_key()
-        keytype = server_key.get_name()
 
         if port == SSH_PORT:
             server_hostkey_name = hostname
         else:
             server_hostkey_name = "[%s]:%d" % (hostname, port)
+        our_server_keys = None
 
         # If GSS-API Key Exchange is performed we are not required to check the
         # host key, because the host is authenticated via GSS-API / SSPI as
         # well as our client.
         if not self._transport.use_gss_kex:
-            our_server_key = self._system_host_keys.get(
-                server_hostkey_name, {}).get(keytype)
-            if our_server_key is None:
-                our_server_key = self._host_keys.get(server_hostkey_name,
-                                                     {}).get(keytype, None)
-            if our_server_key is None:
-                # will raise exception if the key is rejected;
-                # let that fall out
-                self._policy.missing_host_key(self, server_hostkey_name,
-                                              server_key)
-                # if the callback returns, assume the key is ok
-                our_server_key = server_key
+            our_server_keys = self._system_host_keys.get(server_hostkey_name)
+            if our_server_keys is None:
+                our_server_keys = self._host_keys.get(server_hostkey_name)
+            if our_server_keys is not None:
+                keytype = our_server_keys.keys()[0]
+                sec_opts = t.get_security_options()
+                other_types = [x for x in sec_opts.key_types if x != keytype]
+                sec_opts.key_types = [keytype] + other_types
 
-            if server_key != our_server_key:
-                raise BadHostKeyException(hostname, server_key, our_server_key)
+        t.start_client(timeout=timeout)
+        t.set_sshclient(self)
+        ResourceManager.register(self, t)
+
+        if not self._transport.use_gss_kex:
+            server_key = t.get_remote_server_key()
+            if our_server_keys is None:
+                # will raise exception if the key is rejected
+                self._policy.missing_host_key(
+                    self, server_hostkey_name, server_key
+                )
+            else:
+                our_key = our_server_keys.get(server_key.get_name())
+                if our_key != server_key:
+                    if our_key is None:
+                        our_key = list(our_server_keys.values())[0]
+                    raise BadHostKeyException(hostname, server_key, our_key)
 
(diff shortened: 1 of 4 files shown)


bitprophet:
Nah, at this point anything that makes backporting to 1.x is a good argument for "don't bother".

If someone who's stuck on 1.x (despite the security concerns, which was one reason we even did the backend switch in 2.0) is super upset they don't have such a change, they can submit their own PR :D

ploxiln:
I can rebase this and resolve conflicts if you like (... #974 causes conflicts for pretty much all pull requests against 2.0+ ...)

bitprophet:
That'd be great, thanks! (And, yes, it does, but I'd rather not let compatibility with PRs prevent necessary change. For example all of Fabric 1's PRs will get broked good once 2.0 becomes the new master... 🙃

ploxiln:
Rebased, resolved conflicts, and also removed a commit which tweaked a log message which I realized is not related to host keys (it's related to user auth keys).

Manually tested and verified, but I'm working on a proper automated test...

ploxiln:
Added proper tests for this. If I run them against the 2.1 branch (without this PR), the ones which I expect to fail do so:

```
test_host_key_negotiation_1 (test_client.SSHClientTest) ... /Users/pierce/scratch/fabwork/paramiko/paramiko/client.py:708: UserWarning: Unknown ssh-rsa host key for [127.0.0.1]:50456: 60733844cb5186657fdedaa22b5a57d5
  key.get_fingerprint())))
FAIL
test_host_key_negotiation_2 (test_client.SSHClientTest) ... ok
test_host_key_negotiation_3 (test_client.SSHClientTest) ... ERROR
test_host_key_negotiation_4 (test_client.SSHClientTest) ... ok
```

ploxiln:
Now that #981 is merged, I can revise this to use the SecurityOptions api to change the order of the preferred host key types (to put the available key first). I may do very soon that if it's not too late...

ploxiln:
changed to use SecurityOptions, will squash down a bit if that looks good

bitprophet:
Hrm, this appears to conflict with the ResourceManager removal due to the same chunk moving around, but I think it's just re-applying 2 line nukes after merging your changes, so I've done that.

ploxiln:
Your merge conflict resolution looks correct - thanks!

PR #911, opened 2017-02-23
prefer ecdsa host key types

### LINKED ISSUE #794: Connection fails when host key is ecdsa-sha2-nistp384 or ecdsa-sha2-nistp521

ploxiln:
fixes #794 and #900 

second part of an alternative set of PRs to #899

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -122,9 +122,12 @@ class Transport (threading.Thread, ClosingContextManager):
         'hmac-sha1',
     )
     _preferred_keys = (
+        'ecdsa-sha2-nistp256',
+        'ecdsa-sha2-nistp384',
+        'ecdsa-sha2-nistp521',
         'ssh-rsa',
         'ssh-dss',
-    ) + tuple(ECDSAKey.supported_key_format_identifiers())
+    )
     _preferred_kex = (
         'diffie-hellman-group1-sha1',
         'diffie-hellman-group14-sha1',
--- a/sites/www/changelog.rst
+++ b/sites/www/changelog.rst
@@ -2,6 +2,9 @@
 Changelog
 =========
 
+* :bug:`900` Prefer ecdsa-sha2-nistp(256|384|521) host key types, like
+  the openssh client does. Thanks ``@ncoult`` for the report and Pierce
+  Lopez for the patch.
 * :bug:`794` (via :issue:`981`) Prior support for ``ecdsa-sha2-nistp(384|521)``
   algorithms didn't fully extend to covering host keys, preventing connection
   to hosts which only offer these key types and no others. This is now fixed.


ploxiln:
rebased, conflicts resolved

ploxiln:
I decided to change the style of this, and use the literal key type names.

There's only the three ECDSA key types, and that set won't change any time soon, so the balance of simplicity/robustness in this case favors just writing them out, in my humble opinion. I haven't seen an opinion either way from the maintainer, and I would be happy to change the style back if that ends up being preferred.

For reference, here's the previous version of this PR, which I thought was uglier and not really justified by the theoretical robustness:

```diff
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -122,10 +122,10 @@ class Transport (threading.Thread, ClosingContextManager):
         'hmac-md5-96',
         'hmac-sha1',
     )
-    _preferred_keys = (
+    _preferred_keys = tuple(ECDSAKey.supported_key_format_identifiers()) + (
         'ssh-rsa',
         'ssh-dss',
-    ) + tuple(ECDSAKey.supported_key_format_identifiers())
+    )
     _preferred_kex = (
         'diffie-hellman-group1-sha1',
         'diffie-hellman-group14-sha1',
@@ -210,8 +210,11 @@ class Transport (threading.Thread, ClosingContextManager):
     _key_info = {
         'ssh-rsa': RSAKey,
         'ssh-dss': DSSKey,
-        'ecdsa-sha2-nistp256': ECDSAKey,
     }
+    _key_info.update(
+        (name, ECDSAKey)
+        for name in ECDSAKey.supported_key_format_identifiers()
+    )
 
     _kex_info = {
         'diffie-hellman-group1-sha1': KexGroup1,
```

You can see the new version by looking at this PR's [Files Changed](https://github.com/paramiko/paramiko/pull/911/files)

ploxiln:
... and then I split out half of this PR to #981 because it really is a bug fix that should be backported

bitprophet:
Had a second thought about "won't this bite anyone currently relying on the selection of the older key types, and thus shouldn't this be a 3.0 change", but that's more of a worry for _user_ keys and not _host_ keys, I think.

ploxiln:
resolved conflicts, put ecdsa keys even in front of ec25519, because this is the order openssh on macOS seems to use:

> debug2: kex_parse_kexinit: ecdsa-sha2-nistp256-cert-v01@openssh.com,ecdsa-sha2-nistp384-cert-v01@openssh.com,ecdsa-sha2-nistp521-cert-v01@openssh.com,ssh-ed25519-cert-v01@openssh.com,ssh-rsa-cert-v01@openssh.com,ssh-dss-cert-v01@openssh.com,ssh-rsa-cert-v00@openssh.com,ssh-dss-cert-v00@openssh.com,ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,ssh-ed25519,ssh-rsa,ssh-dss

bitprophet:
Seems reasonable. thanks!

ploxiln:
re: selection of keys: that's what #910 is about, often the wrong host key is selected ...

ploxiln:
... to clarify: and then after the wrong host key is selected, connection is silently allowed to continue under a MissingHostKeyPolicy variant which doesn't save. This happens for fabric and I think ansible as well

bitprophet:
Seems #981 also needs the ed25519 related change, I am merging that now so I will take care of it.

Sounds like #910 is additive on top of these changes so I'll get to it when I get to it (it is open...I am trying to see how much of the 2.2 milestone I can chow down on for an actual 2.2 tonight)

bitprophet:
Mildly amusing thought, shouldn't OpenSSH (and us) be preferring the larger-key-size variants of these algorithms over the smaller-key-size ones? (I.e. ` ecdsa-sha2-nistp521,ecdsa-sha2-nistp384,ecdsa-sha2-nistp256`, so that the largest one supported by the host wins out.)

I kinda trust OpenSSH knows what it's doing so as always "do what they do" seems fine, but it made me think.

ploxiln:
The reasoning might be that the larger key sizes are needlessly/wastefully large for today, but available in case of need in the future.

bitprophet:
Honestly, if this is backwards compatible & also just not very instability-causing, we might as well roll it back to 2.0 and treat it as an outright bugfix. Otherwise feels a little odd putting this in 2.2 but not 2.0.x/2.1.x. Can always revert it in those lines if somebody gets megafucked somehow.

ploxiln:
Agree it should probably be in 2.0 as well, but the conflict resolution when up-merging to 2.1 will be a bit tricky. I do have a branch with this change applied to 2.0 already, I'll push it up ...

ploxiln:
oh nvm I was thinking about #910 again

ploxiln:
rebased on 2.0 branch

bitprophet:
Ugh I thought I said I was in the middle of doing that, but, I did not say that. Sorry! Got hung up pushing it because of a stupid docs related rabbit hole.

bitprophet:
Pushed now, includes a newer changelog line. Thanks again.

ploxiln:
no problem, thank you :)

PR #912, opened 2017-02-24
PKey._write_private_key_file: pass the password along.

jhgorrell:
Pass the password along so encrypted ssh keys are written.

Also adds `AES-256-CBC` as that is what was used to write the keys on my system.

### DIFF
--- a/demos/demo_keygen.py
+++ b/demos/demo_keygen.py
@@ -96,6 +96,8 @@ if __name__ == '__main__':
 
     if options.newphrase:
         phrase = getattr(options, 'newphrase')
+        # Turn it into bytes.
+        phrase=phrase.encode("utf-8")
 
     if options.verbose:
         pfunc = progress
--- a/demos/demo_keygen.py
+++ b/demos/demo_keygen.py
@@ -96,8 +96,6 @@ if __name__ == '__main__':
 
     if options.newphrase:
         phrase = getattr(options, 'newphrase')
-        # Turn it into bytes.
-        phrase=phrase.encode("utf-8")
 
     if options.verbose:
         pfunc = progress
--- a/tests/test_pkey.py
+++ b/tests/test_pkey.py
@@ -450,9 +450,10 @@ class KeyTest (unittest.TestCase):
         # the ok case:
         self.assertTrue(BestAvailableEncryption(b"password"))
 
+        # Doesnt work on 2.6, 2.7 & pypy
         # bad: not a string
-        with self.assertRaises(ValueError):
-            BestAvailableEncryption("not bytes")
+        # with self.assertRaises(ValueError):
+        #     BestAvailableEncryption("not bytes")
         # bad: too short
         with self.assertRaises(ValueError):
             BestAvailableEncryption(b"")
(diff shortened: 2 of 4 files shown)


ploxiln:
related: #809 #772

jhgorrell:
Anything I can do to get this, or the other fixes, merged?

demos/demo_keygen.py
```diff
@@ -96,6 +96,8 @@ def progress(arg=None):
 
     if options.newphrase:
         phrase = getattr(options, 'newphrase')
+        # Turn it into bytes.
+        phrase=phrase.encode("utf-8")
```
ploxiln: can be simplified to

```python
if options.newphrase:
    phrase = options.newphrase.encode("utf-8")
```

... but instead, it may be better to change `_write_private_key()` to use the `b()` helper from `py3compat`, which encodes as utf-8 only if the argument is not already bytes.

jhgorrell: Of course! I put it on its own line, so I could comment on the need for it.

Also considered making the change to `_write_private_key()` at the same time, but figured there might be reasons I didnt know about, so didnt.

Would you like me to make the change to `_write_private_key()` as part of this PR, as a separate PR, or something else?

ploxiln: I think the change to accept unicode passwords in `_write_private_key()` should be in this PR, it's related, and it's not yet too big a PR. But, I'm not authoritative here :)

ploxiln:
Please try to be patient - I think the maintainer will have time to consider this in the next week or so. With the topic coming back up, I think there will be more movement on it this time.

jhgorrell:
Thanks for the update!  I have it in my fork, so I can keep working in the meantime.

tests/test_pkey.py
```diff
@@ -436,3 +445,33 @@ def test_stringification(self):
...
         self.assertEqual(str(key), comparable)
+
+    def test_BestAvailableEncryption(self):
+        # the ok case:
+        self.assertTrue(BestAvailableEncryption(b"password"))
+
+        # Doesnt work on 2.6, 2.7 & pypy
+        # bad: not a string
+        # with self.assertRaises(ValueError):
+        #     BestAvailableEncryption("not bytes")
```
ploxiln: This commented-out test is no longer applicable.

Also, I don't think we need to unit-test a function that isn't in paramiko.

tests/test_pkey.py
```diff
@@ -436,3 +445,33 @@ def test_stringification(self):
...
+
+        # Doesnt work on 2.6
+        # bad: too short
+        # with self.assertRaises(ValueError):
+        #     BestAvailableEncryption(b"")
+
+    def test_keyfile_is_actually_encrypted(self):
+        # Read an existing encrypted private key
+        file_ = test_path('test_rsa_password.key')
+        # @todo: The decryptor takes string or bytes; Whereas encryption only takes bytes.
```
ploxiln: no longer "todo" :)

jhgorrell:
Thanks for the review!

(Also remembered to remove the `import` as well.)

jhgorrell:
I tried adding to the changelog, but the tests failed.

bitprophet:
Dunno why I try changing bases, it never works right. Back to finishing up my hand-pull. See #741

bitprophet:
Hand-picked this back to 2.0+ and gussied it up stylewise, added changelog, etc. Thanks!!

PR #913, opened 2017-02-24
add support for servers configured with 'PermitEmptyPasswords yes'

mzakharocsc:
password-less login support

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -616,6 +616,12 @@ class SSHClient (ClosingContextManager):
                 return
             except SSHException as e:
                 saved_exception = e
+        else:
+            try:
+                self._transport.auth_none(username)
+                return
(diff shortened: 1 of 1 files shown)


ploxiln:
The documentation string at the top of the function should be updated.

I think this might "clobber" the exceptions from other key-based authentications, in many cases - saved_exception may often end up as something like "auth_none not allowed" instead of a possibly more useful exception.

bitprophet:
@ploxiln is right but that whole awful "the exceptions just keep cascading til the end" angle is a huge nasty bug anyways that needs killing with fire as per #387. I don't see us merging this until that is fixed either way.

That aside I am +1 on the general issue of supporting passwordless, keyless logins so let's leave this open and link it in the appropriate places. Thanks!

PR #914, opened 2017-02-28
Try second factor in case GSS or password auth requested it

the-dan:
Paramiko just returns in case GSS or password authentication is failed.
In case you have a second factor we need to proceed to some kind of
interactive handler.
I added optional interactive_handler parameter to the `connect` method.
In case GSS or password auth returned some set of allowed
authentication types and it matches types used for a second factor, I
try interactive auth using supplied handler

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -226,7 +226,8 @@ class SSHClient (ClosingContextManager):
         gss_kex=False,
         gss_deleg_creds=True,
         gss_host=None,
-        banner_timeout=None
+        banner_timeout=None,
+        interactive_handler=None
     ):
         """
         Connect to an SSH server and authenticate to it.  The server's host key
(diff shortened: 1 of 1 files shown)


the-dan:
Closing and reopening to kick off Travis build again. Current build error is not very informative `{"error":"Travis encountered an error, sorry :("}`

paramiko/client.py
```diff
@@ -605,17 +609,29 @@ def _auth(self, username, password, pkey, key_filenames, allow_agent,
...
+                allowed_types = set(self._transport.auth_password(username, password))
+                two_factor = (allowed_types & two_factor_types)
+                if not two_factor:
+                    return
             except SSHException as e:
                 saved_exception = e
         elif two_factor:
             try:
-                self._transport.auth_interactive_dumb(username)
+                self._transport.auth_interactive_dumb(username, handler=interactive_handler)
```
ploxiln: It looks like this `elif two_factor:` block is redundant with  the `if two_factor:` block below. I think you can remove this one.

Further, with this block as it is, if password is None and `two_factor` is triggered, it looks like `auth_interactive_dumb()` will be called twice.

ploxiln:
This could use a rebase to resolve conflicts

Issue #915, opened 2017-03-02
hostname canonicalization with GSSAPI


pseudometric:
In `Transport.set_gss_host()`, there is this:

```
# We need the FQDN to get this working with SSPI
self.gss_host = socket.getfqdn(gss_host)
```

This is a common mistake in GSSAPI programming, and a misunderstanding. I'm sure the author needed to canonicalize the hostname in order to get his test case working with SSPI, but that does not mean it's the right thing to do in general.

When using hostbased-service names with GSSAPI (type GSS_C_NT_HOSTBASED_SERVICE) — the usual case — GSSAPI clients should pass the hostname exactly as provided by the user. It should *not* be canonicalized or altered in any way, because Kerberos implementations have their own name canonicalization policies when handling this name type, and it may not match. For example, if you are using MIT Kerberos and you set libdefaults.rdns = no in krb5.conf, this turns off the reverse-address component of hostname canonicalization — but the Python socket.getfqdn() still does it.

At a well-administered Kerberos site, the DNS and Kerberos principal names are coordinated, together with the local name canonicalization policy, so that everything works as expected. Applications that take it upon themselves to alter the hostname before passing it to GSSAPI will just break that correspondence in some cases, and not work properly.

It *is* useful to have a parameter like `gss_host`, so that users can override the hostname in cases where the site setup is wrong and names don't match. (Compare GSSAPIServerIdentity in OpenSSH. Even more useful is an option to pass a literal Kerberos principal name instead, bypassing the hostname entirely and providing complete flexibility) But, that would be the opposite of the current treatment of `gss_host`: it would *not* be canonicalized if passed. Worse, the current behavior of `gss_host` is inconsistent in this regard anyway: calling `set_gss_host()` canonicalizes, while passing `gss_host` as a parameter to the Transport init() routine does not, giving confusing behavior. It means that I can specify the literal GSSAPI hostname and fix the problem with this:

```
  t = paramiko.Transport((server, 22))
  t.connect(username = os.environ['LOGNAME'],
            gss_host = server,
            gss_auth = True,
            gss_kex  = True)
```

but not with this:

```
  ssh = paramiko.SSHClient()
  ssh.connect( hostname = server,
               gss_host = server,
               gss_auth = True,
               gss_kex  = True)
```

because the latter calls `set_gss_host()` and canonicalizes, while the former does not.

[...]

PR #919, opened 2017-03-13
Fix hostname canonicalization with GSSAPI

SebastianDeiss:
I added the parameter ``gss_trust_dns`` in ``client.py`` and ``transport.py``. Set by default to ``True`` the parameter indicates whether or not the DNS is trusted to securely canonicalize the hostname of the target host. If set to ``False`` the hostname entered will be passed to GSSAPI.
This option behaves like ``GSSAPITrustDNS`` from OpenSSH.

Also, the option ``gss_host``, if set, can be used to override the hostname for GSSAPI and therefore is considered as the equivalent to ``GSSAPIServerIdentity`` from OpenSSH.

This pull should fix issue #915.

### DIFF
--- a/demos/demo_simple.py
+++ b/demos/demo_simple.py
@@ -77,8 +77,6 @@ try:
     if not UseGSSAPI or (not UseGSSAPI and not DoGSSAPIKeyExchange):
         client.connect(hostname, port, username, password)
     else:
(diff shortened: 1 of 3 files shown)


SebastianDeiss:
Superseded by #1063.

PR #920, opened 2017-03-21
Fix mpint serialization to properly handle zero.

SethMichaelLarson:
Paramiko incorrectly handles the zero case of serializing the `mpint` type. See examples from [RFC 4251 section 5](https://www.ietf.org/rfc/rfc4251.txt) regarding the `mpint` type. Paramiko fails the first example and encodes zero as `b'\x00\x00\x00\x01\x00'` which functionally isn't incorrect but is still contrary to the RFC which states "The value zero MUST be stored as a string with zero bytes of data."

### DIFF
--- a/paramiko/util.py
+++ b/paramiko/util.py
@@ -60,6 +60,8 @@ deflate_ff = max_byte if PY2 else 0xff
 def deflate_long(n, add_sign_padding=True):
     """turns a long-int into a normalized byte string (adapted from Crypto.Util.number)"""
     # after much testing, this algorithm was deemed to be the fastest
+    if n == 0:
+        return b''
     s = bytes()
     n = long(n)
     while (n != 0) and (n != -1):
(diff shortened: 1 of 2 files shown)


SethMichaelLarson:
Seems like PyPy had a random failure for timing out after 10 minutes on Travis CI? Cycling the PR to restart CI.

bitprophet:
Thanks, seems simple enough. Curious how you ran into this & whether there's any higher level behavior this triggers? Sounds like 'no' from your description, but.

SethMichaelLarson:
@bitprophet Found it from fuzzing tests I was running against the basic types.

ploxiln:
It looks like this part just below was intended to take care of zero, I guess it's wrong? Should it be cleaned up a bit?

```python
     s = bytes()
     n = long(n)
     while (n != 0) and (n != -1):
         # no iterations in this case
     for i in enumerate(s):
         # no iterations in this case
     else:
         # degenerate case, n was either 0 or -1
         i = (0,)
         if n == 0:
             s = zero_byte
```

PR #921, opened 2017-03-22
Improve __hash__ functions

franciscouzo:


bitprophet:
Can you clue me in to _how_ this is an improvement? Context-less PRs are no fun :) thanks!

ploxiln:
Less code and "interesting" logic (multiply by 37) in these implementations, while still incorporating all the same class member data into the hash. And I think tuple hash performance is about as good as you'll get by any other strategy.

(Different hash result than before, but that's fine because it just needs to be consistent within the process.)

bitprophet:
Less code is usually good, I just wish I understood why this "multiply by 37" behavior was added in the first place.

SethMichaelLarson:
Good arbitrary prime number for combining hash values?

SethMichaelLarson:
Also smallest prime greater than 32.

ploxiln:
Compare to the very simple "djb2" hash function http://www.cse.yorku.ca/~oz/hash.html

```c
    unsigned long
    hash(unsigned char *str)
    {
        unsigned long hash = 5381;
        int c;

        while (c = *str++)
            hash = ((hash << 5) + hash) + c; /* hash * 33 + c */

        return hash;
    }
```

So you see, "hash * CONST_INT + next_input" is a common simple hash strategy (hash is expected to overflow).

Issue #929, opened 2017-03-28
Timeout for SSHClient.invoke_shell() raises EOFError, and not timeout


msutton93:
Hi Guys,

I've noticed that the SSHClient.invoke_shell() method does not support a timeout parameter, even though the method seems to revolve around its call to transport.open_session(), which does. Consequently, if the connection is lost between the call to SSHClient.connect(), and SSHClient.invoke_shell(), then the call to invoke_shell seems to hang for several minutes or so before raising an EOFError. Based on my observation and brief testing, the fix seems to be quite simple, but I'll elaborate on the problem and my methods for reproducing it first.

To force this timeout, I'm running the following code:

`fh = open("timeout", "w")`
`fh.write()`
`fh.close()`
`time.sleep(10)`
`self.channel = self.ssh_connection.invoke_shell()`

Where self.connection is an existing connection established via SSHClient.connect(). Once the file has been created, I kill the connection during the timeout so that the invoke_shell() will be unable to communicate with the remote host. The results in a several minute long hang and ultimately, the following exception:

![image](https://cloud.githubusercontent.com/assets/16131646/24399828/079f43aa-13a6-11e7-96cf-0ed61c47f1c5.png)

By overriding the invoke_shell() method and adding a timeout parameter, I was able to solve this issue. After which a more specific paramiko.ssh_exception.SSHException is raised after the specified timeout, instead of a generic EOFError exception after several minutes, as shown:

![image](https://cloud.githubusercontent.com/assets/16131646/24398566/d5a122aa-13a1-11e7-9d28-7439e2f6a257.png)

The changes made to the SSHClient.invoke_shell method can be seen below (changes highlighted in red):

![image](https://cloud.githubusercontent.com/assets/16131646/24399844/1ca3d27a-13a6-11e7-8ee7-196e3b2e9b31.png)

Add my call to invoke shell now looks like this:

`fh = open("timeout", "w")`
`fh.close()`
`time.sleep(10)`
`self.channel = self.ssh_connection.invoke_shell(timeout=self.timeout)`

This is the first time I've submitted an issue, so apologies if I'm not providing the correct information, or if this is intended functionality, etc. It was just something I noticed whilst I was throwing a bunch of edge cases at my SSH library. If you need any additional details or information just let me know :-)

Thanks!
Matt

SethMichaelLarson:
I think there's value in submitting this as a Pull Request. Timeouts are usually something that is always good to expose if it's possible.

[...]

PR #930, opened 2017-03-28
Added timeout parameter to invoke_shell() method in client.py

msutton93:
Allows a timeout parameter to be used when calling SSHClient's invoke_shell() method in client.py This means that a paramiko.ssh_exception.SSHException is raised in good time, as opposed to an EOFError after several minutes. For full details, tests etc. please see issue https://github.com/paramiko/paramiko/issues/929


SethMichaelLarson:
Might want to add a test of the new feature, make sure it can actually register a timeout. Look at the tests for timeouts on `open_shell` for an example. :)

msutton93:
Sorry, are you referring to the test cases in the tests directory?

PR #935, opened 2017-04-04
travis-ci config: fix pypy version to 5.3.1

ploxiln:
version 5.4.1 seems to work on travic-ci,
but that was never a released version of pypy
see http://doc.pypy.org/en/latest/index-of-whatsnew.html


ploxiln:
well I guess pypy-5.4.1 was a release, I found a better list at https://bitbucket.org/pypy/pypy/downloads/

PR #944, opened 2017-04-23
Add HostKeys.clear_host

Gerardwx:
I needed a method to clear an existing host a known_host file, so I added one.

### DIFF
--- a/paramiko/hostkeys.py
+++ b/paramiko/hostkeys.py
@@ -189,6 +189,24 @@ class HostKeys (MutableMapping):
             return None
         return SubDict(hostname, entries, self)
 
+    def clear_host(self, hostname):
+        """
+        Remove a hostkey entry for a given hostname or IP.  Returns True if
+        entry was found or False if it wasn't.
+        :param str hostname: the hostname (or IP) to remove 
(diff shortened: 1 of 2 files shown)


bitprophet:
Thanks, looks good at a glance! Notes mostly to myself for merge time:

- Surely we have a subroutine for the `if h.startswith('|1|')` + `constant_time_bytes_eq`, "is it a hashed entry" snippet by now? If not we should make one because I'm sure this is floating around in a few places...
- Not sure if `clear_host` is the best possible name for this, wondering if `remove_host` or `delete_host` (maybe even just `remove` or `delete`, since it's a `HostKeys` object the subject being a `host` should be kind of implicit...).

Gerardwx:
Changing method name makes total sense. Should we (I) just implement __delitem__ ? It's currently not doing anything:
`    def __delitem__(self, key):`
`    ... k = self[key]`

bitprophet:
That occurred to me too but I was too busy to doublecheck just how dict-like the object currently is. If it's expected to be used as a dict-like object then that may be the best approach, yea. Feel free to take a swing at that if so, and thanks!

PR #945, opened 2017-04-24
SSHClient: request host key type from known_hosts (backport)

ploxiln:
This is a backport of #910 to the 1.18 (PyCrypto) branch. This is because I think this is particularly important.

There are a few good ways to manage a `known_hosts` file. Some of them are pretty secure. A well-managed `known_hosts` is important for preventing man-in-the-middle attacks, if that's relevant to you.

Without this fix, when a host public key is in `known_hosts`, paramiko may act like the public key is missing from `known_hosts`, and not check that it matches.

And, this is a pretty common case - allowing missing host keys is a common default, and this case where paramiko requests a different type of host key than it has from `known_hosts` is common for modern servers which offer both rsa and ecdsa host keys. The openssh client puts the ecdsa key in `known_hosts`, and then paramiko requests the rsa key, and acts like the host is missing.

I did some manual "sanity" testing, with both 1.18 and this branch:
  * matching ecdsa host key
  * mis-matched ecdsa host key (1.18 allows connection, this branch correctly blocks)
  * matching rsa host key
  * mis-matched rsa host key


ploxiln:
Updated just as #910 was - removed actually unrelated log message change, added proper tests.

ploxiln:
I rebased one more time - I still think this is important.
(this change was merged for 2.1 long ago)

bitprophet:
Sorry, yea, I haven't given the milestone this is in any love. Since I did some 1.17/1.18 work recently for the GSSAPI patchstorm, I will do this in a bit too. Thanks!!

PR #946, opened 2017-04-25
A few additions for the ssh agent handling

seveas:
I recently had to write an ssh daemon that manipulates the ssh agent of the
connecting client. Paramiko was missing a few things to make this easy.
Specifically, I wanted to:

* Talk to the agent from the server code without jumping through proxy hoops
* See the comments of keys in the agent
* Add/remove keys from the agent

Arguably, the first of the commits in this branch is even a bugfix. Caching
keys from the agent is not quite right, as keys may be added/removed at any
time.

### DIFF
--- a/paramiko/agent.py
+++ b/paramiko/agent.py
@@ -47,7 +47,6 @@ SSH2_AGENT_SIGN_RESPONSE = 14
 class AgentSSH(object):
     def __init__(self):
         self._conn = None
-        self._keys = ()
 
     def get_keys(self):
         """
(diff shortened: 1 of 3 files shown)


July-NetEase:
I used paramikio as a ssh server，but sth really made me soconfused, who can help me.

when user log onto the jumpserver and ssh to the remote server.  sth is wrong
1、ssh.connect()
I found _agent.get_keys() is empty  though allow_agent is true
how can i got the agent keys

2、whether I can turn on the SSH agent forwarding use paramiko as the OpenSSH does

p-sherratt:
+1..  need access to ssh agent key comments which is a trivial patch, but all of the patches here look good to me

PR #951, opened 2017-04-30
Support for ECDH Key Exchange

shashankv02:
Paramiko doesn't support any of the ecdh-sha2-* family of elliptic curve based DH key exchange algorithms. Some hardened servers and the few servers operating in FIPS mode only support ecdh key exchange algorithms. FIPS standard is going to remove `dh-group14-sha1` from approved algorithms soon which will break lot of existing infrastructure using paramiko. The implementation follows RFC 5656. 

I have tested this with servers only supporting ECDH kex and also client only supporting ECDH kex with paramiko as server.

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -55,6 +55,9 @@ from paramiko.dsskey import DSSKey
 from paramiko.kex_gex import KexGex, KexGexSHA256
 from paramiko.kex_group1 import KexGroup1
 from paramiko.kex_group14 import KexGroup14
+from paramiko.kex_nistp256 import KexNistp256
+from paramiko.kex_nistp384 import KexNistp384
+from paramiko.kex_nistp521 import KexNistp521
 from paramiko.kex_gss import KexGSSGex, KexGSSGroup1, KexGSSGroup14, NullHostKey
 from paramiko.message import Message
 from paramiko.packet import Packetizer, NeedRekeyException
@@ -128,6 +131,9 @@ class Transport (threading.Thread, ClosingContextManager):
         'diffie-hellman-group14-sha1',
         'diffie-hellman-group-exchange-sha1',
         'diffie-hellman-group-exchange-sha256',
+        'ecdh-sha2-nistp256',
+        'ecdh-sha2-nistp384',
+        'ecdh-sha2-nistp521',
     )
     _preferred_compression = ('none',)
 
@@ -217,7 +223,10 @@ class Transport (threading.Thread, ClosingContextManager):
         'diffie-hellman-group-exchange-sha256': KexGexSHA256,
         'gss-group1-sha1-toWM5Slw5Ew8Mqkay+al2g==': KexGSSGroup1,
         'gss-group14-sha1-toWM5Slw5Ew8Mqkay+al2g==': KexGSSGroup14,
-        'gss-gex-sha1-toWM5Slw5Ew8Mqkay+al2g==': KexGSSGex
+        'gss-gex-sha1-toWM5Slw5Ew8Mqkay+al2g==': KexGSSGex,
+        'ecdh-sha2-nistp256': KexNistp256,
+        'ecdh-sha2-nistp384': KexNistp384,
+        'ecdh-sha2-nistp521': KexNistp521
     }
 
     _compression_info = {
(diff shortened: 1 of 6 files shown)


ploxiln:
Nice!

Minor question: why not include KexNistp384 and KexNistp521 in the same source file as KexNistp256? At 4 lines for each additional variant, my humble opinion is that they could all go in a single file named something like "kex_ecdh_nistp.py"

shashankv02:
Hi @ploxiln,

Yup. That's how I have in my local copy but I refactored it into different files before pushing to follow the existing convention ~ see kex_group1.py and kex_group14.py.

ploxiln:
True, but then again, see `kex_gex.py`. Probably best to wait for the maintainer's opinion.

bitprophet:
When we're talking near single digit number of lines in each chunk, for chunks of code which are strongly thematically related, I'm definitely all for minimizing number of files. So I agree with @ploxiln's original assessment.

Sadly, this codebase is nearly 15 years old now (and my personal involvement [I'm a picky bastard] has only been the last ~5, and that at times minimal) so...following existing style doesn't count for quite as much as in some other projects 😁 

Thanks for this!

ploxiln:
Looks good. (One sub-test spuriously failed, and another hit some travis-ci scheduling bug and never started, but I'm pretty sure those events are not related to this branch.)

shashankv02:
@ploxiln @bitprophet Any way to restart the CI build without pushing a commit?

SethMichaelLarson:
Close and reopen.

ploxiln:
you can trivially amend your commit message with `git commit --amend`, then force-push it

ploxiln:
I tested this - modifying `_preferred_kex` was the most convenient way for me to cause these kex algorithms to be chosen in my setup. It seems to work :)

paramiko/transport.py
```diff
@@ -217,7 +221,10 @@ class Transport (threading.Thread, ClosingContextManager):
         'diffie-hellman-group-exchange-sha256': KexGexSHA256,
         'gss-group1-sha1-toWM5Slw5Ew8Mqkay+al2g==': KexGSSGroup1,
         'gss-group14-sha1-toWM5Slw5Ew8Mqkay+al2g==': KexGSSGroup14,
-        'gss-gex-sha1-toWM5Slw5Ew8Mqkay+al2g==': KexGSSGex
+        'gss-gex-sha1-toWM5Slw5Ew8Mqkay+al2g==': KexGSSGex,
+        'ecdh-sha2-nistp256': KexNistp256,
+        'ecdh-sha2-nistp384': KexNistp384,
+        'ecdh-sha2-nistp521': KexNistp521
```
ploxiln: minor nit: it's a bit better to include the comma after the last item, to reduce the diff when adding another item later

shashankv02: Thanks for reviewing. Good idea about the trailing comma. I'll incorporate all the review suggestions in a single go if there are more. :)

bitprophet:
Integration test:

- Debian 8 target (OS X workstation), stock opensshd, `KexAlgorithms` set to hardened default of `<ed25519>,<ecdh-sha2-family>,<dh-group-exchange-sha256>`
- OpenSSH client selects `ed25519`, is happy, by default.
- Stripping `ed25519` out of the server's `KexAlgorithms` results in agreement on `ecdh-sha2-nistp521`.
  - This is due to my local `~/.ssh/config`; the default client `KexAlgorithms` has the keys ordered smallest-to-largest, so `ecdh-sha2-nistp256` would normally have won.
- Master paramiko, as expected, agrees upon `diffie-hellman-group-exchange-sha256`
- Integration branch (this + latest master) now agrees upon `ecdh-sha2-nistp256`

bitprophet:
Merged & gussied up w/ changelog, trailing comma, etc.

@ploxiln I figure like with the changes from yesterday, we should strongly consider moving the new kex algos above the older DH ones, so that we match OpenSSH's default ordering - thoughts? As with the host key changes, since this doesn't impact user key selection it feels relatively safe, compatibility wise. (Especially as it is going into 2.2 and not 2.0/2.1.)

bitprophet:
(For reference, #983's tl;dr is "yes, we should move the new algorithms above the old ones", and we did, and then everybody rejoiced)

Issue #949, opened 2017-04-28
Paramiko has huge memory leaks


agronick:
I wrap all of my ssh connections inside the classes shown below. The classes are invoked like this:

```
        s = SshTask()
        s.timeout = s.run_timeout = 14
        s.to_host = form.get_store('vmh01')
        s.key = settings.KEY
        s.command = "find /datastore/common/repos/ubuntu-*/Packages/ -type f | egrep 'cvs.*deb'"
        packages = s.run()
```

This should leave nothing behind. I scripted a page to repeatedly make the same request to the same server over and over again. I installed Dozer on my Django app. After a few hours of running my memory went form 100mb to 272mb. Paramiko seems to be the only thing that isn't cleaning up after itself.

Now that the app is idle this is what it is showing for counts:
paramiko.agent.Agent Min: 345 Cur: 345 Max: 345  
paramiko.auth_handler.AuthHandler Min: 345 Cur: 345 Max: 345  paramiko.buffered_pipe.BufferedPipe Min: 690 Cur: 690 Max: 690  
paramiko.channel.Channel Min: 345 Cur: 345 Max: 345  
paramiko.client.AutoAddPolicy Min: 351 Cur: 351 Max: 351  
paramiko.client.SSHClient Min: 351 Cur: 351 Max: 351  
paramiko.ecdsakey._ECDSACurve Min: 3 Cur: 3 Max: 3  
paramiko.ecdsakey._ECDSACurveSet Min: 1 Cur: 1 Max: 1  
paramiko.hostkeys.HostKeyEntry Min: 345 Cur: 345 Max: 345  
paramiko.hostkeys.HostKeys Min: 702 Cur: 702 Max: 702  
paramiko.message.Message Min: 345 Cur: 345 Max: 345  
paramiko.packet.Packetizer Min: 351 Cur: 351 Max: 351  
paramiko.py3compat.long Min: 3 Cur: 3 Max: 3  
paramiko.resource.ResourceManager Min: 1 Cur: 1 Max: 1  
paramiko.rsakey.RSAKey Min: 690 Cur: 690 Max: 690  
paramiko.ssh_exception.SSHException Min: 6 Cur: 6 Max: 6  
paramiko.transport.ChannelMap Min: 351 Cur: 351 Max: 351  
paramiko.transport.Transport Min: 351 Cur: 351 Max: 351 
paramiko.util.PFilter Min: 1 Cur: 1 Max: 1 

```

[...]

Issue #520, opened 2015-04-30
Paramiko hangs with final message 'EOF in transport thread'


colinmcintosh:
When running SSH connections to multiple devices the script will sometimes hang once the sys.exit(0) is reached. Once this is reached the script will hang indefinitely. It doesn't happen everytime but the more devices the script runs against the more likely it is to happen. It seems like it's related to the amount of time the script takes to run.

Using the faulthandler module I dumped a stacktrace for when it hangs:

```
Fatal Python error: Aborted

Thread 0x00007f47faff5700 (most recent call first):
  File "build/bdist.linux-x86_64/egg/paramiko/packet.py", line 204 in read_all
  File "build/bdist.linux-x86_64/egg/paramiko/packet.py", line 341 in read_message
  File "build/bdist.linux-x86_64/egg/paramiko/transport.py", line 1590 in run
  File "/usr/local/lib/python2.7/threading.py", line 810 in __bootstrap_inner
  File "/usr/local/lib/python2.7/threading.py", line 783 in __bootstrap

Thread 0x00007f47fbff7700 (most recent call first):
  File "build/bdist.linux-x86_64/egg/paramiko/packet.py", line 204 in read_all
  File "build/bdist.linux-x86_64/egg/paramiko/packet.py", line 341 in read_message
  File "build/bdist.linux-x86_64/egg/paramiko/transport.py", line 1590 in run
  File "/usr/local/lib/python2.7/threading.py", line 810 in __bootstrap_inner
  File "/usr/local/lib/python2.7/threading.py", line 783 in __bootstrap

Thread 0x00007f47fcff9700 (most recent call first):
  File "build/bdist.linux-x86_64/egg/paramiko/packet.py", line 204 in read_all
  File "build/bdist.linux-x86_64/egg/paramiko/packet.py", line 341 in read_message
  File "build/bdist.linux-x86_64/egg/paramiko/transport.py", line 1590 in run
  File "/usr/local/lib/python2.7/threading.py", line 810 in __bootstrap_inner
  File "/usr/local/lib/python2.7/threading.py", line 783 in __bootstrap

Thread 0x00007f47fdffb700 (most recent call first):
  File "build/bdist.linux-x86_64/egg/paramiko/packet.py", line 204 in read_all
  File "build/bdist.linux-x86_64/egg/paramiko/packet.py", line 341 in read_message
  File "build/bdist.linux-x86_64/egg/paramiko/transport.py", line 1590 in run
[...]

PR #952, opened 2017-05-04
fix SSHClient/Transport leak, remove unneeded references

ploxiln:
Fix to #949 - thanks @agronick for discovering and debugging the problem.

Caused by a fix in #891. But, it's a matter of choice and policy.

This should also be back-ported to `1.17.z`

(This issue description has been updated.)

-----

The back-reference from Transport to SSHClient was added because
the SSHClient had a destructor that would close the Transport,
and some users did not want the Transport closed when the SSHClient
was garbage collected.

The SSHClient destructor was not a normal destructor, it was
implemented with the ResourceManager singleton. This sometimes
prevented the GC cycle detector from freeing the SSHClient and
Transport even after the Transport Thread stopped running.

We can simplify these problems by just getting rid of the
ResourceManager, and the back-reference. Transports cannot be
garbage-collected while their Thread is running, `.close()` must
be called (on the SSHClient or the Transport).


ploxiln:
side note: is every test variant (with different python version) resulting in a separate coverage report ...

ploxiln:
I've twiddled the commits a bit, and force-pushed over the previous state.
@aronick I made you "author" of the simple changes you proposed, I hope that's OK.

bitprophet:
Just turned off coveralls comments because yes, this is completely ridiculous. (CodeCov seems better anyhow and I mean to switch to that sometime...)

Don't have time to specifically review & confirm/test this at the moment but I like how much red is in the PR and at a high level I'm +1 on the idea that the resource manager concept is cruft that needs axing. Thanks for this!

ploxiln:
@agronick could you run your test case on this branch, to confirm the leak is fixed for you?

@omeranson could you run your use case on this branch, and confirm that the connection streams keep working for you? (since I removed the back-reference which is theoretically no longer needed)

agronick:
Sure, I will when I get to work tomorrow. Thanks for getting this merged so fast.

agronick:
I'm happy to be the author. I would have made a PR but there were a number of ways to fix it. I didn't know which one would be the most beneficial or if the ResourceManager did anything special I wasn't aware of.

ploxiln:
It's not merged yet ;)

I've tested this branch a bit, and functionality-wise stuff still works, for my use cases (mostly fabric). But I've also run some minimal-case experiments, and Transport objects do not seem to be GCed unless they are explicitly close()ed.

I've tested both python 2.7.13 and 3.6.1 on macOS. SSHClient objects are easily GCed and cause no problems once gone (now). But even if no channels are created on them, Transport objects remain, even after a sleep and a `gc.collect()`. Unless I call `.close()` on the transport (before losing my reference to it). So that's why this was probably done in the first place ...

ploxiln:
It's because of threads.

From `transport.py` top level (just after imports):

```python
# for thread cleanup
_active_threads = []

def _join_lingering_threads():
    for thr in _active_threads:
        thr.stop_thread()

import atexit
atexit.register(_join_lingering_threads)
```

from `transport.py` `Transport` class:

```python
    def run(self):
        # (use the exposed "run" method, because if we specify a thread target
        # of a private method, threading.Thread will keep a reference to it
        # indefinitely, creating a GC cycle and not letting Transport ever be
        # GC'd. it's a bug in Thread.)

        # Hold reference to 'sys' so we can test sys.modules to detect
        # interpreter shutdown.
        self.sys = sys

        # active=True occurs before the thread is launched, to avoid a race
        _active_threads.append(self)
        if self.server_mode:
            self._log(DEBUG, 'starting thread (server mode): %s' % hex(long(id(self)) & xffffffff))
        else:
            self._log(DEBUG, 'starting thread (client mode): %s' % hex(long(id(self)) & xffffffff))

        ... very long run() method ...

            except Exception as e:
                self._log(ERROR, 'Unknown exception: ' + str(e))
                self._log(ERROR, util.tb_strings())
                self.saved_exception = e
            _active_threads.remove(self)

```

So this `_active_threads` reference is to clean up remaining Transport threads when exiting the process - to prevent hangs I guess, I'm not completely sure. Anyway, if I comment out all `_active_threads` code, the Transport still leaks (obviously because of the Thread it created).

Long story short, it looks like `Transport.close()` needs to be called, one way or the other, before a Transport can be GCed (and it can't be the other way around).

So that's why this ResourceManager was created to tie the lifetime of the Transport to the SSHClient (but could still probably be replaced with a `__del__()` method on the SSHClient).

So it seems the choice is between:
  * go back to the way things were, you need to keep a reference to the SSHClient for the Transport and Channel objects to keep working
  * you need to explicitly close the Transport (or the SSHClient)

ploxiln:
`Transport.__init__()` does `self.setDaemon(True)`

Python documentation says:

> Daemon threads are abruptly stopped at shutdown. Their resources (such as open files, database transactions, etc.) may not be released properly.

So I'm not sure how much the "atexit thread cleanup" is needed. But `Transport.stop_thread()` is non-trivial ... and even if `_active_threads` was removed, Transport would still leak. So I dunno...

ploxiln:
I can imagine a possible fix, but it's a really huge refactor:

Make the Transport not a Thread itself, but instead have a reference to a TransportThread. Make Channel a wrapper for a RawChannel (or whatever). The point is that while Channel and Transport refer to each other, and to the inner TransportThread and RawChannel, the inner TransportThread and RawChannel do not have references back out. So if the Transport and all Channel objs are no longer referred to by user code, they can be detected as a cycle and garbage collected, and the Transport can stop the TransportThread on its way out (doing what the SSHClient destructor did, but not while the Transport or a Channel or a ChannelFile are still being used by the user).

I don't think any of us currently have the time for a change of that magnitude though ...

ploxiln:
New conclusion: This branch is an improvement over the current state.

#949 was really about the fact that it was possible for the SSHClient and Transport to leak, _even if `Transport.close()` was called_.

There's a bunch of factors that contributed, but the smallest and most recent one is: the `Transport._sshclient` was cleared in `Transport.close()`, but not at the end of `Transport.run()` where it repeats almost all the logic in `.close()` (and prevents `.close()` from running). So if `.run()` was cut short by an exception, instead of by the user calling `.close()`, the leak occurred.

I think the best fix is (still) to get rid of both the back-reference and the ResourceManager. I'm going to remove the addition of the destructor because I think it's useless (Transport will never be destructed until after `.close()` logic happens, whether in `.close()` or at the end of `.run()`).

I've experimented and tested enough now - I think this change is good.
(If your transports stay connected/running, they'll still leak. That's just how it'll have to be though.)

omeranson:
@ploxiln My scenario works on with this change. Thanks.

agronick:
Transport get destructed when the rekey exception is thrown. I was seeing the same thing. You need to give it a minute while it just loops in read_all on the Packetizer before it goes away.

agronick:
So I download your version. It seems like theres a difference between yours and master. When I made the changes to master branch the transports would sit there for a minute and eventually a rekey exception was thrown. This isn't happening on your branch.

ploxiln:
This is based on branch 2.0, not 2.1/master.
You could cherry-pick the non-changelog commits to master to test there.
(I don't know whether the rekey exception makes sense ... just haven't thought about it.)

agronick:
Its not going away with the pip version (which is what I meant). I swear I remember it sitting there for a minute and cleaning up but its not happening anymore. I was changing a bunch of stuff in the pip paramiko codebase so maybe I changed something else that I don't remember. I uninstalled it along the way thinking the monkey patches solved the problem.

This is particularly toublesome because in addition to memory leaks you now have zombie threads just looping in read_all. It does seem like removing the ResourceManager causes much less transports to be sitting in purgatory. So this might have been two issues.

Maybe the other day I was seeing transports that weren't being cleaned up because of only the first issue.

agronick:
I don't know how the rekey exception could even be called because the transport is stuck in self.packetizer.read_message() in the run() method.

ploxiln:
It looks like, in both 2.0 and 2.1/master, a NeedRekeyException just continues the run() loop.

The ResourceManager was a way to close the Transport when the SSHClient goes away, because Threads can't be garbage collected while they run. (This could probably have been done much more simply with a `__del__()` method on SSHClient.)

It was reported as a bug, people didn't want that to happen. Other references to the Transport and Channels could very well still exist and be in use, and having SSHClient take it down is surprising. But that was an invisible part of the API, maybe it should not have been neutered. But it was, and it might be too late to make sense changing it back.

If you want the SSHClient-takes-down-everything behavior, maybe the more appropriate and un-surprising way is to add a destructor to your own wrapper class to do it. (Or monkey-patch the trivial destructor on to the SSHClient.)

agronick:
A destructor on the client definitely seems to fix the issue. I think not having zombie threads should superseded anything else. There might be another way to fix it. But the destructor on the client works me and probably most people.

ploxiln:
Maybe. But I think this PR would fix your original test case in #949.

I think your program _did_ call `.close()` appropriately, it only suffered leaks when an exception was thrown and caught in `Transport.run()` (in your case, "SSHException: Error reading SSH protocol banner"). That would cause `run()` to try to close/cleanup the Transport itself, but it neglected to clear the back-reference. (And then `Transport.close()` would think there was nothing to do.) Now that there's no back-reference, what `run()` does is sufficient.

I think that failing to call `.close()` on the SSHClient or Transport could be considered a bug in a program, and I think that most programs don't suffer from this bug.

I don't feel super strongly, but I think we'll need some more opinions on the two reasonable choices we have here:

  * A: programs must call `.close()` on SSHClient or Transport (many already do)
  * B: programs must keep a reference to SSHClient (most already do)

agronick:
What I posted a few posts ago was that I did test out this pull request and the transports were not going away. I had to make some other changes to get it working correctly. This was even with calling close() in a finally block. Thats why I'm saying I think we might have two issues. I'm still trying to understand what is going on. I'll make another post when I have a better understanding.

With regards to calling close() can tell you it wasn't always that way in my code and it only got added when debugging issues. Given the dynamic nature of Python users expect objects to be cleaned up automatically. Even in instances where you should use a context manager like opening a file, when it falls out of scope it will go away. I think memory leaks are a greater sin than premature garbage collection that can easily be avoided by keeping the client around. I would bet that a significant amount of people will not call close() on the client and will not notice the memory leaks.

agronick:
Ok heres whats going on. When the packetizer is stuck in its read_all loop its also locking up the transport. Nothing can close itself. Putting a `__del__` method on the client allows it to reach in and shut everything off while its looping around.

Even though I call close() it wasn't getting closed. This is why you can't rely on users to call close().
I have a method that looks like this
```
try:
    try:
        self.ssh = self.connect()
    except Exception as e:
        return "There was an error connecting"
  # Do SSH stuff
except Exception as e:
      return "Error stuff"
finally:
    if self.ssh:
        self.ssh.close()
```

I would think that the finally block would clean everything up but self.ssh is never getting assigned so it can't clean it up. The exception is in self.connect() so it doesn't return anything. You can't rely on everyone to call close() in a finally block - and to do it right.


I have another issue where some of my tasks hang. We run thousands of jobs a night and less than %1 of them will get stuck. I'm wondering if this is all related.

ploxiln:
> Even though I call close() it wasn't getting closed.

What do you mean? ... if we did add a destructor to SSHClient, that's all it would do, call `close()`. If that doesn't work, a destructor won't help.

FWIW both SSHClient and Transport can be used as context managers.

ploxiln:
you may be experiencing #520

agronick:
Transport calls packetizer. Packetizer is looping in read_all and it won't stop until self.__close is set on itself.

Whats going to call it?
Not the transport. Its locked up calling the Packetizer .
Not the `__del__` method on transport. Transport is an active thread. It won't get collected.

If the user has code that does not call close() on the transport or the client, the transport's `__del__` method is still not enough - you'll have a stuck thread.

You can see in my code, even though I thought it was cleaning up it wasn't. Lots of users are not going to do that perfectly. Close either needs to be on the `__del__` of the client or called manually in a finally block with a good reference to the client. Thats what I thought I was doing above but I wasn't because self.ssh never got assigned.

ploxiln:
OK, this is just a misunderstanding: I agree the transport's `__del__()` method is useless, that's why I got rid of it.

Don't say "calling close() isn't enough" if you didn't actually call `close()` for whatever reason ;)

agronick:
We might be asking the wrong questions. I think the best solution is that the packetizer should stop looping when the exception is thrown. Not because something went out of scope somewhere either. A destructor on the tranport would close an open connection but a broken connection shouldn't get stuck in a loop to begin with. There probably shouldn't be a destructor on the client.

ploxiln:
I've pushed up a commit to add the trivial destructor to SSHClient, partly for sake of discussion. Luckily I'm not the one who makes the final determination on this, and I'm really OK either way.

I think it's normal for the Transport thread to usually be in the `Packetizer.read_all()` loop, waiting on `self.__socket.recv(n)`. That's supposed to return if the socket is closed, either from the other end, or more typically when `Transport.close()` calls `Packetizer.close()` calls `__socket.close()`. Otherwise, it's just waiting for more data which might come across the connection. `close()` is how it knows to stop.

The Transport did not previously have a `__del__()` member "destructor". I originally added it to this PR, then I removed it from this PR.

The destructor _on the Transport_ is useless because it won't run until after `close()` happens, one way or the other. Recall that `Transport.run()` has roughly the same code that `close()` does, at the end of it, and sets `self.active = False`. Either `close()` is called to stop the thread, or the thread does the `close()` logic as it stops. The Transport destructor will never be called until after the Transport thread stops, at which point the `close()` logic will already be done.

The banner exception did not cause an endless loop - it caused a leak of Transport and SSHClient objects for you because it caused `Transport.run()` to break out of its loop, do its `close()` logic, and end the thread. That copy of the `close()` logic in `run()` neglected to clear the back-reference, and by setting `self.active = False` prevented the real `close()` method from clearing the reference when you called it.

ploxiln:
It might be worth clarifying: a running Thread is a root to the garbage collector, it will never be garbage collected while running, this is not something that can be fixed. It must stop/end.

ploxiln:
... and the point of that is, if the thread has a live ssh connection, and it's waiting to recv data, it doesn't have an obvious way to know if the rest of the program no longer has a reference to it in order to get that data. It does not have an obvious way to know that nothing will care if it receives more data. It's just waiting for something to come in, and needs some explicit indication to stop.

agronick:
Ah I forgot how the transport worked. You're right, it ending it's run method would be the same as a destructor.

So we can never get the transport to GC normally if a user doesn't want to use a client and doesn't call close on the transport. This is definitely a complicated issue. Maybe some architecture changes are needed. The object that has the thread probably shouldn't have any public APIs so when the users code goes out of scope, it can close the thread. 

I definitely think what you've proposed is an improvement but it definitely breaks some things. It wouldn't be the first time a Python package made backwards incompatible changes. There is definitely a big discussion to be had.

ploxiln:
Right. I suggested the idea of wrapper classes for Transport and Channel in https://github.com/paramiko/paramiko/pull/952#issuecomment-299371634 but realized it would be a huge change.

agronick:
Oh, I haven't looked at the channel much. I guess I didn't get what you were saying.

ploxiln:
The transport thread needs references to the channels, to pass incoming data which is for them. Users can send/recv on Channels. Channels have a `get_transport()` method which the user can call. If that returns the wrapper transport, the wrapper transport will never be garbage collected, because the thread has this indirect chain of references to it. Thus, we need wrapper channels too. That way GC can collect the wrapper channels and wrapper transport, and then a destructor can close the real transport and real channels.

The goal is for the transport thread to never have references to objects which a user of paramiko can have a direct reference to. If it does, GC can't help us.

Anyway, I think that's a project for another pull request. I think that for this pull request, one of the two simple options (mentioned a few times by now) should be chosen. Either is an improvement over the combination of the ResourceManager and the Transport back-reference.

agronick:
What about something like this.
```
class TransportThread(threading.Thread):
    # Old transport class

class Transport(object):

    def __init__(self,
                 sock,
                 default_window_size=DEFAULT_WINDOW_SIZE,
                 default_max_packet_size=DEFAULT_MAX_PACKET_SIZE,
                 gss_kex=False,
                 gss_deleg_creds=True):
        self._transport_thread = TransportThread(
            sock,
            default_window_size,
            default_max_packet_size,
            gss_kex,
            gss_deleg_creds
        )

    def __setattr__(self, key, value):
        if key == '_transport_thread':
            super(Transport, self).__setattr__(key, value)
        else:
            setattr(self._transport_thread, key, value)

    def __getattr__(self, key):
        return getattr(self._transport_thread, key)

    def __del__(self):
        self._transport_thread.close()
```

This solves the problem but doesn't look great in a debugger. If you wanted all the functions and variables to show up you could do something like looping through Transport's `__dict__` and assigning a bunch of lambads that call the functions on transport and assign each variable.

ploxiln:
That's not sufficient because of the Channels.

Consider calling `SSHClient.invoke_shell()` and then just using the `Channel` it returns. (That's roughly what #891 was about.) That Channel has a reference to the real Transport, TransportThread in your case, not the wrapper Transport. So the wrapper transport can be garbage collected while the Channel is still being used, and break the connection.

So, a wrapper transport needs to override all the methods which return channels, and return wrapper channels (which have references to the wrapper transport).

ploxiln:
I have removed the last commit I added, the one adding a destructor to SSHClient.

Now this pull request contains strictly only small improvement for each preference / use-case.

This is not the GC-able Transport you wanted, but it means you need to monkey-patch only one thing (SSHClient destructor) to get that. This is a bit of progress. I encourage you to add the full wrapper Transport solution (or propose re-adding the simple SSHClient destructor) in a separate follow-up pull request.

agronick:
I'll try it. We should iron out specifics first and whats sufficient. Is it enough to implement what I posted above with changes for the channel to reference the transport?

ploxiln:
I *think* that wrapper classes for Transport, Channel, and ChannelFile, should be sufficient. Wrapper ChannelFile should have a reference to wrapper Channel should have a reference to wrapper Transport. Only when the user no longer references any of the wrappers, the wrapper transport can be garbage collected and close all the real stuff.

agronick:
I don't get why you need to wrap channel. As long as everything is accessing Transport and only Transport is accessing TransportThread it should be fine. Why would anything need direct access for TransportThread? If channel is GCed, and Transport has no references, the transport thread gets stopped. 

Now what is sufficient for making a pass-throughs. It is just a class like the one I posted above or do the methods need to be explicitly there? This would be the easiest way.

2nd easiest is to prefix every method on TransportThread with an underscore and have the public methods on transport pass though to TransportThread. That shows up in debuggers.

3rd Easiest is to start by moving the run method to TransportThread and the functions it calls. Try to split up each function between the two in a way that makes logical sense, while keeping all the public methods on Transport and adding pass-throughs where appropriate.

ploxiln:
TransportThread needs references to (real) Channels. So (real) Channels will never be GCed.

The longer this thread gets, the smaller the chance of any change being merged ;)

agronick:
Give me two weeks and I'll have something. I'd really like some clarification on how this should be done though. Is it important that the methods are explicitly declared on the helper classes?

ploxiln:
rebased to resolve conflicts. also tweaked a related test I happened to notice so it could be re-enabled for python3

PR #958, opened 2017-05-15
fixed python3 crash on windows

DavidHowlett:
sys.stdout.write takes a unicode string not bytes

### DIFF
--- a/demos/interactive.py
+++ b/demos/interactive.py
@@ -81,7 +81,7 @@ def windows_shell(chan):
                 sys.stdout.write('\r\n*** EOF ***\r\n\r\n')
                 sys.stdout.flush()
                 break
-            sys.stdout.write(data)
+            sys.stdout.write(u(data))
             sys.stdout.flush()
         
     writer = threading.Thread(target=writeall, args=(chan,))


DavidHowlett:
I only discovered the http://www.contribution-guide.org/ after making the pull request.
:-(

ploxiln:
Looks like the situation is more complicated, this might be appropriate only for python-3.6

https://github.com/Drekin/win-unicode-console
https://stackoverflow.com/a/32176732/2807852

DavidHowlett:
Ahh yes, I was using 3.6 at the time I bumped into this issue.

bitprophet:
@ploxiln please don't tell me this means that stdout only accepts bytes in 2.7-3.5 and only accepts unicode in 3.6+? ... but only on Windows? Hashtag confused.

ploxiln:
On linux / macOS / bsd, it depends on LANG / locale - e.g. if LANG is `en_US.UTF-8` stdout encode unicode as utf-8. If LANG is "C" or unset it will try to encode as "ascii" and fail if there are any interesting unicode characters. Or, you can write bytes regardless of LANG / locale. You probably knew all that.

The situation for windows stdout appears to indeed be different (until python-3.6 when it is finally "fixed"). But I'm no expert. Some people suggest just encoding as utf-8 and writing bytes, but others complain it produces garbled output for non-ascii. It looks like you can set the windows console to cp65001 and then alias/register that as utf-8 in python. Or something. It looks like the most detailed and correct answer is https://stackoverflow.com/a/3259271/2807852 and it's a lot of code, which explains why there's a non-trivial library for this purpose, it's just not simple or easy (until python-3.6).

bitprophet:
Gross. Anyway...this is all about a demo. These demos honestly generate far more bug reports/fixes than I suspect they even help people (Especially people who should just be using Fabric :D which, esp as of v2, is quite literally Using Paramiko: The Reference Implementation.) I don't know that it's worth changing from the base case if the change then causes issues with that base case :(

Going to close as "not usefully actionable" but since this is now the reference spot for "help, I am on Windows and Python <3.6 and the demo blew up on me" I am certainly not locking conversation.

ploxiln:
I guess I'm not sure that `u()` is not the best simple thing to do here. But what I think the result would be is that it would throw an exception for any special characters on python < 3.6 (while making it work right on python 3.6)

PR #962, opened 2017-05-17
Fix zombie process accumulation in ProxyCommand

nbartos:
After a proxy command exits, a defunct process will stick around
until the calling process exits.  This can accumulate to hundreds
or thousands of processes if you are polling a remote server.

This simple fix will cleanup the zombie after it's killed.


ploxiln:
You should still send SIGTERM first, and give it a couple of seconds to exit gracefully.

ploxiln:
(to clarify, the fix is the `.wait()`, thanks for identifying that problem)

nbartos:
The latest changeset should do what you want.  It's a bit more code than I wanted it to be, because it doesn't look like python has a proper function to handle a standard TERM/wait/KILL.

paramiko/proxy.py
```diff
@@ -107,7 +107,21 @@ def recv(self, size):
             raise ProxyCommandFailure(' '.join(self.cmd), e.strerror)
 
     def close(self):
-        os.kill(self.process.pid, signal.SIGTERM)
+        kill_after = datetime.datetime.utcnow() + datetime.timedelta(seconds=5)
+        try:
+            self.process.terminate()
+            while datetime.datetime.utcnow() < kill_after:
+                self.process.poll()
+                if self.process.returncode is not None:
```
ploxiln: this can be slightly consolidated into

```python
if self.process.poll() is not None:
```

(because `.poll()` returns `returncode`)

ploxiln:
That is what I had in mind, thanks!

nbartos:
Should be good now.

ploxiln:
👍 from me

(the test failure appears to be a spurious pypy lockup in an unrelated test)

nbartos:
@bitprophet Any chance I could get a merge on this?

bitprophet:
Rebase against 2.0 and I'll consider it! (same as the other PR :)) thanks!

nbartos:
I've rebased this one against master, and created a backport for the 2.0 branch here:  https://github.com/paramiko/paramiko/pull/996

bitprophet:
Shouldn't need two PRs for this unless I'm missing something obvious; typically we merge to 2.0 and then just merge up to 2.1, 2.2, ..., master.

nbartos:
Oh sorry, I'm just used to the master = mainline and everything is backported.  I'll close the other one and move this to 2.0.

pvicente:
Hi, any plan of merging it into master and backport it to version 2?

tramontano:
This is a pretty old PR, but in case anyone in the future sees this message, for my case (Win11, Python 3.12, Paramiko 3.5.0) the combination of this https://github.com/paramiko/paramiko/pull/962 and https://github.com/JulianEberius/paramiko/commit/bb1f795db46891961b6a816df3c59e132de7f211

Solved two issues:
1. Proxy on windows
2. Access Denied received upon socket closure

Hopefully this gets merged to a future release

PR #963, opened 2017-05-18
Handle case when proxy command dies during a read

nbartos:
If the proxy command dies during a read loop, we end up in a tight
loop that causes the process to max out a cpu until it's forcefully
killed.

In my experience, the easiest way to trigger this condition is to
use a proxy command and add some text to /etc/nologin on the
destination host (not the proxy host).

Another way to trigger this condition is to send a CTRL+C wile the
second command is running:

chan.exec_command('sleep 100')
stderr = chan.makefile_stderr('rb').read()

### DIFF
--- a/paramiko/proxy.py
+++ b/paramiko/proxy.py
@@ -84,6 +84,15 @@ class ProxyCommand(ClosingContextManager):
             buffer = b''
             start = time.time()
             while len(buffer) < size:
+                # The proxy command might die just before this function was
+                # called, or during this loop. If that happens, we can end up
+                # in an infinite loop with os.read() returning 0. The best way
+                # to prevent that seems to be to check to make sure the proxy
+                # command process is still running, and raise if it's not.
+                if self.process.poll() is not None:
+                    raise ProxyCommandFailure(' '.join(self.cmd),
+                                              'died while trying to read')
+
                 select_timeout = None
                 if self.timeout is not None:
                     elapsed = (time.time() - start)


paramiko/proxy.py
```diff
@@ -81,6 +81,11 @@ def recv(self, size):
 
         :return: the string of bytes read, which may be shorter than requested
         """
+        if self.process.poll() is not None:
+            # Cleanup zombie.
+            self.process.wait()
```
ploxiln: `.wait()` is not really needed here - if `.poll()` did not return None, it already did the wait on the child and got the returncode and returned it. poll() is just like using libc wait() with WNOHANG

nbartos: Fixed

ploxiln:
A separate question which I'm not sure of, is if this is `poll()` is added in the best place. Maybe it would be better *inside* the `while` loop?

(Honestly not sure ... also btw I'm not the one who merges :)

nbartos:
In my tests it didn't matter if it was inside or outside, so I put it outside for performance reasons.  What mattered was breaking the infinite loop of whatever was calling recv().

nbartos:
@bitprophet Any chance I could get a merge on this?

bitprophet:
Mildly worried there might be situations where, somehow, `poll() is None` and yet the subproc is still in the middle of tidying up (race condition or otherwise). Not worried enough to say no though. Adding to milestone. thanks!

bitprophet:
IF you could rebase against 2.0 that'd be grand, btw!

nbartos:
I've rebased this PR against master, and created a backport PR to the 2.0 branch:  https://github.com/paramiko/paramiko/pull/995

nbartos:
@bitprophet Does new code still go into the 2.0 branch first before master?  Based on more recent PRs, new changes seem to be going into master.

vincentbernat:
From my observation, when the proxy command dies, `os.read()` will return 0 bytes and we will be stuck inside the while loop. What about also raising a failure when `b` is empty? This happens to me all the time when I move my laptop to another network.

nbartos:
@vincentbernat I believe the updated patch should address your problem as well.  Would you mind testing to confirm?

vincentbernat:
I think it should work for most cases. It doesn't handle the EOF condition (subprocess still alive but read has been closed) and it looses the current buffer (a very uncommon edge case). See #1180 for a proposition handling the root cause.

nbartos:
https://github.com/paramiko/paramiko/pull/1180 is a better implementation of this, closing.

Issue #440, opened 2014-11-18
Cannot return channel or file-like object from a function, will close connection!!


lwcolton:
For some reason I cannot return a file-like object or a channel from inside of a function.  Doing so causes the channel to be immediately closed.  
## Example:

```
import paramiko

def testp():
    ssh = paramiko.SSHClient()
    ssh.set_missing_host_key_policy(
        paramiko.AutoAddPolicy())
    ssh.connect("testserver")
    stdin, stdout, stderr = ssh.exec_command("sleep 5; echo works")
    return stdout

stdout = testp()
print("From function:")
print(repr(stdout))
print(stdout.read())
print("\n")

ssh = paramiko.SSHClient()
ssh.set_missing_host_key_policy(
    paramiko.AutoAddPolicy())
ssh.connect("testserver")
stdin, stdout, stderr = ssh.exec_command("sleep 5; echo works")
print(repr(stdout))
print(stdout.read())
print("\n")
```
## Output:

```
$ python testp.py
From function:
<paramiko.ChannelFile from <paramiko.Channel 1 (closed) -> <paramiko.Transport at 0xc187e240 (unconnected)>>>
b''


<paramiko.ChannelFile from <paramiko.Channel 1 (open) window=2097152 -> <paramiko.Transport at 0xc187e208 (cipher aes128-ctr, 128 bits) (active; 1 open channel(s))>>>
b'works\n'
```
## Version info:

```
$ python
Python 3.4.0 (default, Apr 11 2014, 13:05:11)
[GCC 4.8.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import paramiko
>>> paramiko.__version__
'1.15.1'

$ cat /etc/*release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.1 LTS"
```

ghanshyamgaurav:
you can do this by this way also.
Example:
{code}
import paramiko

def testp():
    ssh = paramiko.SSHClient()
    ssh.set_missing_host_key_policy(
        paramiko.AutoAddPolicy())
    ssh.connect("tesstserver")
    return ssh

ssh = testp()
stdout=ssh.exec_command("sleep 5; echo works")[1]
print("From function:")
print(repr(stdout))
print(stdout.read())
print("\n")

ssh = paramiko.SSHClient()
[...]

PR #964, opened 2017-05-19
Memory leaks with Paramiko / Transport wrapper class  (Bug #949)

agronick:
This is the changes that were discussed in pull request #952 and bug #949.

I moved all public Transport methods and properties to a helper class that closes the transport when it falls out of scope. The class is called Transport as to not break existing code. It's purpose is to act on the TransportThread which was previously the Transport class. The TransportThread keeps a weak reference to Transport just so it can give a strong reference to the Channel.

Once I did this I found one small reference holding the Channel in memory, preventing the Transport from closing. In the Transport's run method it uses a variable called chan. The method is running for the lifetime of the Transport so that variable never goes out of scope. The Tansport is a thread which puts it at the top of the stack, keeping everything it has a reference to in scope. When this variable is no longer needed, it is now set to None, allowing the Channel to close the Transport. The SshClient or the user's code has a strong reference to Channel, so doing this will not cause it to be lost prematurely.

By setting the chan variable to None when it is no longer needed, the Channel is not kept in scope by the Transport, and the Channel can trigger the closing of the Transport.

During my testing I found that this was sufficient to eliminate all of the memory leak issues.


paramiko/transport.py
```diff
@@ -1791,6 +2760,8 @@ def run(self):
                             self._log(ERROR, 'Channel request for unknown channel %d' % chanid)
                             self.active = False
                             self.packetizer.close()
+                        # This line is key. With it, the channel can go out of scope and shut everything down
+                        chan = None
```
ploxiln: But `self._channels` also has references to the channels ... but now that I look into it, I see it uses a `WeakValueDictionary`. So that's the key here, along with your `weakref.ref` used for `transport_proxy` - "weak refs"

ploxiln: (when I was describing why various strategies would not work in the previous PR, I was not considering use of weak refs, but used as they are here they seem somewhat reasonable)

paramiko/transport.py
```diff
@@ -67,20 +67,990 @@
...
+
+class Transport(object):
+    """
+    An SSH Transport attaches to a stream (usually a socket), negotiates an
+    encrypted session, authenticates, and then creates stream tunnels, called
+    `channels <.Channel>`, across the session.  Multiple channels can be
+    multiplexed across a single session (and often are, in the case of port
+    forwardings).
+
+    Instances of this class may be used as context managers.
```
ploxiln: For this comment to be true, this `Transport` class should sub-class `ClosingContextManager` (and `TransportThread` probably no longer needs to sub-class it).

paramiko/transport.py
```diff
@@ -67,20 +67,990 @@
...
+    """
+
+    def __del__(self):
+        """
+        This method is the reason why this class exists. The TramsportTread
+        will never be garbage collected if it is not explicitly stopped.
+        When this goes out of scope it can collect the TransportThread
+        """
+        print('Killed')
```
ploxiln: looks like a debug print left in

ploxiln:
I think that most of the pydoc strings should not be duplicated between `Transport` and `TransportThread`, most should just be on the `Transport`, and the `TransportThread` should be considered internal/private somehow.

agronick:
I missed a few methods and properties from the tests. I also forgot about a print method I added. Trying to get the updates merged in.

agronick:
As far as making everything on Transport private, I did consider prefixing every method with an underscore but decided against it.

ploxiln:
I agree, renaming all the methods on TransportThread is not a good idea, I just wonder if there's a way to make the whole TransportThread class not show up in the generated docs.

ploxiln:
just leaving note of some older related attempts: #440 and #367

ploxiln:
this would probably be a good time to rebase and resolve conflicts

maludwig:
This bug is actively making me sad. Is this still going to be merged in eventually? 2017 was a while ago.

agronick:
I'm really out of the loop on this. Has it been fixed? I don't notice any more paramiko memory leaks in my code.

ploxiln:
There has been no change regarding the need to call `close()`

stuartonyx:
Is this MR in motion because our use case is hitting the memory leak issue

PR #969, opened 2017-05-25
Allow any buffer type to written to BufferedFile (1.17)

### LINKED ISSUE #967: BufferedFile TypeError on write with buffer API objects
Unlike normal files, the paramiko `BufferedFile.write()` method throws a `TypeError` if passed an object that implements the [buffer API](https://docs.python.org/2/c-api/buffer.html), such as the builtin [`memoryview`](https://docs.python.org/2/library/stdtypes.html#memoryview) type.

Note this a Python 3 flavoured version of #285 from porting that code away from the old `buffer` type.

```
  ...
  File ".../osutils.py", line 753, in pump_string_file
    write(view[offset:offset+segment_size])
  File ".../paramiko/file.py", line 375, in write
    data = b(data)
  File ".../paramiko/py3compat.py", line 47, in b
    raise TypeError("Expected unicode or bytes, got %r" % s)
TypeError: Expected unicode or bytes, got <memory at 0x7f0ce4d08510>
```

Paramiko should not care about the types it is passed, provided they are `len`-able and can be written to the underlying (`BytesIO` here) object.

bz2:
Fixes #967

Also adds test coverage for writing various types to BufferedFile which required some small changes to the test LoopbackFile subclass, and a new skipUnlessBuiltin helper.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.sw[nop]
 *.pyc
 build/
 dist/
--- a/paramiko/file.py
+++ b/paramiko/file.py
@@ -17,7 +17,7 @@
 # 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA.
 from paramiko.common import linefeed_byte_value, crlf, cr_byte, linefeed_byte, \
     cr_byte_value
-from paramiko.py3compat import BytesIO, PY2, u, b, bytes_types
+from paramiko.py3compat import BytesIO, PY2, u, bytes_types, text_type
 
 from paramiko.util import ClosingContextManager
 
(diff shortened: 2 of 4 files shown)


bz2:
One test failed on Python 2.6, not enough of the buffer api had been backported.

```
ERROR: test_write_bytearray (tests.test_file.BufferedFileTest)
----------------------------------------------------------------------
  File ".../paramiko/tests/test_file.py", line 203, in test_write_bytearray
    f.write(bytearray(12))
  File ".../paramiko/paramiko/file.py", line 384, in write
    self._write_all(data)
  File ".../paramiko/paramiko/file.py", line 501, in _write_all
    count = self._write(data)
  File ".../paramiko/tests/test_file.py", line 49, in _write
    self.buffer.write(data)
TypeError: write() argument 1 must be string or pinned buffer, not bytearray
```

Simple enough to skip the bytearray test on 2.6.

bitprophet:
Rolling into the final PR. Thanks!

PR #970, opened 2017-05-25
Allow any buffer type to be sent to Channel (1.17)

### LINKED ISSUE #968: Channel Exception on write with buffer API objects
Similar to #967, the paramiko socket-like `Channel.send()` method throws an `Exception` if passed an object that implements the [buffer API](https://docs.python.org/2/c-api/buffer.html), such as the builtin [`memoryview`](https://docs.python.org/2/library/stdtypes.html#memoryview) type.

From the `Channel` docstring:

> A Channel is meant to behave like a socket, and has an API that should be indistinguishable from the Python socket API.

```
  ...
  File ".../osutils.py", line 2181, in send_all
    sent = sock.send(view[sent_total:sent_total+MAX_SOCKET_CHUNK])
  File ".../paramiko/channel.py", line 715, in send
    return self._send(s, m)
  File ".../paramiko/channel.py", line 1080, in _send
    m.add_string(s[:size])
  File ".../paramiko/message.py", line 285, in add_string
    s = asbytes(s)
Exception: Unknown type
```

A native socket raises `TypeError` for send with non string or buffer types.

bz2:
Fixes #968

Changes the behaviour of the underlying asbytes helper to pass along unknown types. Most callers already handle this by passing the bytes along to a file or socket-like object which will raise TypeError anyway.

Adds test coverage through the Transport implementation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.sw[nop]
 *.pyc
 build/
 dist/
--- a/paramiko/common.py
+++ b/paramiko/common.py
@@ -20,7 +20,7 @@
 Common constants and global variables.
 """
 import logging
-from paramiko.py3compat import byte_chr, PY2, bytes_types, string_types, b, long
+from paramiko.py3compat import byte_chr, PY2, bytes_types, text_type, long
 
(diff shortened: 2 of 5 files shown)


bitprophet:
Rolling into the final PR. Thanks!

PR #971, opened 2017-05-26
Allow any buffer type to be written to SFTPFile (1.17)

### LINKED ISSUE #967: BufferedFile TypeError on write with buffer API objects

bz2:
Fixes #967 #968

Rollup of earlier branches proposed as #969 and #970 with additional fix inside sftp_client.

Includes new tests for SFTPFile usage.

### DIFF
--- a/paramiko/common.py
+++ b/paramiko/common.py
@@ -160,15 +160,16 @@ else:
 
 
 def asbytes(s):
+    """Coerce to bytes if possible or return unchanged."""
     if isinstance(s, bytes_types):
         return s
     if isinstance(s, text_type):
-        # GZ 2017-05-25: Accept text and encode as utf-8 for compatibilty only.
+        # Accept text and encode as utf-8 for compatibility only.
         return s.encode("utf-8")
     asbytes = getattr(s, "asbytes", None)
     if asbytes is not None:
         return asbytes()
-    # May be an object that implements the buffer api, let callers decide
+    # May be an object that implements the buffer api, let callers handle.
     return s
 
 
--- a/paramiko/common.py
+++ b/paramiko/common.py
@@ -20,7 +20,7 @@
 Common constants and global variables.
 """
 import logging
-from paramiko.py3compat import byte_chr, PY2, bytes_types, string_types, b, long
+from paramiko.py3compat import byte_chr, PY2, bytes_types, text_type, long
 
 MSG_DISCONNECT, MSG_IGNORE, MSG_UNIMPLEMENTED, MSG_DEBUG, MSG_SERVICE_REQUEST, \
     MSG_SERVICE_ACCEPT = range(1, 7)
@@ -160,15 +160,17 @@ else:
 
 
 def asbytes(s):
-    if not isinstance(s, bytes_types):
-        if isinstance(s, string_types):
-            s = b(s)
-        else:
-            try:
-                s = s.asbytes()
-            except Exception:
-                raise Exception('Unknown type')
+    if isinstance(s, bytes_types):
+        return s
+    if isinstance(s, text_type):
+        # GZ 2017-05-25: Accept text and encode as utf-8 for compatibilty only.
+        return s.encode("utf-8")
+    asbytes = getattr(s, "asbytes", None)
+    if asbytes is not None:
+        return asbytes()
+    # May be an object that implements the buffer api, let callers decide
     return s
+
 
 xffffffff = long(0xffffffff)
 x80000000 = long(0x80000000)
--- a/paramiko/file.py
+++ b/paramiko/file.py
@@ -373,8 +373,7 @@ class BufferedFile (ClosingContextManager):
         :param str/bytes data: data to write
         """
         if isinstance(data, text_type):
-            # GZ 2017-05-25: Accepting text on a binary stream unconditionally
-            # cooercing to utf-8 seems questionable, but compatibility reasons?
+            # Accept text and encode as utf-8 for compatibility only.
             data = data.encode('utf-8')
         if self._closed:
             raise IOError('File is closed')
--- a/paramiko/file.py
+++ b/paramiko/file.py
@@ -17,7 +17,7 @@
 # 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA.
 from paramiko.common import linefeed_byte_value, crlf, cr_byte, linefeed_byte, \
     cr_byte_value
-from paramiko.py3compat import BytesIO, PY2, u, b, bytes_types
+from paramiko.py3compat import BytesIO, PY2, u, bytes_types, text_type
 
 from paramiko.util import ClosingContextManager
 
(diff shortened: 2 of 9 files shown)


paramiko/file.py
```diff
@@ -372,7 +372,10 @@ def write(self, data):
 
         :param str/bytes data: data to write
         """
-        data = b(data)
+        if isinstance(data, text_type):
+            # GZ 2017-05-25: Accepting text on a binary stream unconditionally
+            # cooercing to utf-8 seems questionable, but compatibility reasons?
```
bitprophet: Feel like I recall other tickets about this (re: preserving truly binary data instead of passing through Unicode), but unable to find them at a glance. `¯\_(ツ)_/¯`

paramiko/sftp_client.py
```diff
@@ -740,12 +740,10 @@ def _async_request(self, fileobj, t, *arg):
                     msg.add_int64(item)
                 elif isinstance(item, int):
                     msg.add_int(item)
-                elif isinstance(item, (string_types, bytes_types)):
-                    msg.add_string(item)
                 elif isinstance(item, SFTPAttributes):
                     item._pack(msg)
                 else:
-                    raise Exception('unknown type for %r type %r' % (item, type(item)))
```
bitprophet: I'd love a comment here noting explicitly how "it was not any of the other expected types, so we assume it's something that can be `asbytes()`'d into a packet-level string type, and hand it to `add_string()` which does so" (assuming I'm interpreting that intent correctly!)

paramiko/common.py
```diff
@@ -160,16 +160,18 @@
...
-            s = b(s)
-        else:
-            try:
-                s = s.asbytes()
-            except Exception:
-                raise Exception('Unknown type')
+    if isinstance(s, bytes_types):
+        return s
+    if isinstance(s, text_type):
+        # GZ 2017-05-25: Accept text and encode as utf-8 for compatibilty only.
```
ploxiln: compatibilty -> compatibility

what is "GZ"? probably don't need signature/date for this comment

bitprophet:
Appreciate the rigorous ticket + PR setup! Though in the interests of corralling conversation/changelog entries/etc, I've closed all but this one.

First, thanks for cleaning this up - sounds at a glance that most of the broken API promises re: buffer/socket type objects, happened in the big Python 3 shakeup (release 1.13). (The gift that keeps on giving!)

Second, it's probably a non-issue but I wonder how this will work out when merged up into the 2.x line - something to be aware of. Given the current state of branches, I may only release this in the 1.18+ or even 2.2+ lines, depending. Super appreciate the port to 1.17+ either way though.

Third, I'm currently going through and reviewing / leaving occasional line notes on the patchset. Will report back here with anything major, otherwise once you have a chance to respond to those I'll look at merging.

paramiko/common.py
bitprophet: Was wondering that myself. Was wondering if some sort of joke around the submitter's Github username (`bz2`) :D

paramiko/file.py
ploxiln: Truly binary data would not be `text_type` (`str` for py2, `unicode` for py2), it would be bytes or something like that, so it should be passed through fine. If someone calls `write()` with a unicode type, this kicks in, turning unicode to utf-8 bytes. This seems like the best thing to do if someone passes unicode here.

bitprophet: Well, on Python 2, it well might be `str`, thus the confusion :D but yea, not really that worried, just making a note of it.

bitprophet:
Tagging as both bug and feature since I'm torn on whether this qualifies as one of those "bugfix, but potentially instability-causing" changes that like to come out in feature releases.

paramiko/file.py
ploxiln: It's confusing - you're thinking of `string_types`. `text_type` is just `unicode` on python2

bitprophet: Yup, I sure am.

ploxiln: ... I also typoed my first comment, was "`str` for py2, `unicode` for py2", first py2 should have been py3, I just corrected it ... that couldn't have helped ...

paramiko/common.py
bz2: 'GZ' is indeed a person id for me, which I agree is confusing given the nick I use on this site.

I like id/date as markers with comments for review on tricky things that may need discussion. Am happy to move to docstring, strip or delete the comment for landing.

paramiko/sftp_client.py
bz2: That's reasonable, I'll add a comment along those lines.

paramiko/file.py
bz2: There's an important distinction from `b()` in this change, in that it allows all other types through. This *is* a tradeoff, callers get duck typing, but internal library interfaces can't rely on having a bytes type. For `BufferedFile` we get sane type handing when writing to `BytesIO` later so just passing along is best.

The reason I added this comment is that while handling here makes sense on existing branches for compatibility reasons, I'm not sure it does for master. It might be better to just raise if given unicode, and add and encode step/encoding parameter to higher level interfaces.

bz2:
Thanks for the review! I've responded to the inline comments, and will push up a change with the suggested comment tweaks shortly.

I tried to write the change to minimise conflicts, so it should apply cleanly-ish to all later branches. That said, there is some fixing up that would be nice to do around 2/3 compat and tests after this lands on master, which I'd be happy to propose as well.

paramiko/common.py
bitprophet: I'm fine with them being comments instead of docstrings, it's just that we're not used to the "initials: date" prefix formatting. Stripping them out is probably the way to go, for consistency's sake. Git blame is a thing :D

bz2:
Added a possible changelog entry in f0124d9 - feel free to use/rewrite as desired.

bitprophet:
- Directly merged into 1.17 then up to 1.18
- Couldn't find an alternate 2.0 version of the patchset so attempted to manually cherry-pick the whole range to the 2.0 branch (included some minor style/flake8 tweaks since the 2.0 lines got flake8'd recently)
- tests still passing on both interpreter families so hopefully I didn't screw it up too badly
- merged up to master

bz2:
Thanks!

The 1.17 merge to 2.0 looked pretty clean so I didn't propose a separate branch for that series, but just double checked your cherrypick and it all looks good to me.

PR #976, opened 2017-06-01
compatibility import cleanup (foreport)

ploxiln:
A forward port of #906 to 2.0 as requested in https://github.com/paramiko/paramiko/pull/906#issuecomment-282115502 a while ago.

Not including the fix for `SFTPClient.getfo()` since that was re-discovered and fixed in the recent flake8 workover.

### DIFF
--- a/paramiko/hostkeys.py
+++ b/paramiko/hostkeys.py
@@ -20,16 +20,11 @@
 import binascii
 import os
 
+from collections import MutableMapping
 from hashlib import sha1
 from hmac import HMAC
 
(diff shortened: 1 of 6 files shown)


ploxiln:
btw that flake8 warning fix is only needed for newer flake8, the warning is not produced by flake8-2.6.2 from dev-requirements.txt, I'll remove that commit if desired

bitprophet:
Yea for now we cannot use flake8 `>3` as it dropped Python 2.6 support, so you might as well rip that back out for now. Thanks!

ploxiln:
removed that commit

PR #977, opened 2017-06-02
Added missing NIST curve ECDSA key entries, should fix issue #794.

realpersonguy:


ploxiln:
this change is also included in #911 and #900 (unfortunately)

bitprophet:
Let's close in favor of #911, which is slated for the next feature release. Thanks!

PR #979, opened 2017-06-03
use cryptography's sign/verify methods instead of signer/verifier

reaperhulk:
cryptography is planning to deprecate signer/verifier. They will be on an extended deprecation cycle, but it'd be nice to switch paramiko away sooner rather than later.

Note that this will require the minimum version be bumped to 1.5.

refs https://github.com/pyca/cryptography/issues/3659


ploxiln:
pynacl and bcrypt dependencies were added in paramiko-2.2.0, so it may make sense to bump the cryptography dependency min-version in a minor release like paramiko-2.3.0

bitprophet:
I'm always on the fence and very inconsistent about my feelings re: manipulating dependencies in minor versions. Most of the time the intent is to go "if someone was happy with <version X of a dep>, we shouldn't break that".

Adding _new_ dependencies arguably doesn't break that promise (we're not forcing someone who is, say, somehow limited to that version X of the existing dep, to figure out how to upgrade) but of course, one can easily argue that it _does_ (as there's no guarantee this mythical user's environment has ready access to whatever the new deps are).

Lately I suspect that dragging my feet on this sort of thing is just too conservative and I should at _least_ wait for somebody to raise a major stink; and chances are good nobody will, because, c'mon now.

spatz:
Version 2.0 of cryptography was release on July 17th and signer and verifier are now deprecated.

PR #981, opened 2017-06-05
transport: _key_info for ecdsa-sha2-nistp384 and -nistp521

ploxiln:
To support host keys of these key types, which are already in _preferred_keys! This fixes the bug pointed out in #900 where the security options setter would fail with the list of key types the getter returned. So you couldn't get `key_types`, re-order or trim the list, and then set that list, because it would contain an un-supported type (even though you didn't add any).

I'm splitting this out from #911 because I think this fix is appropriate for release branches of paramiko. (If this is approved, I'll backport it to 1.8 as well).

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -210,6 +210,8 @@ class Transport (threading.Thread, ClosingContextManager):
         'ssh-rsa': RSAKey,
         'ssh-dss': DSSKey,
         'ecdsa-sha2-nistp256': ECDSAKey,
+        'ecdsa-sha2-nistp384': ECDSAKey,
+        'ecdsa-sha2-nistp521': ECDSAKey,
     }
 
     _kex_info = {
--- a/sites/www/changelog.rst
+++ b/sites/www/changelog.rst
@@ -2,6 +2,9 @@
 Changelog
 =========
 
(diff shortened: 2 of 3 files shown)


ploxiln:
I confirmed that there's no point in backporting this to 1.18 - it seems to have multiple other problems with ecdsa-sha2-nistp384/521 (but does work with ecdsa-sha2-nistp256 as-is) (if that's the only key type offered by the remote host)

bitprophet:
@ploxiln yea I'm largely over the idea of backporting anything that's not a) super high priority or b) super trivial, patchwise. The 1.x-2.x gap is only going to grow and make it harder over time, and given the security implications of PyCrypto, 1.x would inevitably become increasingly insecure even with frequent backports. Also, simply, prioritization is a thing. I'd like to prioritize fixing longstanding pain points over having a 100% solid backport story, when the latter is difficult.

ploxiln:
Agreed, I thought this was a trivial bugfix, but it turns out not to be for 1.18.

The real "bug" is key types in `_preferred_keys` but not in `_key_info`, and 1.18 actually only has "ecdsa-sha2-nistp256" in `_preferred_keys`, so it doesn't have the "bug" anyway. So we're all good :)

PR #986, opened 2017-06-06
Fix failure in listdir when server uses a locale (1.17)

### LINKED ISSUE #985: Server listdir fails with files saved during Feburary, August, and December
From downstream bug [lp:1522916](https://bugs.launchpad.net/brz/+bug/1522916), a paramiko server using a locale such as `fr_CH.UTF-8` will 'randomly' fail with `IOError: Failure` when `listdir()` is used.

To reproduce the server must have a suitable locale installed, by for instance running `sudo dpkg-reconfigure locales` on a Debian-based system. Then export the locale with the `LANG` envvar and call `setlocale(LC_ALL, '')` or similar in the server process. Finally, list a directory with files of suitable vintage. For French, the abbreviated month names 'févr.' 'août' and 'déc.' will trigger the problem.

On the server side, the following traceback is recorded:

```
  File ".../paramiko/sftp_server.py", line 111, in start_subsystem
    self._process(t, request_number, msg)
  File ".../paramiko/sftp_server.py", line 386, in _process
    self._read_folder(request_number, folder)
  File ".../paramiko/sftp_server.py", line 239, in _read_folder
    msg.add_string(attr)
  File ".../paramiko/message.py", line 273, in add_string
    s = asbytes(s)
Exception: Unknown type
```

Or with code is modified to allow the original exception to propagate:

```
  File ".../paramiko/sftp_server.py", line 111, in start_subsystem
    self._process(t, request_number, msg)
  File ".../paramiko/sftp_server.py", line 386, in _process
    self._read_folder(request_number, folder)
  File ".../paramiko/sftp_server.py", line 239, in _read_folder
    msg.add_string(attr)
  File ".../paramiko/message.py", line 273, in add_string
    s = asbytes(s)
  File ".../paramiko/common.py", line 168, in asbytes
    s = s.asbytes()
  File ".../paramiko/sftp_attr.py", line 223, in asbytes
    return b(str(self))
  File ".../paramiko/sftp_attr.py", line 220, in __str__
    return '%s   1 %-8d %-8d %8d %-12s %s' % (ks, uid, gid, size, datestr, filename)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xb7 in position 47: ordinal not in range(128)
```

The solution is to decode the result from [`time.strftime`](https://docs.python.org/2/library/time.html#time.strftime) as the documentation suggests. For Python 3, this has already been done by the standard library so the issue does not appear. Another way to avoid the issue is using a language neutral locale like `C.UTF-8`.

bz2:
Fixes #985

`SFTPAttributes` uses the locale-aware `%b` directive for the abbreviated month name with `time.strftime`, but was not decoding the result on Python 2.7.

Add a `strftime` alias in `py3compat` that will always return unicode, and resolve the mixing of bytes and text in `SFTPAttributes` methods.

Add a test at the `listdir` level, and a more specific test for the `SFTPAttributes` `asbytes` method.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,11 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
+addons:
+  apt:
+    packages:
+      # Include language pack for locale testing
+      - language-pack-ko
 install:
   # Self-install for setup.py-driven deps
   - pip install -e .
--- a/paramiko/py3compat.py
+++ b/paramiko/py3compat.py
@@ -1,5 +1,6 @@
-import sys
 import base64
+import time
+import sys
 
 __all__ = ['PY2', 'string_types', 'integer_types', 'text_type', 'bytes_types', 'bytes', 'long', 'input',
            'decodebytes', 'encodebytes', 'bytestring', 'byte_ord', 'byte_chr', 'byte_mask',
@@ -8,6 +9,9 @@ __all__ = ['PY2', 'string_types', 'integer_types', 'text_type', 'bytes_types', '
 PY2 = sys.version_info[0] < 3
 
 if PY2:
+    import __builtin__ as builtins
+    import locale
+
     string_types = basestring
     text_type = unicode
     bytes_types = str
@@ -17,8 +21,6 @@ if PY2:
     input = raw_input
     decodebytes = base64.decodestring
     encodebytes = base64.encodestring
-
-    import __builtin__ as builtins
 
 
     def bytestring(s):  # NOQA
@@ -101,6 +103,11 @@ if PY2:
         # 64-bit
         MAXSIZE = int((1 << 63) - 1)        # NOQA
     del X
+
+    def strftime(format, t):
+        """Same as time.strftime but returns unicode."""
+        _, encoding = locale.getlocale(locale.LC_TIME)
+        return time.strftime(format, t).decode(encoding or 'ascii')
 else:
     import collections
     import struct
@@ -167,3 +174,5 @@ else:
     next = next
 
     MAXSIZE = sys.maxsize       # NOQA
+
+    strftime = time.strftime    # NOQA
--- a/paramiko/sftp_attr.py
+++ b/paramiko/sftp_attr.py
@@ -19,7 +19,7 @@
 import stat
 import time
 from paramiko.common import x80000000, o700, o70, xffffffff
-from paramiko.py3compat import long, b
+from paramiko.py3compat import long, PY2, strftime
 
 
 class SFTPAttributes (object):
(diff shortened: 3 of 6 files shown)


bz2:
Tried enabling the new tests for ci based on [the Travis docs](https://docs.travis-ci.com/user/common-build-problems/#system-required-language-pack-isnt-installed) suggestion:

> ` - sudo apt-get update && sudo apt-get --reinstall install -qq language-pack-en language-pack-de`

But [that failed](https://travis-ci.org/paramiko/paramiko/builds/240159488) with:

> This job is running on container-based infrastructure, which does not allow use of 'sudo', setuid, and setgid executables.

Not sure if they have a way of enabling locale support in containers. It's not really essential.

bitprophet:
Thanks for this, will try to review it in more depth later.

Re: travis: we could potentially tell Travis to run us on the non-containerized, sudo-capable images; my other projects have to do this anyways because their integration suites need to test interactions with `sudo`. Means a bit of a speed hit but Paramiko's suite is one of the slower ones I manage now anyway (sob) so it's not like it really matters a lot.

bz2:
Got an alternative spelling for the langpack install from Travis support which works with containers, see latest passing run.

Will just note as well that this change shares the first to revs with #971 - if that gets merged there's a teeny bit less to review here.

bz2:
Have rebased on top of latest 1.17 changes and added a changelog entry.

bitprophet:
At this point I'm not sure I want to truck with non-critical changes on the 1.x line; would you be opposed to rebasing this onto 2.0 instead? Thanks!

bz2:
I'm happy to propose this against 2.0 as well, but is there any reason not to just land it on the 1.X series?

I can probably get a debian patch applied to fix to make our test suite work, but that's easier if the same patch is applied upstream.

ploxiln:
I'd humbly argue for this being appropriate for 1.18 - the non-test-related changes aren't too big.

bitprophet:
The main reason is just that it's a much larger pain in the ass the cross the 1.x->2.x barrier because I was an idiot and put out a 1.x feature release (1.18) after 2.0 landed. This means one cannot simply "merge up" from 1.x to master anymore, or one will pollute 2.0.x with the feature changes from 1.18.x. Instead, one has to do two separate PRs or otherwise cherry-pick; and it's easy to mess up doing this, too.

If the contributors involved have the time to cross all the T's and dot all the I's such that everything including changelog entries is set up, with 2x PRs for 1.17 and 2.0, and I can simply merge-then-merge-up...that's mostly acceptable. But eventually 1.x is going to have to die anyways (see: all the reasons 2.0 and Cryptography happened) so it's only a matter of when, in my mind.

bitprophet:
In this particular case, all the T's and I's have been crossed and dotted appropriately ( :heart: @bz2) so...sure! Why not. But that's still my rationale for why, by default, I'd really prefer to limit how much stuff "must" go back to 1.x, and I'd rather users find ways to get on 2.0 when possible.

ploxiln:
Agreed 😁

bitprophet:
Much later: I've moved on and the 1.x  series is unsupported at this point (arguably a good thing given PyCrypto remains completely dead!) so I'm going to look at merging #992 but close out this one.

PR #987, opened 2017-06-06
sshclient / transport leak partial fix (backport)

ploxiln:
backport of #952 

I'll recap the discoveries and reasoning here, since the other thread is very long:

#891 fixed "SHSClient may be garbage collected, and the connection's state is undefined", so it added an extra reference from the Transport to the SSHClient. But that wasn't the real problem or the best fix - the problem is that this ResourceManager was intentionally closing the Transport when the SSHClient was garbage collected.

So, if we don't want a garbage-collected SSHClient to close the Transport, we can just remove the code that intentionally does that.

The secondary problem with #891 is that there are two sections of code that do the internal work to close the Transport - the `close()` method, but also some code at the end of `run()` (which runs in a separate thread.) If the `run()` thread ended before the `close()` method was called, the back-reference to the SSHClient would never be cleared, and a real un-handle-able leak (rooted at the ResourceManager) would result.

This PR simplifies the code by just removing a bunch of it, satisfies #891 by not closing the Transport when the SSHClient is GCed, and fixes the "secondary problem" of the un-handle-able leak.

What this PR does not handle is the original motivation of adding the ResourceManager - the Transport, while it is running, cannot be GCed. It must be explicitly closed, or disconnect or stop for some other reason, before it can be freed. This aspect is addressed in #964 (which is big and invasive and will only be applied (eventually) to the master branch).


bitprophet:
Still scared about a change with this much instability potential going into the stable branches, but I figure if we start getting lots of complaints we can always revert, and of course ideally it'll cause more good than harm.

PR #991, opened 2017-06-11
Add Python 3.6 to classifiers

felixonmars:


bitprophet:
I gotta start doing this across all my projects on new interpreter releases...! because this must be the 3rd such PR I've gotten in the last few months. Thanks :)

PR #992, opened 2017-06-12
Fix failure in listdir when server uses a locale (2.0)

### LINKED ISSUE #985: Server listdir fails with files saved during Feburary, August, and December

bz2:
Fixes #985

SFTPAttributes uses the locale-aware %b directive for the
abbreviated month name with time.strftime, but was not
decoding the result on Python 2.7.

Add a strftime alias in py3compat that will always return
unicode, and resolve the mixing of bytes and text in
SFTPAttributes methods.

Add a test at the listdir level, and a more specific test
for the SFTPAttributes asbytes method.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,11 @@ python:
   - "3.5"
   - "3.6"
   - "pypy-5.4.1"
+addons:
+  apt:
+    packages:
+      # Include language pack for locale testing
+      - language-pack-ko
 install:
   # Self-install for setup.py-driven deps
   - pip install -e .
--- a/paramiko/py3compat.py
+++ b/paramiko/py3compat.py
@@ -1,5 +1,6 @@
-import sys
 import base64
+import sys
+import time
 
 __all__ = ['PY2', 'string_types', 'integer_types', 'text_type', 'bytes_types',
            'bytes', 'long', 'input', 'decodebytes', 'encodebytes',
@@ -10,6 +11,9 @@ __all__ = ['PY2', 'string_types', 'integer_types', 'text_type', 'bytes_types',
 PY2 = sys.version_info[0] < 3
 
 if PY2:
+    import __builtin__ as builtins
+    import locale
+
     string_types = basestring  # NOQA
     text_type = unicode  # NOQA
     bytes_types = str
@@ -19,8 +23,6 @@ if PY2:
     input = raw_input  # NOQA
     decodebytes = base64.decodestring
     encodebytes = base64.encodestring
-
-    import __builtin__ as builtins
 
 
     def bytestring(s):  # NOQA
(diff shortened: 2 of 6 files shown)


bz2:
Travis on pypy failed hard on an untouched test:

```
test_2_big_file (tests.test_sftp_big.BigSFTPTest) ... ........ 7s 

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

Perhaps it's right to remove the pypy target from `.travis.yml` for this series?

ploxiln:
It's an intermittent failure, a re-test is likely to pass. With an updated pypy it might be more reliable. Travis-CI hasn't updated the available versions of pypy in quite a while, unfortunately ...

bitprophet:
Confirm, pypy sometimes times out on Travis, it's unfortunate but hard to avoid. Typically if everything-but-pypy passes you're gtg.

bz2:
Is there anything I can do to help you with the landing/release process?

jelmer:
Any news on this?

bitprophet:
Rebased this into latest 2.x stable line; it's now out as part of 2.8.1. Thanks!

PR #999, opened 2017-06-15
Don't attempt to obtain server keys if not used

funkyhepcat:
When the parameter "look_for_keys" is False, the connect() method should not attempt to obtain the server key. Note that this fix also works all the way back to Paramiko 1.12.0, where lines 315-332 should be indented under "if look_for_keys:".
We have been using Paramiko with expensive Raritan Power Strips that are SSH enabled. These devices use "username" and "password" authentication. We have thousands of test devices on a subnet, and this proves to be the most convenient automated switching to use, and it's safe.
With our fix, connect() works correctly, and has been using this fix for 4 years. Please add it ASAP, as we're now installing your product using "pip install paramiko".  Thanks much!

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -365,7 +365,7 @@ class SSHClient (ClosingContextManager):
 
         t.start_client(timeout=timeout)
 
-        if not self._transport.use_gss_kex:
+        if not self._transport.use_gss_kex and look_for_keys:
             server_key = t.get_remote_server_key()
             if our_server_keys is None:
                 # will raise exception if the key is rejected


funkyhepcat:
PLEASE DO observe my 2-word change! It is simple, and absolutely necessary. Paramiko is broken when NOT using keys at all (server has no keys present), and using "username", "password" instead.

ploxiln:
`look_for_keys` is intended to prevent looking for, and trying to use, user auth keys. At least some (if not most) users who are already setting `look_for_keys=False` still want the host key to be checked (before their password is potentially passed to a man-in-the-middle attacker). So, this purpose would need its own separate option.

However, I'd suggest using the Transport directly, similar to how SSHClient.connect() uses it, for this unusual use-case.

funkyhepcat:
...Thank you for the response!!  Very kind of you.

I did indeed use the transport directly in the past, it just seemed like a 'hack' at the time as I felt I should be using 'connect'.

I'll look for my old code.  >:-)

Thanks,
Gary

funkyhepcat:
Using the Transport directly worked great. Here's a code snippet for anyone who desires to do the same.
# ==================
# create the connection object
ssh_obj = paramiko.Transport((myIPaddress, 22))
ssh_obj.connect(username="myUser", password="myPassword")
ssh_channel = ssh_obj.open_session()

# create cmd to switch outlet, run it, get stdout
runCmd = "set /system1/outlet14 powerState=off"
ssh_channel.exec_command(runCmd)
ssh_response = str(ssh_channel.recv(2000))
print('INFO: SSH response = ' + ssh_response)

# close session on exit
if hasattr(ssh_obj, "close"):
    ssh_obj.close()
# ================

PR #1009, opened 2017-07-03
getcwd return none

zhouronghua:
when use paramiko, it is not easy to understand cwd is none
so we can give a default cwd to user

### DIFF
--- a/paramiko/sftp_client.py
+++ b/paramiko/sftp_client.py
@@ -626,6 +626,7 @@ class SFTPClient(BaseSFTP, ClosingContextManager):
         .. versionadded:: 1.4
         """
         # TODO: make class initialize with self._cwd set to self.normalize('.')
+        if not self._cwd: chdir('.')
         return self._cwd and u(self._cwd)
 
     def _transfer_with_callback(self, reader, writer, file_size, callback):


ploxiln:
With this change, the functionality contradicts the documentation, which says

> If no directory has been set with `chdir`, this method will return `None`.

ploxiln:
(and it is probably a backwards-incompatible change)

bitprophet:
Backwards incompat, yea. Also, see the TODO: comment above the added line - probably want to use `normalize`. (I don't recall what that does exactly, but I defer to whichever past individual put that comment there :D)

zhouronghua:
YES, I know it is backward incompatible, but i don't know when will user want to get a "none" when he call getcwd.
If getcwd return "none" is useless, why we don't make a change?
If getcwd returns "none" means force our user to call "chdir" first before "getcwd", why we make our users(NEW users, our old users maybe have swallowed the bitter fruit) to do another step which they may not want to.
AND getcwd return current directory is compatible to traditional ftp client.

Users like us use ftp and sftp as well, may not want to have different routine to get one same file.

For paper work, i think we can modify it, too.

lighth7015:
There is a packet you can send to the server to find out the CWD.

fiendish:
@bitprophet I can tell you based on a recent experience that some servers really need that TODO comment to be done because they will not let you do certain operations on relative paths. I cannot give specific names because of NDA, but a colleague and I just spent ages figuring out that `sftpclient.listdir_attr(foo)` would only work on a specific server after first setting `sftpclient._cwd = sftpclient.normalize(".").encode()` because the server refused `CMD_OPENDIR` (which listdir_attr starts with) on the path `"."`.

bskinn:
Closing this PR in favor of #2223

bskinn:
Right. #2223 is an issue, not a PR. Sorry about that.

fiendish:
> Right. #2223 is an issue, not a PR. Sorry about that.

The PR is #2224

bskinn:
Ugh, typos.

Yes, closing here in favor of #2224.

PR #1013, opened 2017-07-05
server: Support pre-authentication banners

seveas:
In the same project that caused me to file #946, I now need to send pre-authentication banners. Wasn't too hard to implement and I think it may be useful for others, hence another pull request :)


ploxiln:
looks clean and small 👍

bitprophet:
Some tests would be rad but otherwise looks good, thanks!

PR #1025, opened 2017-07-25
Adding connect_timeout kwargs in Transport __init__

martintaleski:
I had a problem where I would create a Transport object with a wrong port in socket tuple argument, and the object initialization would hang indefinitely.

I was using the Transport object in order to connect to an SFTP server, as per this : 
https://stackoverflow.com/questions/3635131/paramikos-sshclient-with-sftp

I have added a connect_timeout kwargs in the __init__ method. 

If you don't like kwargs, I can use a positional argument.

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -288,7 +288,7 @@ class Transport(threading.Thread, ClosingContextManager):
             arguments.
         """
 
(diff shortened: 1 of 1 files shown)


bskinn:

Thanks!

PR #1036, opened 2017-08-06
Use `client.exec_command` in forwarding example

fedordikarev:


ploxiln:
"session" is a [Channel](http://docs.paramiko.org/en/2.2/api/channel.html#paramiko.channel.Channel) which does have an `exec_command()` method - is there some other reason why it is not appropriate for this example?

fedordikarev:
I can't get (stdin, stdout, stderr) from session.exec_command:
> Traceback (most recent call last):
>   File "test.py", line 20, in <module>
>     stdin_s, stdout_s, stderr_s = session.exec_command(cmd+"session")
> TypeError: 'NoneType' object is not iterable

And if I run my code without `recv_exit_status` then nothing happens:
> cmd = "ssh {0} touch with_".format(remote_host)
> session.exec_command(cmd+"session")
> if len(sys.argv) > 1:
>   stdin, stdout, stderr = client.exec_command(cmd+"client")
>   retcode = stdout.channel.recv_exit_status()
>   print(retcode)
>   print(stdout.readlines())

> python test.py && ssh tegra ls with\*; python test.py client && ssh tegra ls with\*
> ls: cannot access with*: No such file or directory
> Killed by signal 1.
> 0
> []
> with_client
> with_session
> Killed by signal 1.

ploxiln:
`Channel.exec_command()` does not return the stdin, stdout, stderr pipes. Instead, the channel has methods like `recv()`, `makefile()`, and `recv_exit_status()`. You can see that the file "with_session" was created by the second run, the first run didn't work because you didn't wait for the command to complete in any way (but in the second, you did, indirectly.)

If you use `exec_command()` on the client instead of on the channel, then the `AgentRequestHandler`, which was applied to the channel, would not apply to the command. That sort of defeats the purpose of this this example code for `AgentRequestHandler`.

fedordikarev:
It's strange but my code works when I use `client.exec_command()`:

> \# Connect
> client = SSHClient()
> client.set_missing_host_key_policy(MissingHostKeyPolicy())
> client.connect(host, port, username)
> \# Obtain session
> session = client.get_transport().open_session()
> \# Forward local agent
> **if len(sys.argv) > 1:
>   AgentRequestHandler(session)**
> \# Commands executed after this point will see the forwarded agent on
> \# the remote end.
> remote_host="..."
> cmd = "ssh {0} touch with_".format(remote_host)
> if True:
>   stdin, stdout, stderr = client.exec_command(cmd+"client")
>   retcode = stdout.channel.recv_exit_status()
>   print(retcode)
>   print(stdout.readlines())

> ssh tegra rm -v with\*; python test.py agent; ssh tegra ls with\*
> removed ‘with_client’
> Killed by signal 1.
> 0
> []
> with_client
> Killed by signal 1.

> ssh tegra rm -v with\*; python test.py ; ssh tegra ls with\*
> removed ‘with_client’
> Killed by signal 1.
> 255
> []
> ls: cannot access with*: No such file or directory
> Killed by signal 1.

PR #1041, opened 2017-08-19
Common up break out of Transport.run() loop

radssh:
Can’t seem to reason out any advantage of clearing self.active and
calling self.packetizer.close() in these situations instead of simply
breaking out of loop and allowing the additional conditional cleanups
to be done. Currently looking into tackling some needed cleanup in
auth_handler, and not having the auth_handler.abort() called on server
disconnect feels like a bug - who knows?

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -1825,8 +1825,6 @@ class Transport(threading.Thread, ClosingContextManager):
                         continue
                     elif ptype == MSG_DISCONNECT:
                         self._parse_disconnect(m)
-                        self.active = False
-                        self.packetizer.close()
                         break
                     elif ptype == MSG_DEBUG:
                         self._parse_debug(m)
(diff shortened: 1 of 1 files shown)


bitprophet:
Thanks for poking at this!

Have you looked at the git blame history for these lines to see when/why they were added? Packetizer close calls are often aligned with hard-to-troubleshoot hang issues so I get really twitchy when anyone touches them :grin:

If these lines are in the class of "existed since time immemorial" (like much of the codebase, dating from e.g. 2004-ish) and we can eg run some simple tests in a loop for a long time w/o encountering issues, probably fine; but if anyone's added or moved these lines in the last few years, probably requires more thought.

radssh:
I did not dig into the code history initially (but should have). Jumping in the way-back machine to https://github.com/paramiko/paramiko/blob/a22719c338decc9cad28992cad7dc04e9e0696d7/paramiko/transport.py#L991-L1017 the pertinent part of that section looks surprisingly untouched in the last 12 years!

There are additions to the end of loop cleanup code since then, which appear to be reasonable enough additions, with conditional `if`s and `try:`, but (shock) not matching the steps listed in `Transport.close()` or `Transport.stop_thread()`, which may be leading to the thread hanging joy.

Looking at the code base then, and seeing that `Transport` was a base class that `auth_transport.py` was building on to bolt-on authentication, goes a long way toward explaining why the authentication logic is currently so convoluted 😈

bitprophet:
Color me surprised that the last few years' worth of periodic hang/shutdown bugfixes didn't actually catch everything. So this falls under my 1st category, meaning, sure, let's try it. For funsies I'll run the test suite in your branch in an error-counting loop thing I have, just in case.

bitprophet:
Ran test suite 100 times in a row, zero errors of any kind. worksforme.

Issue #989, opened 2017-06-09
Add timeout to global_request (same as open_channel)


shahargl:
In case of global_request sent (e.g. "Sending global request "direct-tcpip"") but somehow, the server does not response to the packet, global_request will hang forever.

There is easy solution - do it exactly like open_channel does it.

bitprophet:
Rolling into PR #1047, thanks!

PR #1047, opened 2017-09-01
Global request timeouts #989

hvltaj:
Adds timeout to global_request in Transport.
Tests included.

Implements #989

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -1009,7 +1009,7 @@ class Transport(threading.Thread, ClosingContextManager):
             return x.global_request('keepalive@lag.net', wait=False)
         self.packetizer.set_keepalive(interval, _request)
 
-    def global_request(self, kind, data=None, wait=True):
+    def global_request(self, kind, data=None, wait=5):
         """
         Make a global request to the remote host.  These are normally
         extensions to the SSH2 protocol.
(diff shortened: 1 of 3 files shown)


tests/test_transport.py
```diff
@@ -927,3 +929,49 @@ def test_channel_send_memoryview(self):
...
             self.assertEqual(sfile.read(len(data)), data)
+
+    def test_global_request_timeout(self):
+        """
+        verify global_request timeouts.
+        """
+        self.setup_test_server()
+
+        with mock.patch.object(self.server, 'check_port_forward_request',
+                               new=lambda x, y: time.sleep(2)):
```
bz2: Should not use `time.sleep()` in tests. With the real timeout of 1 second below, that ensures this test will always be slow. Can either mock `time.time()` in the function for the same effect, or pass in a timer at some point.

paramiko/transport.py
```diff
@@ -1021,10 +1021,16 @@ def global_request(self, kind, data=None, wait=True):
         :param bool wait:
             ``True`` if this method should not return until a response is
             received; ``False`` otherwise.
+        :param float timeout:
+            optional timeout making a global request, no default set
```
bz2: I realise this is somewhat modelled on `open_channel()`, but having both `wait=` and `timeout=` is weird. Timeout is a superset of the wait behaviour, but means nothing without it.

philusky:
Timeout can be passed in the ''wait'' argument now.
I set up the default timeout to 5s, it will work as well if someone passes 'wait=True'.
I mocked time.time() calls to be 6 second apart.

ploxiln:
You should rebase/squash such that only a couple relevant git commits appear in this PR.

paramiko/transport.py
```diff
@@ -1038,13 +1043,17 @@ def global_request(self, kind, data=None, wait=True):
         self._send_user_message(m)
         if not wait:
             return None
-        while True:
+        # sets default timeout to 5 in case ``wait`` == True
+        timeout = 5 if isinstance(wait, bool) else wait
```
ploxiln: This is a bit awkward - there's no elegant value to mean "forever" (which might be useless), and 5 is written in two places.

Here's another idea: let `True` mean "forever", and change all the places in paramiko.transport which call `global_request(... wait=True)` to just not specify wait, allowing the default to be used. Then, perhaps in a later PR, the appropriate timeout configured in the Transport could be specified when `global_request()` is called.

ploxiln: Also, the other timeouts in Transport have higher default values, 15 or 30 seconds.

bskinn:

Thanks!

PR #1048, opened 2017-09-01
always log socket.error with str() or repr()

### LINKED ISSUE #1046: IndexError: tuple index out of range in  transport.py" line 1878
File "/Library/Python/2.7/site-packages/paramiko/transport.py", line 1878, in run
    emsg = '%s (%d)' % (e.args[1], e.args[0])
IndexError: tuple index out of range

ploxiln:
on rare occasion, assuming the tuple had two parts caused an exception

but no need to format the 2-tuple into a message anyway,
the class can do that itself

still fallback to repr() because I think I've seen empty str() result,
on very rare occasion

fixes #1046  and also just simplifies the code

I based this on 2.0 but it should be fine to re-target it to a later 2.x branch

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -372,7 +372,7 @@ class Transport(threading.Thread, ClosingContextManager):
                     try:
                         retry_on_signal(lambda: sock.connect((hostname, port)))
                     except socket.error as e:
-                        reason = str(e)
+                        reason = str(e) or repr(e)
                     else:
                         break
             else:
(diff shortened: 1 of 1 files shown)


bz2:
Feels like this is the kind of change that needs to come with a unit test with the sort of `socket.error` instances that should be handled. The `str(e) or repr(e)` formulation is odd, and I'm not convinced it's better than just using a literal string.

```
>>> "socket error: %s" % socket.error()
'socket error: '
>>> "socket error: %s" % socket.error('broken')
'socket error: broken'
>>> "socket error: %s" % socket.error(111, 'Connection refused')
'socket error: [Errno 111] Connection refused'
```

ploxiln:
The "or repr(e)" was just left in out of a sense of paranoia. I think I've seen strange blank error strings before - but I don't know what they were, or if it's possible for `socket.error`. Bad example:

```
>>> repr(socket.error())
'error()'
>>> str(socket.error())
''
```

So I thought that maybe the reason "or repr(e)" was already in there was as a work-around for that. Anyway, I'd be OK with removing it.

bz2:
Sorry, not clear enough, my suggestion is `str(e) or '<empty socket error>'` or similar, given you know what condition you're hitting that in.

ploxiln:
Ah, that makes sense. `repr()` might give you a bit more info in the form of a subclass name. maybe `"<empty>"` is better then `"error()"` though ...

bz2:
For the `socket.timeout()` case specifically that original formulation is better, so maybe it's still clearer overall like that?

Anyway, change in either form seems like a slight improvement.

ploxiln:
ok, back to the original "or repr(e)" - it does work better for `socket.timeout()`

ploxiln:
side note: CI of 2.2 branch is broken because it tries to use new pytest-4.6.3 with old python-3.3 and python-2.6

Issue #1001, opened 2017-06-21
Add support for wildcards in known_hosts


OOTS:
Please add support for wildcards in `known_hosts` files.

For example, a known_hosts file like the following is accepted by running plain `ssh`:
    `[?xample.net]:42 ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAICdQrznEieszf9lLMzYnc+0i44nUCvhActa4dcf5C65q`

You can verify this by running `ssh -p 42 example.net` or `ssh-keygen -F [example.net]:42`.

Paramiko, however, throws an exception:

    $ python3
    >>> import paramiko
    >>> client = paramiko.SSHClient()
    >>> client.load_system_host_keys() # or client.load_host_keys(filename)
    >>> client.connect("example.net", port=42)
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/usr/local/lib/python3.5/dist-packages/paramiko/client.py", line 373, in connect
        self, server_hostkey_name, server_key
      File "/usr/local/lib/python3.5/dist-packages/paramiko/client.py", line 700, in missing_host_key
        raise SSHException('Server %r not found in known_hosts' % hostname)
    paramiko.ssh_exception.SSHException: Server '[example.net]:42' not found in known_hosts

I'm using paramiko 2.2.1.

Please add support for the `?` and `*` wildcards (for host names as well as for ports) in `known_hosts` files, as documented in the [sshd man page](https://linux.die.net/man/8/sshd).

sanjeevrohila:
Is it implemented ?

backbord:
There is a 5 year old PR https://github.com/paramiko/paramiko/pull/1053 -- can it be revived?

filipeg:
There is also this more recent PR #1920

PR #1053, opened 2017-09-05
Add support for wildcards in known_hosts #1001

philusky:
Added support for two wildcards: '?' and '*' in known_hosts as well as negation: '!' - same as in sshd.


paramiko/hostkeys.py
```diff
@@ -196,14 +197,34 @@ def _hostname_matches(self, hostname, entry):
...
+        """
+        wildcard_negation = False
+        if hostname.startswith('|1|'):
+            return False
+        if wildcard.startswith("!"):
+            wildcard_negation = True
+
+        wildcard = wildcard.lstrip("!")
+        wildcard = re.escape(wildcard)
+        wildcard = wildcard.replace("\\?", ".?")
```
ploxiln: `.?` means "0 or 1 of any character" but what you want here is just `.` "exactly one of any character"

paramiko/hostkeys.py
```diff
@@ -196,14 +197,34 @@ def _hostname_matches(self, hostname, entry):
...
+        :param str hostname: target host name
+        :return bool:
+        """
+        wildcard_negation = False
+        if hostname.startswith('|1|'):
+            return False
+        if wildcard.startswith("!"):
+            wildcard_negation = True
+
+        wildcard = wildcard.lstrip("!")
```
ploxiln: It would be more efficient to put this in the above if body and just slice off the first char:

```python
if wildcard.startswith("!"):
    wildcard_negation = True
    wildcard = wildcard[1:]
```

paramiko/hostkeys.py
```diff
@@ -196,14 +197,34 @@ def _hostname_matches(self, hostname, entry):
...
+        wildcard = re.escape(wildcard)
+        wildcard = wildcard.replace("\\?", ".?")
+        wildcard = wildcard.replace("\\*", ".*")
+
+        return re.match(wildcard, hostname) and wildcard_negation is not True
```
ploxiln: It is simpler to write `and not` instead of `is not True`. But we see here that this can't really support negation properly - it just disables the pattern completely. You might as well `return False` above for `if wildcard.startswith("!")`. To support negation properly you would have to make a match here cause all other hostnames/patterns on the same line to also fail to match. For that, you would need a whole different way to parse and handle `known_hosts` in paramiko.

philusky:
I really got lost in git ;/ I squashed all commits into one, but some merges happened in the way and there are more than just my changes now. Any tip how to fix that?
Anyhow, I changed `.?` to `.`, removed lstrip() and actually was able to implement proper negation thanks to the fact that `HostKeyEntry` stores all host names / patterns from a single line in `known_hosts`. I wonder if the exception approach is a good one.

ploxiln:
That does look like it would work. I'm not sure if `LookupError` is the best exception to use here, maybe a custom one should be created for this.

PR #1060, opened 2017-09-12
Transport: fix the preferred KEX algorithms for gssapi-keyex (2.0)

akruis:
Add additional KEX algorithms for gssapi-keyex in front of the
default preferred KEX algorithms, if gssapi-keyex is enabled.

Before this change, Transport used a hard coded (and out-dated) list of
algorithms, if gssapi-keyex was enabled.

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -132,6 +132,11 @@ class Transport(threading.Thread, ClosingContextManager):
         'diffie-hellman-group-exchange-sha1',
         'diffie-hellman-group-exchange-sha256',
     )
+    _preferred_gsskex = (
+        'gss-gex-sha1-toWM5Slw5Ew8Mqkay+al2g==',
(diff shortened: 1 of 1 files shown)


ploxiln:
It's good to see gss get some testing and maintenance :)

bitprophet:
Torn on whether this should be considered a bugfix or a "feature" (it's technically adding new functionality that did not previously exist.) What's the impact (functionality or security wise) on users if this only went out in 2.3?

ploxiln:
I'd lean towards just considering this a bugfix for gss.
Note also that `_preferred_kex` was updated in 2.2

ploxiln:
(for the 2.2.0 release, so consistent with your idea that kex preference changes should be in minor releases - I'm just noting it :)

akruis:
It a bug fix. Without this fix you get an outdated set of non gss-kex algorithms, if you enable gss-kex.

PR #1062, opened 2017-09-12
demo_simple: gss_auth and gss_kex are independent of each other.

akruis:
Simplify the demo script.


bitprophet:
FWIW I don't see why this can't go back to 2.0, so I am cherry-picking it back there & merging up. Thanks!

PR #1063, opened 2017-09-13
Fix hostname canonicalization with GSSAPI

SebastianDeiss:
I added the parameter ``gss_trust_dns`` in ``client.py`` and ``transport.py``. Set by default to ``True`` the parameter indicates whether or not the DNS is trusted to securely canonicalize the hostname of the target host. If set to ``False`` the hostname entered will be passed to GSSAPI.
This option behaves like ``GSSAPITrustDNS`` from OpenSSH.

Also, the option ``gss_host``, if set, can be used to override the hostname for GSSAPI and therefore is considered as the equivalent to ``GSSAPIServerIdentity`` from OpenSSH.

This pull should fix issue #915 and it supersedes pull #919 .

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -228,6 +228,7 @@ class SSHClient (ClosingContextManager):
         gss_host=None,
         banner_timeout=None,
         auth_timeout=None,
+        gss_trust_dns=True
     ):
         """
         Connect to an SSH server and authenticate to it.  The server's host key
@@ -277,6 +278,9 @@ class SSHClient (ClosingContextManager):
         :param bool gss_deleg_creds: Delegate GSS-API client credentials or not
         :param str gss_host:
             The targets name in the kerberos database. default: hostname
+        :param bool gss_trust_dns: Indicates whether or not the DNS is trusted to
+                              securely canonicalize the name of the host being
+                              connected to (default ``True``).
         :param float banner_timeout: an optional timeout (in seconds) to wait
             for the SSH banner to be presented.
         :param float auth_timeout: an optional timeout (in seconds) to wait for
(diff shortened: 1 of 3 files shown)


paramiko/client.py
```diff
@@ -228,6 +228,7 @@ def connect(
         gss_host=None,
         banner_timeout=None,
         auth_timeout=None,
+        gss_trust_dns=True
```
ploxiln: This should probably be next to the other gss args - I think it's pretty unlikely that anyone is depending on the args order instead of using kwargs by midpoint of this args list.

I also wonder if most of the gss args can be boiled down to a single "gss context" argument. (For backwards compat it would still need an additional kwarg here, but it would be the last one for gss ;)

bitprophet:
Thanks for this. Have a branch with cleanup going; the thing I'm paused on right now is the very similar, but almost inverted, bits of logic around how to call `transport.set_gss_host`, one in Client and one in Transport. I'd really like these to be refactored because it makes the logic hard to follow & seems like an excellent way for bugs to get introduced.

(This kinda follows from my earlier notes re: just not liking how this API is spread across Client/Transport to begin with, though that is again not the OP's fault :))

bitprophet:
(I'm still working on it, btw!)

paramiko/client.py
bitprophet: I think for now the (yes, minor) benefit of preserving backwards compat arg order is worth leaving this alone, but partly _because_ I agree that later we should probably compress them into a single argument. (Counterpoint: doing that makes it harder to do things like create partial applications of the function...but I'm gonna guess not many people are doing that.)

bitprophet:
OK I have my changes published at https://github.com/paramiko/paramiko/compare/1063-int - please let me know if the 'new' `Transport.set_gss_host` seems like it still does the right thing from your perspective, @SebastianDeiss - thanks!

bitprophet:
It's not listed here because my branch isn't technically related (I gotta figure out a better way to handle that...) but Sebastian +1'd my tweaks inline - https://github.com/paramiko/paramiko/compare/1063-int#commitcomment-24299580 - so I'm merging this to master.

bitprophet:
EDIT: merging to master did not trigger this PR as 'merged' since it was originally set on 2.2. (We're adding a new kwarg so it's a new feature, so it'll be in 2.3, not 2.2 :)) Closing manually. Thanks!

PR #1064, opened 2017-09-18
Fix gss_host setting and cleanup its logic

SebastianDeiss:
The parameter ``kex_requested`` is misleading, since setting ``gss_host``
is also required for gssapi-with-mic.  
Also, I changed the parameter order of ``set_gss_host()`` to preserve backwards compatibility.


bitprophet:
I admit I was wondering at the time whether this was gss_kex specific, but since I'm pretty ignorant about GSSAPI I kinda assumed it was all good. Thanks for the update!

PR #1059, opened 2017-09-12
SSHClient: fix the host key test (2.2)

akruis:
Same as pull request #1057, but for paramiko 2.2.

PR #1066, opened 2017-09-18
Transport.connect(): fix host key test

SebastianDeiss:
Skip the host key check in ``Transport.connect()`` only, if the transport actually used gssapi-keyex.

Before this change, a MITM attack on the paramiko ssh client with ``gss_kex=True`` was possible by having a server that does not support gssapi-keyex and gives any or no host key.

This pull fixes the same issue as #1059, but for ``Transport.connect()``.

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -1125,7 +1125,7 @@ class Transport(threading.Thread, ClosingContextManager):
         # check host key if we were given one
         # If GSS-API Key Exchange was performed, we are not required to check
         # the host key.
-        if (hostkey is not None) and not gss_kex:
+        if (hostkey is not None) and not self.gss_kex_used:
             key = self.get_remote_server_key()
             if (
                 key.get_name() != hostkey.get_name() or


paramiko/transport.py
```diff
@@ -1125,7 +1125,7 @@ def connect(
         # check host key if we were given one
         # If GSS-API Key Exchange was performed, we are not required to check
         # the host key.
-        if (hostkey is not None) and not gss_kex:
+        if (hostkey is not None) and not self.gss_kex_used:
```
ploxiln: is this "too early"? might `gss_kex_used` be set just below via `self.auth_gssapi_keyex(username)`?

SebastianDeiss: ``gss_kex_used`` is not set in ``transport.py``. It is set in ``kex_gss.py``. The host key check is performed during key exchange (or not, if GSS-API is used) and therefore must be done in ``Transport.connect()``.

In ``Transport.auth_gssapi_keyex(username)`` the user is authenticated if a successful GSS-API key exchange was performed. That's a different layer in the SSH protocol.

ploxiln: I see, makes sense (gss is setup via `start_client()` -> `start()` -> `run()` -> `_send_kex_init()` / `_expect_packet()`) - I'm not familiar with gss, just clarifying :)

PR #1075, opened 2017-09-26
sftp: fix BytesWarning: str() on a bytes instance

Rezney:
when running the code with python3 we get BytesWarning as what we
receive from hexlify() function is bytes. Use u() from py3compat to
overcome the issue.

### DIFF
--- a/paramiko/sftp_client.py
+++ b/paramiko/sftp_client.py
@@ -341,7 +341,7 @@ class SFTPClient(BaseSFTP, ClosingContextManager):
         handle = msg.get_binary()
         self._log(
             DEBUG,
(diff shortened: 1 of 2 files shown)


Rezney:
This helped me to get rid of the warnings. Please let me know if it is ok.

bitprophet:
Looks like an OK fix to me, thanks!

FWIW, pytest seems to have a bunch of other errors when it runs our test suite (it is extremely old and barely even qualifies as correct unittest usage) and I have my own todo list item for porting it to a newer pytest-driven plugin I'm writing. So keep that in mind!

PR #1086, opened 2017-10-05
Added timeout when calling wait on thred

HenkeA:
When connected to a ssh server that goes down during call to
'paramiko.client.SshClient.exec_command', it can cause a hang during
the call to 'wait' on the event thread. 'wait' does take a timeout, and
such does 'exec_command', but the timeout in exec_command is never
passed down to 'wait'. This patch does so.

Change-Id: I70347eac422ce6b060d159eeb6fe9e862d574ca7


bskinn:

Thanks!

PR #1094, opened 2017-10-14
Remove old Python 3.2 code

hugovk:
Update the only line in the entire program that had to be treated specially to support the long-gone Py3.2 :)


bitprophet:
Nice catch! I didn't think to search for 3.2, only 2.6 and 3.3 :D thanks!

PR #1103, opened 2017-10-29
Use sha256 digest for key fingerprint

ioparaskev:
- Changes the `get_fingerprint` method of `PKey` to return sha256
base64-trimmed digest instead of md5 digest. This allows better collision
resistance. For more info about fast collision attacks on MD5 check
the paper "Fast Collision Attack on MD5" by Tao Xie1, Fanbao Liu,
Dengguo Feng
- Adds methods `get_sha256_fingerprint`, `get_sha512_fingerprint`,
`get_md5_fingerprint` to support different algorithms for calculating
key fingerprint. sha1 fingerprint calculation is not added due to recent
practical collision findings while md5 is still provided for backwards
compatibility.
- Method `get_fingerprint` now calls `get_sha256_fingerprint`
which is the default behavior of OpenSSH since version 6.8


ploxiln:
ssh-add from OpenSSH-7.4p1 shows auth key fingerprints like so:

```
$ ssh-add -l
3072 SHA256:k1kzdj2A9DXxs9cc7MRd9TaIzzbcPvqjpEHZtq/wwfk pierce@plo-pro.local (RSA)
```

Just to say, this PR is consistent with that, and thus seems like a generally good idea.

But this will likely be deferred to a major release because it's a "breaking" output-format change for the `get_fingerprint()` method.

ioparaskev:
@ploxiln I agree
Maybe we could keep the fingerprint being md5 but also provide 2 methods for sha512 and sha256?
I also need to make changes for python3 compatibility (forgot to check before making the PR)

ioparaskev:
Changes made, now it's py2/3 compatible.
As I said, if we consider this a "breaking" change, maybe it's better to make this a 2-part change to avoid delaying the extra methods until a major release.
1. Provide sha256  & sha512 fingerprint support in minor release
2. In a major release, push the `get_fingerprint` to use the sha256 method instead of the md5

ioparaskev:
any updates for this?

avidspartan1:
Would love to see this PR accepted.

ioparaskev:
I'm not sure why after all these years this is still unmerged but if the reason is backwards compatibility, I think we can figure a way to roll this change (maybe still keeping the md5 as default behaviour?). It's really unfortunate that in 2020 paramiko is still stuck with md5

subodh2691:
I am facing similar issue. can someone help when these changes can be merged?

guiri14:
I would also like to see this merged. We use paramiko at work, and since moving to FIPS this has become a big issue. It looks like discussion about this has been on-going since 2014.

avidspartan1:
@ioparaskev, looks like there are merge conflicts. I'm not sure if that's what's standing in the way, but do you have time to fix those? I'd take a whack at it but I don't have a great way to test at the moment.

ioparaskev:
@avidspartan1 I'm not sure since I never got any feedback on why this can't be merged.
In any case, I decided to rebase this PR to be in sync with the latest HEAD and I also decided to keep the md5 behaviour as the default for now.
This means that this PR can be merged without introducing any breaking changes and at the same time adding extra options of using the sha256/sha512 hashes. I decided to make this change since from what I understand, changing the default behaviour might break a lot of things for people who relied on the bad md5. So this PR is actually a type of "grace period" for people to convert to the newer algorithms before introducing a breaking change. 
I don't know if this was the show-stopper for merging this PR but in any case I considered it better to split this into two parts. So hopefully when/if this is merged a roadmap for rolling out to sha256 can be set.

paramiko/pkey.py
```diff
@@ -168,17 +168,74 @@ def can_sign(self):
...
+
+    def get_sha512_fingerprint(self):
+        """
+        Returns an openssh SHA512 fingerprint of the public part of this key.
+        SHA512 fingerprint is calculated by base64 encoding the sha512 digest
+        of the key and then removing the two base64 trailing == chars
+        Nothing secret is revealed.
+
+        :return:
+            an 86-byte `string <str>` (base64) of the SHA512 fingerprint,
```
mateuszstompor, line 192:
```suggestion
            a 86-byte `string <str>` (base64) of the SHA512 fingerprint,
```

paramiko/pkey.py
```diff
@@ -168,17 +168,74 @@ def can_sign(self):
...
-    def get_fingerprint(self):
+    def get_sha256_fingerprint(self):
+        """
+        Returns an openssh SHA256 fingerprint of the public part of this key.
+        SHA256 fingerprint is calculated by base64 encoding the sha256 digest
+        of the key and then removing the base64 trailing == chars
+        Nothing secret is revealed.
+
+        :return:
+            an 43-byte `string <str>` (base64) of the SHA256 fingerprint,
```
mateuszstompor, line 179:
```suggestion
            a 43-byte `string <str>` (base64) of the SHA256 fingerprint,
```

paramiko/pkey.py
```diff
@@ -168,17 +168,74 @@ def can_sign(self):
...
+    def get_fingerprint(self, algorithm="md5"):
+        """
+        Returns an openssh fingerprint of the public part of this key.
+        Nothing secret is revealed.
+
+        :return:
+            Depending on the algorithm variable can return
+            a 16-byte `string <str>` (binary) of the MD5 fingerprint, in SSH
+            format. (default)
+            an 43-byte `string <str>` (base64) of the SHA256 fingerprint,
```
mateuszstompor, line 232:
```suggestion
            a 43-byte `string <str>` (base64) of the SHA256 fingerprint,
```

paramiko/pkey.py
```diff
@@ -168,17 +168,74 @@ def can_sign(self):
...
+            in SSH format.
+            an 86-byte `string <str>` (base64) of the SHA512 fingerprint,
```
mateuszstompor, line 234:
```suggestion
            a 86-byte `string <str>` (base64) of the SHA512 fingerprint,
```

paramiko/pkey.py
```diff
@@ -168,17 +168,74 @@ def can_sign(self):
...
         :return:
             a 16-byte `string <str>` (binary) of the MD5 fingerprint, in SSH
             format.
         """
         return md5(self.asbytes()).digest()
 
+    def get_available_fingerprint_methods(self):
+        """
+        Return a dictionary of available hash algorithms and  their method
+        callbacks
```
mateuszstompor, line 211:
```suggestion
        callbacks.
```

paramiko/pkey.py
```diff
@@ -168,17 +168,74 @@ def can_sign(self):
...
+
+        :return:
+            a `dictionary <dict>` containing the available hash algorithms
+            and their method callbacks
```
mateuszstompor, line 215:
```suggestion
            and their method callbacks.
```

paramiko/pkey.py
ioparaskev: Please let me know when you finish these review comments. I hate to create multiple commits for such minor remarks as missing `.` in docstrings

paramiko/pkey.py
```diff
@@ -168,17 +168,74 @@ def can_sign(self):
...
+            a 86-byte `string <str>` (base64) of the SHA512 fingerprint,
+            in SSH format.
+        """
+        return base64.b64encode(b(sha512(self.asbytes()).digest()))[:-2]
+
+    def get_md5_fingerprint(self):
         """
         Return an MD5 fingerprint of the public part of this key.  Nothing
-        secret is revealed.
+        Nothing secret is revealed.
```
MrMino, line 200: "Nothing" is used twice.

```suggestion
        secret is revealed.
```

aronwk-aaron:
Really need this PR merged for so that FIPS is supported. It's 2020, we shouldn't be using MD5 for anything

dfresh613:
This may be a bit of scope creep for this bug. But it would be very nice if the default AutoAddPolicy and AutoReject were also updated to use these new fingerprints by default, and fall back to md5 only if sha is not available

In addition to updating pkey with additional algorithms, may I also suggest sometime in the future the paramiko https://github.com/paramiko/paramiko/blob/master/paramiko/client.py#L805

spanktar:
Holy moley folks, just push the merge button!

malthe:
Bump?

Also – what about using a fingerprint as a host key. That doesn't seem possible today. A [StackOverflow answer](https://stackoverflow.com/a/46839480/647151) suggests using a custom `FingerprintKey` class.

nchigwamba:
Is there a view of when this PR will be merged?

michalxo:
Hello! Any news on this PR please?

bittner:
@alex Would you be willing to merge this PR if the author resolves the merge conflicts?

bskinn:
I'm not deeply familiar with the code here, but per xxx it looks like paramiko 3.2 introduced a [`fingerprint` property](https://docs.paramiko.org/en/latest/api/agent.html?highlight=fingerprint#paramiko.agent.AgentKey.fingerprint) that does SHA256... but _only_ SHA256.

I figure the changes in this PR will need to be adapted to this new code... it seems likely that either (i) the existing code will need to be converted to a method, so that the desired fingerprint type can be passed as an argument, or (ii) multiple properties will need to be created, one for each desired fingerprint type.

I'm not sure how bitprophet will want to resolve this. I've flagged it for his policy decision.

mnasiadka:
> I'm not deeply familiar with the code here, but per xxx it looks like paramiko 3.2 introduced a [`fingerprint` property](https://docs.paramiko.org/en/latest/api/agent.html?highlight=fingerprint#paramiko.agent.AgentKey.fingerprint) that does SHA256... but _only_ SHA256.
> 
> I figure the changes in this PR will need to be adapted to this new code... it seems likely that either (i) the existing code will need to be converted to a method, so that the desired fingerprint type can be passed as an argument, or (ii) multiple properties will need to be created, one for each desired fingerprint type.
> 
> I'm not sure how bitprophet will want to resolve this. I've flagged it for his policy decision.

Hello - it's been over a year without a decision - @bskinn is there any way this could progress?

ioparaskev:
I haven't seen any roadmap plans on integrating this functionality so I'm not extremely enthusiastic on refactoring and syncing the PR so that it can sit waiting for another 4+ years tbh

mnasiadka:
> I haven't seen any roadmap plans on integrating this functionality so I'm not extremely enthusiastic on refactoring and syncing the PR so that it can sit waiting for another 4+ years tbh

I rather wanted to get an insight from the repository maintainers - but thanks for replying!

PR #1106, opened 2017-11-03
Detect what authentication types are allowed

abadger:
Some systems are configured to close the connection after one
authentication attempt.  If this is the case and we are provided with
both a pulic key and a password then the password will never be tried.

To fix that, we need to detect the allowed authentication methods and
only try public key or password if those are allowed.

Tested by taking an openssh server and changing the following default values:
```
MaxAuthTries 1
PubkeyAuthentication no
```

Connecting to the remote machine with something like this:
```
import paramiko
ssh = paramiko.SSHClient()
ssh.load_system_host_keys()
ssh.connect('192.168.122.7', username='badger', allow_agent=True, key_filename='/home/badger/.ssh/id_rsa', password='A password'
```

Before this patch is applied, the ssh.connect will fail because publickeys are always tried before the password.  With this patch applied, ssh.connect succeeds because paramiko will check whether publickeys are an allowed method of connecting before attempting to connect with them.


abadger:
Note: in addition to ```SSHClient``` having a connect method, I also saw that ```Transport``` has a connect method.  That one would seem to suffer from the same problems as ```SSHClient.connect``` but I don't have any code to test it so I didn't want to risk changing it as part of this pull request.  If I had to hazard a guess, though, I think something like the following would fix it:
``` python

diff --git a/paramiko/transport.py b/paramiko/transport.py
index 68519f9..1e11dc1 100644
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -1191,16 +1191,31 @@ class Transport(threading.Thread, ClosingContextManager):
             self._log(DEBUG, 'Host key verified ({})'.format(
                 hostkey.get_name()))
 
+        # Detect what authentication methods the server supports (Mirrors what
+        # Openssh client does)
+        try:
+            self.auth_none(username)
+        except BadAuthenticationType as e:
+            allowed_auth_methods = frozenset(e.allowed_types)
+        else:
+            # Server allowed auth with no credentials
+            return
+        # Some servers don't follow the standard and don't return their allowed
+        # auth methods Fall back to trying anything rather than trying nothing
+        if not allowed_auth_methods:
+            allowed_auth_methods = frozenset(('gssapi-keyex', 'gssapi-with-mic',
+                                              'password', 'publickey'))
+
         if (pkey is not None) or (password is not None) or gss_auth or gss_kex:
-            if gss_auth:
+            if gss_auth and 'gssapi-with-mic' in allowed_auth_methods:
                 self._log(DEBUG, 'Attempting GSS-API auth... (gssapi-with-mic)') # noqa
                 self.auth_gssapi_with_mic(
                     username, self.gss_host, gss_deleg_creds,
                 )
-            elif gss_kex:
+            elif gss_kex and 'gssapi-keyex' in allowed_auth_methods:
                 self._log(DEBUG, 'Attempting GSS-API auth... (gssapi-keyex)')
                 self.auth_gssapi_keyex(username)
-            elif pkey is not None:
+            elif pkey is not None and 'publickey' in allowed_auth_methods:
                 self._log(DEBUG, 'Attempting public-key auth...')
                 self.auth_publickey(username, pkey)
             else:
```

deed02392:
@bitprophet Hi Jeff, can you consider merging this? It fixes a few bugs and is a very small patch to change authentication behaviour to more closely match that of OpenSSH.

lk-chen:
Another friendly ping. This is to fix #1448

twojciac:
@bitprophet This would be very useful if this could be merged.

bitprophet:
I either didn't see this or assumed it would clash with #387 somehow (big auth overhaul), but at a glance I'm not sure the latter is really the case. I'll pop this into the next feature milestone for when I have time to read it over / compare to OpenSSH (tho IIRC, this is bang on re: how they do it, as stated).

bitprophet:
> Note: in addition to `SSHClient` having a connect method, I also saw that `Transport` has a connect method. That one would seem to suffer from the same problems as `SSHClient.connect` but I don't have any code to test it so I didn't want to risk changing it as part of this pull request.

I need to refresh my memory of how/whether those overlap. As stated on another ticket recently, Client generally doesn't actually want big feature updates (typically we want to push folks towards Fabric and leave Paramiko focused on the lower level APIs), so if we only add it to one, it might be Transport - but I can figure that out when I merge/hack.

PR #1112, opened 2017-11-10
Add sock_timeout to transport

ashishshukla78:
Need ability to change paramiko transport socket connetion timeout.
Added optional 'sock_timeout' to let user pass timeout in secs.

Verified user supplied timeouts.

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -252,7 +252,8 @@ class Transport(threading.Thread, ClosingContextManager):
                  default_window_size=DEFAULT_WINDOW_SIZE,
                  default_max_packet_size=DEFAULT_MAX_PACKET_SIZE,
                  gss_kex=False,
-                 gss_deleg_creds=True):
+                 gss_deleg_creds=True,
+                 sock_timeout=None):
         """
         Create a new SSH session over an existing socket, or socket-like
         object.  This only creates the `.Transport` object; it doesn't begin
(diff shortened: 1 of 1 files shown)


ploxiln:
Worth mentioning: the equivalent functionality in `SSHClient.connect()` uses the kwarg name "timeout" (though "sock_timeout", which this PR is adding for Transport(), is a bit clearer).

Also, it may be a bit confusing that if you pass an actual socket instead of a host-port tuple, then (as this PR currently works) sock_timeout does not apply to it. Maybe if not None it should be applied either way.

ashishshukla78:
On Nov 9, 2017 6:42 PM, "Pierce Lopez" <notifications@github.com> wrote:
>
> Worth mentioning: the equivalent functionality in SSHClient.connect()
uses the kwarg name "timeout" (though "sock_timeout", which this PR is
adding for Transport(), is a bit clearer).
>
> Also, it may be a bit confusing that if you pass an actual socket instead
of a host-port tuple, then (as this PR currently works) sock_timeout does
not apply to it. Maybe if not None it should be applied either way.
I believe it is more appropriate to apply this when socket is created from
host port tuple for two reasons. First, without this there is no way to
control timeout for subsequent connect and secondly, when actual sock is
passed (instead of host pair), the responsibility to set timeout should be
on the caller; which can be set outside of paramiko transport.
I agree it is tricky to define semantics of this new timeout parameter.

PR #1118, opened 2017-11-14
Clarify how certain timeouts work.

DanLipsitt:
Document explicitly when timeouts aren't relative to when the connection opened.


astrojuanlu:
After digging everywhere for information about timeouts in Paramiko this saved my week. It would be great to have this or a similar clarification in the docs, because it looks like a common source of confusion...

PR #1130, opened 2017-11-28
get_pty is undocument in SSHClient

waghanza:


bitprophet:
Thanks for the catch!

Issue #1115, opened 2017-11-11
Race conditions between creating and sending messages


yieldi:
I get a lot of "Received ieof for nonexistent channel" messages sent from openSSH server when connecting as paramiko client. These messages are due to an EOF being sent after already closing the channel it is sent on. It results in the entire transport being shutdown.

The root cause is a race condition when closing a connection and a shutdown_write in Channel:

```
shutdown_write                      close
------                              -----
lock()
m = create EOF msg
unlock()
                                      lock()
                                      internal_close
                                      m1 = create shutdown msg # (None because already created)
                                      m2 = close message
                                      unlock()
                                      send m1, m2  # (None, close)
send m # (EOF)
```

By putting an extra lock around the `close`, `shutdown`, `_request_failed`, and `_handle_close` I don't see the issue anymore. However, I think it is still possible to have race conditions between send and close (create a message to send, then close is interleaved, then send the send message).

However, it seems like there was a [good reason](https://github.com/paramiko/paramiko/commit/305f5e09a58d4c80740faf564dfdca875d98245f) that locks like that didn't exist in the first place what ideas do you have for fixing the race condition?

DavidQing:
Hi, I have been using paramiko-ng but it currently don't support Jammy Jellyfish due to rsa-sha2-256 and rsa-sha2-512 change. I want to switch back to paramiko but this bug is still causing failures.
Is there a way to pull in the changes in https://github.com/ploxiln/paramiko-ng/pull/101 to paramiko? 
Thanks

bskinn:
Looks like #1137 was the original inspiration for the work in ploxiln#101.

@DavidQing, bitprophet is in the middle of modernizing the key/auth systems (see #387 #23), which will be occupying much of his `paramiko` time for a while, spanning the next couple of minor releases.

This seems like something of great enough a magnitude that it will have to wait until after the key/auth work is done. So, it will likely be some months before this can be addressed. The fact that #1137 is available will hopefully speed this one up.

Keep an eye on this issue for activity; and, if you have any additional information about the details of the race condition, please add them to the thread here.

sdodsley:
@bskinn any news on when this will be resolved?

nmorad:
@bskinn Any updates on this bug?

AlonMarko:
@bskinn 
is there an update on this bug atm?

bskinn:
[...]

PR #1137, opened 2017-12-18
Fix race condition between Channel.close() and other channel send met…

yieldi:
…hods.

This is a fix for #1115 .


bskinn:
Labeling as `needs tests` -- tests may or may not be practical, but they should be considered.

PR #1140, opened 2017-12-22
Support nested ProxyCommand with escape

### LINKED ISSUE #1139: Nested ProxyCommand
In my office, I should use a nested `ProxyCommand` to connect my hosts.

```
[ my-computer ] -> [ outer-bastion ] -> [ inner-bastion ] -> [ target-host ]
```

We can config the above nested `ProxyCommand` in the SSH config by a config for `inner-bastion`:

```
Host target-host
  ProxyCommand ssh -W %h:%p inner-bastion

Host inner-bastion
  ProxyCommand ssh -W %h:%p outer-bastion

Host outer-bastion
  ...
```

There's one more special condition.  `my-computer` can resolve the domain of `target-host`, but `outer-bastion` cannot.  For the reason, I should use `$(dig +short %h)` instead of `%h` to let `outer-bastion` know the IP address of `target-host`.

So the config for `target-host` looks like:

```
Host target-host
  ProxyCommand ssh -W "$(dig +short %h)":%p inner-bastion
```

`ssh target-host` command works well with the config.  But `paramiko.SSHConfig(open('.ssh/config')).lookup(host)` does not.

There're 2 problems:

1. `paramikoSSHConfig.lookup()` reads only 1 config for the given host.
2. `paramiko.ProxyCommand()` does not expand `$(dig +short %h)`.

To fix the first problem, I can inline the config for `inner-bastion` into the config for `target-host`:

```
Host target-host
  ProxyCommand ssh -W "$(dig +short %h)":%p inner-bastion -o ProxyCommand='ssh -W %%h:%%p outer-bastion'
```

Now I should pass `%%h` in the nested `ProxyCommand` to make sure `outer-bastion` connects to `inner-bastion`.  If I pass `%h`, `outer-bastion` will try to connect to `target-host` instead.

It also works well with `ssh target-host` command.  But Paramiko doesn't.  Because `paramiko.SSHConfig._expand_variables()` doesn't understand its escape rule.  It replaces `%%h` with `%` + IP address of `target-host`.

So now, we have 2 problems to fix:

1. ~~`paramikoSSHConfig.lookup()` reads only 1 config for the given host.~~
2. `paramiko.ProxyCommand()` does not expand `$(dig +short %h)`.
3. `paramiko.SSHConfig._expand_variables()` expands `%%h` as `%` + `%h`.

I know, my network topology is very weird.  But I cannot change it against my employers.  And I think a config which works well in `ssh` command, it should work well in Paramiko also.  So I made a fix.

sublee:
Fix #1139.

### DIFF
--- a/paramiko/config.py
+++ b/paramiko/config.py
@@ -212,9 +212,13 @@ class SSHConfig (object):
                             ('~', homedir),
                             ('%h', config['hostname']),
                             ('%p', port),
-                            ('%r', remoteuser)
+                            ('%r', remoteuser),
+                            ('%%', '%')
                         ]
                         }
+
+        def _replace(value, find, replace):
+            return re.sub(r'(?<!%)' + find, str(replace), value)
 
         for k in config:
             if config[k] is None:
@@ -224,12 +228,11 @@ class SSHConfig (object):
                     if isinstance(config[k], list):
                         for item in range(len(config[k])):
                             if find in config[k][item]:
-                                config[k][item] = config[k][item].replace(
-                                    find, str(replace)
-                                )
+                                config[k][item] = _replace(config[k][item],
+                                                           find, replace)
                     else:
                         if find in config[k]:
-                            config[k] = config[k].replace(find, str(replace))
+                            config[k] = _replace(config[k], find, replace)
         return config
 
     def _get_hosts(self, host):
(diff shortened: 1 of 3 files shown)


sublee:
Perhaps you don't like `shell=True` for a security reason.  But you may need the escape of `%` character.

PR #1144, opened 2018-01-07
Add auth interactive handler to sshclient

jciskey:
This update adds a kwarg to the SSHClient.connect() method that allows users to provide a handler that gets passed to auth_interactive instead of wholly relying on auth_interactive_dumb (which just does the same thing, but with a default handler instead).

Originating Use Case:
* Using Ansible to connect to a server that has the Google 2FA PAM module active.

Reason why:
* Even though the documentation states that the password kwarg should work for providing the 2FA code, it doesn't in this case.
* Dropping down to auth_interactive_dumb() to push parsing the server responses onto the user doesn't work when a system like Ansible is using SSHClient non-interactively.
* Adding this hook gives users the option to provide their own code for processing the keyboard-interactive server prompts (such as recognizing the "Verification code:" prompt and generating a TOTP automatically).

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -231,6 +231,7 @@ class SSHClient (ClosingContextManager):
         auth_timeout=None,
         gss_trust_dns=True,
         passphrase=None,
+        auth_interactive_handler=None,
     ):
         """
         Connect to an SSH server and authenticate to it.  The server's host key
@@ -261,6 +262,7 @@ class SSHClient (ClosingContextManager):
                 the private key and used for authentication.
 
             - Plain username/password auth, if a password was given
+            - Keyboard-Interactive authentication (such as 2FA)
 
         If a private key requires a password to unlock it, and a password is
         passed in, that password will be used to attempt to unlock the key.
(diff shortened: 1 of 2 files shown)


bitprophet:
This may (but may not) conflict with ongoing work re #387, leaving a note in hopes I come back to this once that's done & see how it can be slotted in (or if it's already solved.)

If you've time, detailed instructions on how to reproduce this specific 2FA case (I personally haven't used Google's 2FA, only Duo's, and have not used Ansible much) would help me prove/disprove whether it's been fixed at that point - thanks!

jciskey:
This all actually arose from me trying to build an Ansible patch for connecting to 2FA-enabled servers, since that's currently a no-go.

I've got a Vagrantfile with a provisioning playbook that will stand up a VM and enable 2FA + public key authentication, but to test the Ansible portion you'd need to pull the forked version of Ansible I'm working on. That's why I tried to add some unit-tests to verify that the functionality will work and doesn't break the rest of the authentication process.

I will note that Fabric authenticates to the same server without needing this Paramiko patch, so it's also possible that there's just something really funky going on with how Ansible is using SSHClient that makes the password-passing functionality not work for 2FA.

I just figured that even independently of Ansible, having the option to integrate with the auth_interactive Transport method would be useful for people trying to interact with other PAM modules that use keyboard-interactive.

How would you prefer I handle documenting the repro steps for this? A completely new git repo? A wiki page? Something else?

PR #1168, opened 2018-02-21
Added Ed25519Key to __all__

patriksevallius:
PyCharm complains when I try to do:

from paramiko import Ed25519Key

It says Ed25519Key is missing from __all__.

This patch adds it to __all__ and then PyCharm stops complaining.

### DIFF
--- a/paramiko/__init__.py
+++ b/paramiko/__init__.py
@@ -84,6 +84,7 @@ __all__ = [
     'PKey',
     'RSAKey',
     'DSSKey',
+    'Ed25519Key',
     'Message',
     'SSHException',
     'AuthenticationException',


ploxiln:
Yup, looks like it should be added there. Also, `ECDSAKey` seems to be in the same boat.

patriksevallius:
Do you want me to add ECDSAKey to my branch and create a new PR?

bitprophet:
Gah I hate `__all__` sometimes. Suspect it gets overlooked because using `from module import *` is an antipattern anyways...

I'll add ECSDAKey when I merge, sec.

bitprophet:
Yea pycharm must be specifically introspecting `__all__`, we were importing those names prior to this PR. Bleh.

bitprophet:
Thanks both!

PR #1178, opened 2018-03-21
Update twine requirement in dev-requirements.txt

brainwane:
I'm one of the maintainers of [Twine](https://twine.readthedocs.io/) and I thank you for using it! You're using a version that's about ten months old, and I'd appreciate if you'd upgrade to a more recent version ([changelog](https://twine.readthedocs.io/en/latest/changelog.html)), so here's a PR for your convenience. Best wishes.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,7 +5,7 @@ alabaster>=0.7.5,<2.0
 releases>=1.1.0,<2.0
 semantic_version<3.0
 wheel==0.24
-twine==1.9.1
+twine>=1.11.0
 flake8==2.6.2
 pytest==3.2.1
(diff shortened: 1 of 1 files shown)


bitprophet:
If only all my deps got updated (upstream or just by myself) more often than ten months 😂  

Yea I haven't had to bump it since, IIRC, PyPI made a breaking change (SSL related?) requiring an update. Heck I'm not even sure all my projects are at 1.9 yet...!

You don't explicitly say but I'm guessing one reason you're posting this around is because of the pypi.org changes? Either way I'll just-merge and hopefully it'll be smooth next time I release...thanks!

bitprophet:
Backported to 1.17+.

PR #1180, opened 2018-03-25
Fix infinite loop when proxy command dies

vincentbernat:
When proxy command dies, reading from the file descriptor returns an
empty string. From documentation, this is the expected way to signal
an end-of-file for this file descriptor (no exception is raised). If
we have something in the buffer, let's give it back, otherwise,
trigger an exception.


ploxiln:
see also #963

ploxiln:
nevermind, I see you've seen that PR :) this does look better

nbartos:
@vincentbernat This does look better than my old PR https://github.com/paramiko/paramiko/pull/963, I just closed it.  It looks like you may need to write a test to get it merged, the code coverage checks are complaining.

nbartos:
@vincentbernat Would you mind resolving the merge conflicts and coverage test failures?  It would be really great to get this merged in.

vincentbernat:
I have rebased on master.

paramiko/proxy.py
```diff
@@ -96,9 +96,14 @@ def recv(self, size):
...
                 if r and r[0] == self.process.stdout:
-                    buffer += os.read(
-                        self.process.stdout.fileno(), size - len(buffer)
-                    )
+                    b = os.read(
+                        self.process.stdout.fileno(), size - len(buffer))
+                    if not b:
+                        if buffer:
+                            return buffer
+                        raise ProxyCommandFailure(' '.join(self.cmd),
```
nbartos: When trying out this code, it looks like this exception is always getting raised (paramiko==2.4.1) after a command successfully exits.  I think this should be:
```
                if r and r[0] == self.process.stdout:
                    b = os.read(self.process.stdout.fileno(), size - len(buffer))
                    if not b:
                        return buffer
                    buffer += b
            return buffer
```

vincentbernat:
I have updated the PR with your suggestion.

nbartos:
Looks like the python 3.6 gate wants a line length change:
```                     b = os.read(
-                        self.process.stdout.fileno(), size - len(buffer))
+                        self.process.stdout.fileno(), size - len(buffer)
+                    )
```

vincentbernat:
OK, this should be fixed.

nbartos:
@ploxiln I don't suppose you know who we should ping to try and get this merged?

ploxiln:
@bitprophet

bitprophet:
It'd be sweet if we could whip up some tests for this, but I'll probably only be a stickler about that if there are existing tests for this functionality that can be easily extended. I've put it in the next bugfix milestone. No ETA.

Issue #1182, opened 2018-03-27
TypeError Excepiton during connect (probably incorrect usage of instance method)


a1ds:
Hello.

The problem looks like a bug, but I don’t sure.

**My environment:**

```
###
uname -a
Linux motkb-dev-app 3.10.0-514.el7.x86_64 #1 SMP Wed Oct 19 11:24:13 EDT 2016 x86_64 x86_64 x86_64 GNU/Linux

###
cat /etc/redhat-release
Red Hat Enterprise Linux Server release 7.3 (Maipo)

###
yum list installed | grep python36
rh-python36-python.x86_64             3.6.3-1.el7               @asm_dev_local
rh-python36-python-devel.x86_64       3.6.3-1.el7               @asm_dev_local
rh-python36-python-libs.x86_64        3.6.3-1.el7               @asm_dev_local
rh-python36-python-pip.noarch         9.0.1-2.el7               @asm_dev_local
rh-python36-python-setuptools.noarch  36.5.0-1.el7              @asm_dev_local
rh-python36-python-tkinter.x86_64     3.6.3-1.el7               @asm_dev_local
rh-python36-python-tools.x86_64       3.6.3-1.el7               @asm_dev_local
rh-python36-python-virtualenv.noarch  15.1.0-2.el7              @asm_dev_local
rh-python36-runtime.x86_64            2.0-1.el7                 @asm_dev_local

###
python --version
Python 3.6.3

###
Python 3.6.3 (default, Oct 25 2017, 13:35:50)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-16)] on linux

###
pip list
Package        Version
-------------- -------
asn1crypto     0.24.0
bcrypt         3.1.4
cffi           1.11.5
cryptography   2.2
docopt         0.6.2
idna           2.6
jedi           0.11.1
paramiko       2.4.1
parso          0.1.1
pip            9.0.1
prompt-toolkit 1.0.15
ptpython       0.41
pyasn1         0.4.2
pycparser      2.18
Pygments       2.2.0
PyNaCl         1.2.1
setuptools     28.8.0
six            1.11.0
wcwidth        0.1.7
wheel          0.29.0
```

**Steps to reproduce the problem:**

_steps_to_reproduce_the_problem.py_
```python
import paramiko

host = '127.0.0.1'

client = paramiko.SSHClient()
client.set_missing_host_key_policy(paramiko.AutoAddPolicy())

client.connect(hostname=host)


[...]

PR #1184, opened 2018-03-28
Add safe binding of sha* functions

a1ds:
Depending of openssl, sha* functions from hashlib module can be
represented by different objects (to be precise, with decorated or
not decorated function from _hashlib module).
    In our case, this functions added to classes as class attributes.
It is not safe to call them using self without wrapping with
staticmethod.

Example:
    When type(sha256) is "<class 'builtin_function_or_method'>"
original code will work without an issues, but when type(sha256) is
"<class 'function'>" the code will fail with an exception.

resolves paramiko/paramiko#1182


busyworker:
When will this pull request be merged?

bitprophet:
@busyworker popped some questions over in the main bug thread, once those are resolved I'll try to get this into the appropriate bugfix lines, thanks!

a1ds:
I added test and changes for kex* modules.
Asking for a code review)

a1ds:
Added tests and changes for kex* modules.

ploxiln:
I gotta say, the tests are pretty crazy, they don't seem worth it - I'd just test that you can call the `.hash_algo` attr of the classes/objects as expected, in all the versions of python the tests are already run with.

a1ds:
I agree, it's kinda overkill. :laughing: 
The idea here is the following: it's easy to forget about `staticmethod` stuff. If there will be a new Kex* class, the test will grab this new class.

PR #1189, opened 2018-04-11
use poll() instead of select() if possible

multi-io:
This PR converts all usages of `select()` (I identified 3) to `poll()`, if the platform supports it (the check is based on https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/default.py). `select()` passes the set of file descriptors to monitor as a fixed-length bitmask and thus stops working if the process has more than 1024 open files and you try to monitor file descriptors >1023.

There is a new `paramiko.util.wait_until_readable` API that encapsulates the functionality (it can only check for readable descriptors, as that's all that's needed by Paramiko, it seems). The function receives and returns numerical file descriptors rather than file objects because that's all `select.poll()` can handle.

This isn't just an academic change; we have a real-world legacy application here that opens many long-running proxied connections in parallel and runs into this pretty reliably.


kaimast:
Are there any chances of this being merged soon? I too keep running into problems with paramiko when my application has many files (i.e. > 1024) open.

PR #1194, opened 2018-04-22
Support include directives in ssh_config files

tobald:
This is a followup of #872. I have squashed original author commits and added a fix for supporting relative file path in case of system ssh_config files.

### DIFF
--- a/paramiko/config.py
+++ b/paramiko/config.py
@@ -84,9 +84,11 @@ class SSHConfig (object):
             elif key == 'include':
                 # support for Include directive in ssh_config
                 path = value
-                # according to SSH documentation if the path is relative, we look forward in ~/.ssh
+                # the path can be relative to its parent configuration file
                 if os.path.isabs(path) == False and path[0] != '~':
-                    path = '~/.ssh/'+path
+                    folder = os.path.dirname(file_obj.name)
+                    path = os.path.join(folder, path)
+
                 #expand the user home path
                 path = os.path.expanduser(path)
                 if parsed_files == None:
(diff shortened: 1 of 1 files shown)


paramiko/config.py
```diff
@@ -49,7 +50,7 @@ def __init__(self):
         """
         self._config = []
 
-    def parse(self, file_obj):
+    def parse(self, file_obj, parsed_files=[]):
```
ploxiln: There's a potential gotcha here - if `parse()` is called on two separate occasions (non-recursively), the default value the second time will retain the modifications (appends) from the first time `parse()` was called.

Generally, whenever you want a default value for a param, and you mutate that param, you want it like so:

```python
def func(param=None):
    if param is None:
        param = []
```

tobald:
Thanks for the review, I pushed a fix for the issue you mentioned.

ploxiln:
Looks good to me.

(With the caveat that this won't work for `Include` inside a `Host` block - but that can be a feature for another day.)

GhostLyrics:
Could you revisit this and let us know what is still missing?

viable-hartman:
I too would like to see this revisited and or merged.  I'd add a "strip('\"') function to the value part though to handle when paths like this "~/.ssh/group/*," which are valid in ssh config.

czubik8805:
How about a merge?

danihodovic:
How about a merge? :pray:

oresk:
How about a merge?

bskinn:
I can't speak for @bitprophet, but this PR would be considerably strengthened by the addition of tests and documentation.

PR #1202, opened 2018-05-01
SSHClient.open_sftp: Add timeout arg; pass to open_channel

evanj:
The default channel timeout is 1 hour. When connecting to slow SFTP
servers, they can take a long time to respond, which results in a
hung connection. Add an optional timeout argument to
SSHClient.open_sftp which gets passed into Transport.open_channel
so callers can get an early SSHException.


tests/test_sftp.py
```diff
@@ -769,3 +770,36 @@ def test_write_memoryview(self, sftp):
...
+        return paramiko.OPEN_SUCCEEDED
+
+@slow
+class TestSFTPTimeout(object):
+    def test_auth_session_timeout(self, sftp_server):
+        """
+        verify that the open_sftp has a configurable timeout
+        """
+
+        # copied from conftest.sftp_server
```
bitprophet: Was there a reason this was copied instead of just used directly?

evanj: That function connects the client socket directly to a `Transport` and returns it. I needed just the socket, so I can pass it into `SSHClient` to test `SSHClient.open_sftp` and all the other functions it calls.

Now that I think about it: I could split that function into two, a lower level one that returns the client socket, which is used by `conftest.sftp_server` and by this test. Would that make sense?

evanj:
I've attempted to refactor the test to eliminate the test duplication. Hopefully this makes more sense, thanks for the helpful feedback!

tests/test_sftp.py
bitprophet: I think so, yea. Give it a shot!

ManniCS:
@bitprophet any thoughts on how this Pull Request could be improved in order for it to be merged to master?

PR #1207, opened 2018-05-10
ProxyCommand: run the command verbatim, via the shell

jordiecometrica:
I have fairly complicated ssh setup for AWS in my ~/.ssh/config. It
looks something like this:

    Host bastion
    ProxyCommand nc $(aws ec2 describe-instances --filters "[{\"Name\":\"tag:aws:cloudformation:stack-name\", \"Values\":[\"Bastion\"]}, {\"Name\": \"tag:role\", \"Values\": [\"bastion\"]}]" --query="Reservations[*].Instances[0].PublicIpAddress" | grep -oP '(\d+\.){3}\d+' | head -1) 22

    Host prod
    ProxyCommand ssh -W $(aws ec2 describe-instances --filters "[{\"Name\":\"tag:aws:cloudformation:stack-name\", \"Values\":[\"PROD\"]}, {\"Name\": \"tag:role\", \"Values\": [\"web\"]}]" --query="Reservations[*].Instances[0].PrivateIpAddress" | grep -oP '(\d+\.){3}\d+' | head -1):22 bastion

In this way, I can do `ssh prod` and hop via a bastion server into the
prod server.

The point is that there are two complicated commands in there with
some mildly sophisticated shell syntax that really needs to be
interpreted by the shell.

OpenSSH handles this okay. It runs the actual shell. That's what I
want it to do. However, paramiko does not and my proxy commands don't
work.

I don't think there is a security risk here of shell injection. These
proxy commands are shell commands that the user put into their
.ssh/config file and if they're good enough for OpenSSH to run
verbatim via a shell, then paramiko can do that too.


jordiecometrica:
#1139 is related to this and has a workaround in order to avoid the shell parsing.
#773 would have also been superseded by this fix.

jordiecometrica:
Hey @bitprophet would you like me to reopen this against `main`?

bitprophet:
Huh, I thought Github's explicit support for renaming wouldn't close stuff 😠  that said - I actually have something in the pipe that changes this functionality to use Invoke's run() (generally more powerful and easier to manage than raw subprocess) and I think this would probably get squashed by that when it lands :(

Likely out in one of the next major (X.0) releases but I might land it even sooner with an opt-in feature 🤔

PR #1209, opened 2018-05-14
Test and dev dependencies cleanup and sync with Fabric

offbyone:
* Update invocations min version
* Update pytest to be consistent with fabric
* Add pytest-xdist to enable `guard` task
* Other dependencies as found to be inconsistent with Fabric2
* add `inv guard` task

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -15,13 +15,14 @@ def test(
     ctx,
     verbose=True,
     color=True,
-    capture='sys',
+    capture="sys",
     module=None,
     k=None,
     x=False,
     opts="",
     coverage=False,
     include_slow=False,
+    loop_on_fail=False,
 ):
     """
     Run unit tests via pytest.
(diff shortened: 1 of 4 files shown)


tasks.py
```diff
@@ -31,19 +32,21 @@ def test(
     well with explicit ``--opts="-m=xxx"`` - if ``-m`` is found in ``--opts``,
     ``--include-slow`` will be ignored!)
     """
-    if verbose and '--verbose' not in opts and '-v' not in opts:
+    if verbose and "--verbose" not in opts and "-v" not in opts:
```
ploxiln: You should probably avoid unrelated style changes in pull requests. If general style changes are desired they should probably be a separate independent pull request.

offbyone: True that, but the line length with the new task tripped flake8, and instead of mucking around formatting it correctly I elected to just blacken the file and have done. I *can* remove this if you want, but I'm not convinced it's worth the extra CPU cycles on Travis 😄

bitprophet:
@ploxiln I normally agree but in this case we're slowly starting to run `black` on my various repos so I don't mind too hard, esp since it currently only affects the tasks module and this entire PR is largely administrivia and not a discrete feature `¯\_(ツ)_/¯`

bitprophet:
Still WTB a magical unicorn rainbow puppy feature where github changes the base _and also fucking rebases_ 😠 

I will just cherry pick this back to 2.4...sec

bitprophet:
OH wait that's in the merge button drop down now, isn't it. Hmmm

bitprophet:
Cherry-picked to 2.4 and merged that to master, should be set now.

PR #1212, opened 2018-05-16
Add type converting config

offbyone:
In order to implement fabric/fabric#1753 I'm adding a configuration helper into paramiko to make it more reasonable to get config values in mapped types (yes -> True &c)


paramiko/config.py
```diff
@@ -295,3 +295,29 @@ def __str__(self):
...
+
+class SSHConfigDict(dict):
+    """A dictionary wrapper for ssh host configurations.
+
+    This class introduces some usage niceties for consumers of SSHConfig,
+    specifically around the issue of variable type conversions. This offers
+    as_bool(key) and as_int(key) for the current raw string values in
+    SSHConfig"""
+
+    def __init__(self, *args, **kwargs):
```
bitprophet: Do we still need this now that we're subclassing just dict and not UserDict?

bitprophet:
- When you have a sec, please try rebasing or merging onto latest master, which should get rid of all the incidentals in dev-reqs.txt etc.
- Left a line note re: dunder-init, see above
- Can you cave to my unreasonable formatting demands and make your docstrings of this format?
  ```
  """
  hello yes this is text
  """
  ```

paramiko/config.py
offbyone: Good question. Probably not, but it doesn't hurt.

PR #1232, opened 2018-06-11
Updated dependencies for ed25519 key support

khomutoff:
After upgrade to paramiko version 2.4.1 one of my apps has stopped working and throw following exception:

> BadHostKeyException: (u'xx.xx.xx.xx', <paramiko.ed25519key.Ed25519Key object at 0x7f9d0dbbb150>, <paramiko.ed25519key.Ed25519Key object at 0x7f9d0dbf4f10>)

This issue occurs with these versions of dependencies:
cryptography==2.1.1
pynacl==1.0.1

The problem has been fixed by upgrade to the next versions:
cryptography==2.2.2
pynacl==1.2.1

To enable support of ed25519 type keys (#325) I've updated the minimum required versions specified in installation prerequisites.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -71,5 +71,5 @@ setup(
         "Programming Language :: Python :: 3.5",
         "Programming Language :: Python :: 3.6",
     ],
-    install_requires=["bcrypt>=3.1.3", "cryptography>=1.5", "pynacl>=1.0.1"],
+    install_requires=["bcrypt>=3.1.3", "cryptography>=2.2.2", "pynacl>=1.2.1"],
 )


ksnavely:
+1, would love to have this

ksnavely:
@bitprophet @khomutoff This would be a great change to have in, what do you think?

bitprophet:
Not sure about this; it's not like users cannot manually upgrade if they hit this issue, and there's always a chance some folks are stuck on older versions of these dependencies (whether or not they are even using Ed25519 keys) and would then find the package uninstallable.

My gut says an FAQ entry would be the best immediate "fix" for this and we'd then tweak dependencies later on (eg at 3.0 or some other "psst hey dependencies are changing slightly" minor version).

Also, has anybody identified what exactly changed in Cryptography and/or PyNaCl to fix the underlying issue? Just curious there, it's always nice to link to specific bugs or otherwise explain in detail why we might change our own dependencies.

adiroiban:
Just a comment. Latest version of Cryptography support ED25519 .
So maybe when working on this, the pynacl dependency can be removed and only use OpenSSL via Cryptography.

bitprophet:
Rolling this into either #1806 or #2103 depending on which eventually gets adopted.

bitprophet:
Also, we are bumping our overall dep to Crypto 3.3 as part of our 3.0 release!

PR #1234, opened 2018-06-12
RFC8268 added kex dh modp groups support

hirura:
This is to add support for RFC8268 KEX DH groups 14, 15, 16, 17, 18 with SHA2 hash algorithm.


hirura:
Wow, edgsousa opened a similar request 10 min ago...

PR #1239, opened 2018-06-15
Query publickey prior to sending signature

f1shbot:
Public key authentication requires a query prior to transmitting a
signature (rfc4252 section 7).  A query is now sent prior to sending a
signature.

### DIFF
--- a/paramiko/auth_handler.py
+++ b/paramiko/auth_handler.py
@@ -39,6 +39,7 @@ from paramiko.common import (
     cMSG_USERAUTH_INFO_REQUEST,
     WARNING,
     AUTH_FAILED,
+    MSG_USERAUTH_PK_OK,
     cMSG_USERAUTH_PK_OK,
     cMSG_USERAUTH_INFO_RESPONSE,
     MSG_SERVICE_REQUEST,
@@ -283,7 +284,7 @@ class AuthHandler(object):
                 password = b(self.password)
                 m.add_string(password)
             elif self.auth_method == "publickey":
-                m.add_boolean(True)
+                m.add_boolean(False)
                 # Use certificate contents, if available, plain pubkey
                 # otherwise
                 if self.private_key.public_blob:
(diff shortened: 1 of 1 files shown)


radssh:
Is there a server implementation that actually requires this? The two primary reasons to defer sending the signature would be to 1) only prompt for a passphrase if the key would be accepted; and 2) avoid the cycles computing the signature for a key that would not be accepted. And according to the RFC: `The client MAY send the signature directly without first verifying whether the key is acceptable.`.
There is some potential value here, if the private key decryption could be deferred until that preliminary check succeeds, but Paramiko currently includes the decryption at load time. Might want to include this PR in the #387 bucket.

shawndfisher:
> Is there a server implementation that actually requires this? 

My poorly implemented libSSH server did, but that has been fixed to handle both query/signature AND plain signature.

My confusion occured with "send_pubkey_test" being sent from OpenSSH's log output:

> $ ssh -vvv admin@192.168.0.1
> . . .
> debug2: service_accept: ssh-userauth
> debug1: SSH2_MSG_SERVICE_ACCEPT received
> debug3: send packet: type 50
> debug3: receive packet: type 51
> debug1: Authentications that can continue: publickey,password
> debug3: start over, passed a different list publickey,password
> debug3: preferred publickey,keyboard-interactive,password
> debug3: authmethod_lookup publickey
> debug3: remaining preferred: keyboard-interactive,password
> debug3: authmethod_is_enabled publickey
> debug1: Next authentication method: publickey
> debug1: Offering public key: RSA SHA256:XXX <key_path>
> debug3: send_pubkey_test
> debug3: send packet: type 50
> debug2: we sent a publickey packet, wait for reply
> debug3: receive packet: type 60
> debug1: Server accepts key: pkalg ssh-rsa blen 279
> debug2: input_userauth_pk_ok: fp SHA256:XXX
> debug3: sign_and_send_pubkey: RSA SHA256:XXX
> debug3: send packet: type 50
> debug3: receive packet: type 52
> debug1: Authentication succeeded (publickey).
> . . .

From the output here, you can see the query (send_pubkey_test) before the actually signature is sent (send_packet: type 50).  I misunderstood the probe as being a rule of law for SSH servers, but as you pointed out, servers should handle a signature as well as the query/signature.

PR #1243, opened 2018-06-20
Don't even look for keys when look_for_keys is False

teran-mckinney:
This streamlines the code, improves readability, and fixes unnecessary failures where getpw functions may fail (systemd DynamicUsers and similar concepts).

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -700,25 +700,24 @@ class SSHClient(ClosingContextManager):
         if not two_factor:
             keyfiles = []
 
-            for keytype, name in [
-                (RSAKey, "rsa"),
-                (DSSKey, "dsa"),
-                (ECDSAKey, "ecdsa"),
-                (Ed25519Key, "ed25519"),
-            ]:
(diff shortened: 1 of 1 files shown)


bitprophet:
This may clash with the (...still impending!) rewrite of all the auth stuff in #387 but I'll try and remember to scan it when that's complete to make sure we account for the behavior you're trying to express here. Thanks!

teran-mckinney:
Great, thank you! Do you think that will come out fairly soon? I could use this fix as I do a lot with DynamicUsers.

bitprophet:
I've got a lot on my plate but that auth overhaul is my number 1 priority for Paramiko specifically. Once I give another higher-level and equally high-impact project (Fabric 2) a bit of post-release love, my plan is to finish that partly done work and get it out as Paramiko 2.5 or 3.0 (depending on how backwards compat I can make it.)

Please follow #387 for progress updates!

teran-mckinney:
I understand. I appreciate it!

PR #1246, opened 2018-06-26
Raise PKCS8NotSupportedException for PKCS#8 keys

jchampio:
We have a need to bubble up a warning to end users if they happen to be using a PKCS#8 private key (e.g. because they're on a FIPS system). This patch allows Paramiko clients to differentiate between generic `SSHException`s and those raised because a PKCS#8 key is in use.

The new unit-test keys are PKCS#8-format copies of the existing `test_rsa*.key` files.

This was developed against the 1.18 patch line (which we're currently using with Python 2); I expect you'll want me to cherry-pick this to other branches, but I wasn't really sure which ones.


jchampio:
> All checks have failed

Hmm, Python 2.6 isn't a fan of `with self.assertRaises()`. Any preferences on how that test *should* be written?

ploxiln:
something like:

```python
        self.assertRaises(PKCS8NotSupportedException,
            RSAKey.from_private_key_file, test_path('test_rsa.pkcs8.key')
        )
```

bitprophet:
@jchampio I super appreciate you taking the time to think about which branch this ought to go against! Unfortunately I'm not sure 1.18 is appropriate, this feels more like a feature-level change to me (a hint at this is the need to use `versionadded`/`versionchanged`).

Complicating things further is that I'm not doing feature support on the 1.x lines anymore (they rely on now-insecure PyCrypto; plus managing dual feature-release families is more of a chore than straight bugfixes). Is there a reason you're on 1.18 instead of e.g. a lower 2.x (if you're concerned about stability)? The only thing that really changed in Paramiko 2.0 was swapping PyCrypto out for Cryptography, the public APIs are still the same.

----

In terms of the actual feature under discussion, I think I'm +1 on it, papering over the "SSHException everywhere!" design I inherited when I took over has been a very long standing goal.

jchampio:
@ploxiln I'll give that a try, thanks!

> [...] I'm not doing feature support on the 1.x lines anymore

@bitprophet Completely understood. We're on the 1.x line at the moment for much the same reason -- no major bumps to pieces of the software stack, if we can avoid them. Knowing that the APIs remained constant into 2.x should help push us towards our next upgrade, though.

> So, all else equal, I feel like this could actually go against master and come out in Paramiko 2.5.

Sounds good. I will reroll this PR against master when I get a chance (probably tomorrow morning). Thanks!

jchampio:
@bitprophet Re-rolled. I'm not sure how to bump Travis after the base branch change...

jchampio:
@bitprophet Anything else I can help with on this PR?

bitprophet:
@jchampio Nope, I just need the time to jump back over to Paramiko from my other related repos. Just made sure this went in the 2.5 milestone. Thanks!

bskinn:
Relevant to the key/auth work of #387, though somewhat outside it since this is about declaring a lack of support via exception, as opposed to implementing new machinery to handle supported auth.

PR #1248, opened 2018-06-29
If a timeout is defined, use it in Channel._wait_for_event

alexbrett:
At present after we have sent a command (e.g. a channel request), we wait on a
threading.Event object for a signal.
This signal will be delivered either once we get a response to the request, or
if there is a socket error, such as a timeout.

A badly behaved server can fail to respond to the request, but not cause a
socket error, resulting in a potentially infinite wait.

As such, if we have a timeout defined, use it in _wait_for_event and throw a
socket timeout exception if we hit it.

### DIFF
--- a/paramiko/channel.py
+++ b/paramiko/channel.py
@@ -1198,7 +1198,8 @@ class Channel(ClosingContextManager):
         self.event_ready = False
 
     def _wait_for_event(self):
-        self.event.wait()
+        if not self.event.wait(timeout=self.timeout):
+            raise socket.timeout()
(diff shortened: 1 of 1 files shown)


gotostack:
Cloud we release a new version with this patch? We met some timeout may related to this.

jrobins2:
Can this update be included.

bskinn:
Flagging for evaluation by @bitprophet -- it's unclear to me whether it's a good idea to reuse `self.timeout` in this way, coupling this timeout to the other timeout(s) associated with the value.

PR #1259, opened 2018-08-09
Fixes #1255: Add os.path-like stat-based methods to SFTPClient

### LINKED ISSUE #1255: Emulate exists, isdir, isfile from os.path in SFTP connection
It would be very useful if the `SFTPClient` class supported similar behavior to `exists`, `isfile`, and `isdir` from `os.path`.

The current recommendation appears to be using `stat` to identify resources:

https://stackoverflow.com/questions/18205731/how-to-check-a-remote-path-is-a-file-or-a-directory

GPHemsley:
Add methods to `SFTPClient` that resemble equivalent methods from
`os.path`:
 * `exists`
 * `lexists`
 * `isfile`
 * `islink`
 * `isdir`

All of these methods are based on the results of calling `stat` on a
path, except `lexists`, which is based on `lstat`.

These methods reduce the need for a user of the paramiko library
to import the Python `stat` module and inspect the return value of
an `SFTPClient.stat` call.


bitprophet:
Thanks, popped it into next milestone so I don't forget it.

Also a note to myself: this _may_ impact Patchwork (Fabric 2's contrib library) which reimplements some of this stuff via shell commands. However, since not all remote targets will necessarily have SFTP enabled, that's probably no bueno; but we could maybe implement it in a tiered fallback manner (try SFTP, if SFTP seems disabled, use `test` or w/e).

uscivics:
is this going to be merged?  Should I fork this in the interim?

JohnPreston:
This is old, but still. Bump

wbarnha:
I'd like these features added so I can move away from https://bitbucket.org/dundeemt/pysftp. Bumping for attention since this feature is perfectly functional.

wbarnha:
Seems like this PR has been forgotten, so I created a fork in https://github.com/wbarnha/paramiko-stat that acts like a wrapper.

bskinn:
Not forgotten, @wbarnha, just falling below some other, major work in the priority queue for bitprophet's limited open source time. First was dropping Python 2 and old Python 3.x; now the key/auth overhaul of #387, #23, and related.

bitprophet has this sorted into a prioritization milestone -- once these big tasks are out of the way, he'll be moving on to features & fixes like this.

Sorry for the wait; we'll get there, though.

PR #1262, opened 2018-08-18
Add *.pub files to manifest

kapsh:


kapsh:
Currently tarball on PyPI does not include *.pub files under tests/cert_support. Because of this tests fail if run against it (they can't find test_rsa.key-cert.pub). 

[tests.log](https://github.com/paramiko/paramiko/files/2299788/tests.log)

bitprophet:
Ooh nice catch! Thank you.

bitprophet:
Cherry picked back to 2.3 (which is when those .pub files were added, apparently) and merged up. Will prob have to manually close this in a minute.

PR #1266, opened 2018-08-22
fix exceptions to catch

### LINKED ISSUE #1257: Exception list doesn't match the current API (ECDSAKey._decode_key(self, data))
https://github.com/paramiko/paramiko/blob/6f0e7c58d0296d157d33194a30de1789da167480/paramiko/ecdsakey.py#L290

Compare to:

https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/?highlight=load_der_private_key#cryptography.hazmat.primitives.serialization.load_der_private_key

> Raises:
>
> - `ValueError` -  If the DER data could not be decrypted or if its structure could not be decoded successfully.
> - `TypeError` - If a password was given and the private key was not encrypted. Or if the key was encrypted but no password was supplied.
> - `cryptography.exceptions.UnsupportedAlgorithm` - If the serialized key is of a type that is not supported by the backend or if the key is encrypted with a symmetric cipher that is not supported by the backend.

This appears to be a mismatch. `TypeError` is a sibling to `AssertionError` and thus will not be caught. The library-specific exception is not being caught at all.

Proposed fix: update the exception list to match that of the cryptography.io API.

tylergarcianet:
closes #1257

### DIFF
--- a/paramiko/ecdsakey.py
+++ b/paramiko/ecdsakey.py
@@ -20,7 +20,7 @@
 ECDSA keys
 """
 
-from cryptography.exceptions import InvalidSignature
+from cryptography.exceptions import InvalidSignature, UnsupportedAlgorithm
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import ec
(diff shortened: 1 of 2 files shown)


bitprophet:
Thanks! I think overall it might be safest to just add `UnsupportedAlgorithm` - I assume `AssertionError` was added for a reason and not sure we should take it out; and I wonder about `TypeError`, it's possible higher level libraries are currently catching that, in which case turning it into `SSHException` could cause issues there.

Tests around this would be great either way too - thanks!

tylergarcianet:
Good point. Thanks. I made the change. Looks like one of the builds failed in the CI when attempting to curl?

I'll see if I can work out some tests for this.

bitprophet:
This no longer applies cleanly (not exactly your fault :D) so I'll make the changes myself & credit in changelog. Thx!

PR #1274, opened 2018-08-27
Use Popen methods for closing proxy object

parantapa:
Using process.poll() in closed reaps up any dead child processes.

The unconditional os.kill can raise exception in case the the proxy died for some reason.
The process.terminate() call in close needs to be wrapped up in a try/catch as there may be a race condition, between the issue of process.poll() and process.terminate().


ploxiln:
related: #962

PR #1296, opened 2018-09-18
Remove unnecessary shim, BytesIO; io.BytesIO on all Pythons

jdufresne:
`io.BytesIO` is available on all supported platforms, including Python 2.7. For code requiring an in memory bytes stream, simply always use the stdlib `io.BytesIO`. Makes the code more forward compatible with Python 3.


jdufresne:
Thanks @mimi1vx I've made the suggested changes in the latest revision.

PR #1166, opened 2018-02-09
Add GSS-API support for using the 'gssapi' package (in addition to older 'python-gssapi')

akruis:
I rebased pull request #861 for 2.2. Additionally I improved backwards compatibility a bit (keep the internal name of the old python-gssapi class).

This pull request could be applied to 2.3 too.
@SebastianDeiss: please review this pull request.

Issue #584, opened 2015-09-08
Upgrade GSSAPI support to the 'gssapi' (not 'python-gssapi') module


metaperture:
[Maintainer edit: the action required on this ticket is to change Paramiko's GSSAPI functionality to use the API of `gssapi`, vs right now where it uses the API of `python-gssapi`. Then update `setup.py`.]

It looks like paramiko/ssh_gss.py refers to an older version of gssapi. See from https://github.com/paramiko/paramiko/blob/master/paramiko/ssh_gss.py#L253, all the references to gssapi.C_\* fail.

Note that gssapi 0.6.\* has these attributes (https://github.com/sigmaris/python-gssapi/blob/master/gssapi/__init__.py#L6), but the new gssapi (v >= 1.0, https://github.com/pythongssapi/python-gssapi/tree/master/gssapi/raw) doesn't appear to have these flags anywhere I can find.

Specifically, when I call connect in client.py, I get the following:

```
/home/mtartre/.conda/envs/std/lib/python3.4/site-packages/paramiko/ssh_gss.py in __init__(self,     auth_method, gss_deleg_creds)
    251 
    252         if self._gss_deleg_creds:
--> 253             self._gss_flags = (gssapi.C_PROT_READY_FLAG,
    254                                gssapi.C_INTEG_FLAG,
    255                                gssapi.C_MUTUAL_FLAG,

AttributeError: 'module' object has no attribute 'C_PROT_READY_FLAG'
```

I wish I could help fix this, but I'm out of my depth when it comes to C bindings like this. I'd be happy to test any ideas anyone has.

References:
- Issue https://github.com/paramiko/paramiko/issues/82
- Pull https://github.com/paramiko/paramiko/pull/267

bitprophet:
Thanks for the report!

linuxluser:
The [gssapi module](https://github.com/sigmaris/python-gssapi) which paramiko GSSAPI support was built with now states:
> There is a newer project which has superseded this one, https://github.com/pythongssapi/python-gssapi. Please use that package if suitable for your needs, as this project is unlikely to be developed further.

It looks like the dependency that Paramiko's GSSAPI support was built with seems to be deprecated in favor of a new fork of this module with the same name. If you `pip install gssapi` you will get the latest version which is now completely incompatible with what Paramiko currently expects to be using when you `import gssapi`.

Old python-gssapi module: https://github.com/sigmaris/python-gssapi
[...]

Issue #1191, opened 2018-04-19
Update documentation with non-optional dependencies


stevenwinfield:
Hi,

On [the installation page](http://www.paramiko.org/installing.html) at paramiko.org it says:
> Paramiko has only one direct hard dependency: the Cryptography library

...but in setup.py:
```    
install_requires=[
        'bcrypt>=3.1.3',
        'cryptography>=1.5',
        'pynacl>=1.0.1',
        'pyasn1>=0.1.7',
```

...which was a bit of a surprise!

Could the docs be updated, please?

bitprophet:
Pondering whether we can/should remove `pyasn1` on older versions; even though it is _removing_ something, not adding, it still worries me that it'll goof somebody up somewhere (esp eg downstream packagers whose scripts, etc might suddenly break.)

Think I'll just make the changes on master for now?

bitprophet:
Seems all right, updated the pyasn1 imports in the gssapi module (`paramiko.ssh_gss`) to all be inlined, and now I can import paramiko in a virtualenv without that module present. Tests all still pass etc.

So, Paramiko 2.5.x, when it comes out, will have this change in its code, and the docs will have the change as soon as I push. Thanks again! 👍

PR #1311, opened 2018-10-05
Enable GSSAPI tests and add support for the package 'gssapi'

akruis:
[Maintainer note: the original feature/request ticket for this is #584]

This pull request replaces pull request #1166.

Previously testing of GSSAPI (Kerberos) related functions required an
externally provided Kerberos environment. Therefore all GSSAPI tests were
skipped.
    
Now the package k5test is used to setup a self-contained Kerberos environment.
Because k5test requires the new GSSAPI, this commit also merges
pull request #1166 and fixes broken GSSAPI test. If k5test is not available
(i.e. on Windows), the tests still get skipped.
    
The test case test_kex_gss.test_2_gsskex_and_auth_rekey is expected to fail. This needs to be resolved as a separate issue.

This pull request does not change the flow of control (except for the additional case in ssh_gss) and does not touch non-GSSAPI/SSPI related code. Therefore I recommend inclusion in 2.4. With this patch applied, GSSAPI could be tested by travis.


akruis:
@SebastianDeiss Sebastian, provided you have time, could you review this pull request? And could you look at #1312 too?

SebastianDeiss:
@akruis I'm quite busy this week, but I try to take a look at it next week.

bitprophet:
FWIW this could probably get pushed onto master; it's in that gray area between bugfix and feature but given it involves dependencies & is a nontrivial shakeup of an established feature, my gut says feature.

But I'll perform any necessary merging/rebasing when I get to it so don't worry about opening yet another separate PR for now 😂 

I'm also real happy to see the notes about self-testing using the new library as that's always been a real sore spot for this feature set. Thanks!

SebastianDeiss:
@akruis LGTM.

JAORMX:
We need this patch, any chance to get it merged?

aaron-p-lehmann:
I'm with @JAORMX , getting this merged and a new egg/wheel made would save me a lot of shenanigans.

jborean93:
It would be great if this could be merged in at some point. Currently paramiko is used in a few places in Ansible but some other libraries use the newer gssapi library for Kerberos auth support. Due to the conflicts this causes import errors and paramiko cannot be used in those situations.

ecederstrand:
Ping? We also need this patch.

bitprophet:
Looking this over now, thoughts:

- I like that it adds tests ❤️ 
- `flake8` found that one of the new methods (`ssh_init_sec_context`) lacked an internal import from pyasn1
  - Normally, this would just be an understandable braino (those internal imports seem to have been a pattern from the original GSSAPI support patch, because of the optional nature of this feature)
  - However it makes me worry that that part of the code (if not other parts) are not being hit by the new tests 😢 
  - I fixed it and pushed to Travis to see how the new tests run, at any rate, but noting this for the record.
- I'm not doing a deep dive of most of the diff, since at this time I consider the GSSAPI support to be "community supported" and I have bigger fish to fry. Fixed one or two very minor things that jumped out at me and that's it.
- I'd love it if some of the folks in this ticket (esp @akruis and @SebastianDeiss but any help is good) could triple check the `1311-int` branch to make sure that it still works for them (it's this patchset + master merge + my minor tweaks/fixes as above).
  - I may delay merging this until I get at least one firm +1 on that. But if I do, I'll happily put out a Paramiko 2.6 as soon as that +1 occurs 👍

bitprophet:
Interestingly there's already some failures in Travis: https://travis-ci.org/paramiko/paramiko/jobs/539948050

- Suspect the one about `pyasn1` is just a minor oversight and we need the Travis GSSAPI section to `pip install pyasn1`, though this might be some unrelated issue around Cryptography (which recently-ish stopped using that lib, IIRC, and master Paramiko just landed a change requiring a newer Crypto).
- The other about SSH keys is weird, we have other unrelated tickets about new key types, but I haven't done anything with them yet, so...??? Maybe it'll magically go away when the other is fixed! That sometimes happens. Computers!

bitprophet:
Also, realized the docs did not get updated here:

- Clearly we need to `s/python-gssapi/gssapi` in the install docs, though I'll keep mention of `python-gssapi` since the code went to the effort of remaining backwards compatible.
  - Do we know what the minimum supported version of that new dependency is? I'm going to try for 1.0.0 on up, but...
- Changelog entry, which I will handle, noting that this is a backwards-compatible change

bitprophet:
Digging harder at that pyasn1 issue - I made some faulty assumptions:

- That `gssapi` was not already indicated in the .travis.yml - it is
- That `pyasn1` was a dependency of `gssapi` - it isn't

Turns out, ok: in a31818c28556055341ae3d249e3893d40eb2b232 (for #1191) I made pyasn1 optional, but I don't seem to ever have added it to the docs. Not sure if this is because it was an implicit dep of `python-gssapi` or not, but it's moot now I think.

bitprophet:
Cool, that seems to have fixed all the issues: https://travis-ci.org/paramiko/paramiko/builds/539961575

bitprophet:
Assuming my full rerun of that build's still green (initially only ran the Kerberos test cells), that means what's left is:

- Want others to triple check the branch still kosher for them
- Would like a sanity check on whether the `gssapi above 1.0.0` in the install docs is accurate or if we need to scope it to eg `>=1.5.0`

akruis:
Hi Jeff, I'm out of office this week, but I'll have a look at it during my leave next week.

akruis:
> Assuming my full rerun of that build's still green (initially only ran the Kerberos test cells), that means what's left is:
> 
>     * Want others to triple check the branch still kosher for them

I compared branch 1311-int with my pull request. It looks good. All tests pass.

>     * Would like a sanity check on whether the `gssapi above 1.0.0` in the install docs is accurate or if we need to scope it to eg `>=1.5.0`

We (s+c) use version 1.4.1 of the new gssapi. Therefore I propose `>=1.4.1`, see pull request #1461.

ecederstrand:
Seems tests are now green except for https://github.com/paramiko/paramiko/pull/1311/commits/f3af0b3e697adc8902039b21fde93871048160e4 which introduced a flake8 error (line too long).

bitprophet:
Back on this, thanks @akruis for the thumbs up & doc tweaks.

With fresh eyes, and spurred by that doc update, I am re-examining whether we can use `extras_require` here. There's two obvious ways we actually can:

- Just offer two different extras, eg `paramiko[gssapi]` for Unix and `paramiko[gssapi-win]` or something for Windows. Not ideal but still saves folks one or two minor steps.
- I'd forgotten that modern setup.py includes those semicolon-delimited dependency specifiers: https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies - suspect we can use this to avoid the previous bullet point.
  - I don't have a good Windows testbed to triple check, but can do my usual, which is to ensure it works OK for Unix and hope one of our Windows users files a bug (and uh, that I actually see that bug report 😒)

bitprophet:
Cool, that (ecae381a) seems to Just Work on my macOS system 👍

bitprophet:
Integration branch builds pass - https://travis-ci.org/paramiko/paramiko/builds/548752153

bitprophet:
Merged latest master in and fixed conflicts (and saw / fixed the flake8 failure @ecederstrand noticed). Now merging back into master & pushing - this should be done now. Will see whether I can easily bundle something else into a 2.6 release and then put that out. Thanks all!

bitprophet:
Ah, this won't autoclose because it's targeting 2.4 branch for w/e reason.

ecederstrand:
My GitHub skills are failing me. Does this mean that the `gssapi` package is supported in the latest release (`2.6`)?

ploxiln:
yes

Issue #1299, opened 2018-09-20
Feature Request: SilentHostKeyPolicy


mcandre:
Would be convenient for devops if paramiko included a *don't care* policy for host keys, maybe `paramiko.SilentHostKeyPolicy` or such.

smheidrich:
I'd like this as well. There is a very simple implementation of such a policy in [testinfra](https://github.com/philpep/testinfra)'s codebase called [``IgnorePolicy``](https://github.com/philpep/testinfra/blob/c9a06fc10d415f3ab5cf095aa27c1238f7d7234e/testinfra/backend/paramiko.py#L32-L35).

PR #1324, opened 2018-10-26
Help users ignore unimportant errors

mcandre:
Offer an SSH key policy for users who want to ignore unregistered key errors


mcandre:
This would close #1299

mcandre:
I just want to say that this PR was opened before Halloween, and it's quite small and clean, and it has not been reviewed by Christmas.

PR #1326, opened 2018-10-27
RFC-8322 SHA2 hash algorithm for RSA key signing/verifying

radssh:
Extend RSA key handling to support new SHA256 & SHA512 hashing for signatures. The RSA key format itself is not changing at all, just the algorithm for the hashing from SHA1 to SHA256/SHA512.

Code using RSAKey class will retain legacy SHA-1 hashing for signing/verifying, as the new algorithms are only enabled when using one of the two new key types RSA256Key or RSA512Key explicitly. This PR introduces the code support for the new keys, but no forced adoption or behavioral change to RSAKey (other than standardizing the key `generate()` methods to be classmethods rather than staticmethods).

Host key exchange _preferred_keys amended to include the new key/algo types, but **after** the legacy `ssh-rsa` type, so for now, that should only be visible in the case where hosts only offer rsa-sha2 keys, or the coder has explicitly altered the _preferred_keys list and should be aware of the listed caveat "do not reorder without reason!" :)


bitprophet:

Issue #852, opened 2016-11-26
Can we have paramiko respect the ControlMaster/ControlPersist options?


omani:
AFAIK paramiko does not support or respect the multiplexing capability of openssh (defined in .ssh/config).

```bash
ControlMaster auto
ControlPath $somepath
ControlPersist 3600 # 1 hour
```

Is this possible to use in paramiko somehow? is this planned to be implemented soon?

bitprophet:
No, AFAIK there's no support for this out of the box.

It's actually something I personally could use at my current job so I was planning to give it a whack sometime soon-ish. Thanks for making the ticket for me :)

bitprophet:
http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/PROTOCOL.mux?annotate=HEAD may be a good spot for me to start here. Haven't actually read it yet.

bitprophet:
Also, for search result purposes, noting that this feature is often referred to as 'multiplexing'.

omani:
hi @bitprophet,

just wanted to know if you have any news regarding this feature?

omani:
hi @bitprophet, any news about this issue?

omani:
@bitprophet, any news?

omani:
@bitprophet,

what is the status of this issue?

bitprophet:
Dude, asking over and over won't make me magically find more time! Sorry :(

omani:
Dude, not responding over and over won't make me think the answer is: "I have no time".

Sorry ;)

SethMichaelLarson:
Many people in Open Source would rather spend time actually working on projects (or heaven forbid, anything but open source for once) rather than triaging the hundreds of notifications you receive daily. Please be respectful to everyone's time. If the issue is really pertinent, you might take a first stab at a PR, even if it's not complete, just to generate discussion and possibly bring in others to help tackle this issue.

We're all in this together. :)

omani:
yes that's the way to go. but since bitprophet told us that he would have a look into this I wanted to get some feedback about his progress.

omani:
hi bitprophet,

what is the status of this issue?

ploxiln:
You can browse the issue and pull request activity of this project yourself. If you do, you will see that there is no progress being made on this feature. Asking periodically only results in "noise". The fastest way for you to get this feature in paramiko would most likely be for you to implement it yourself.

cpatulea:
[...]

PR #1341, opened 2018-11-25
Mux transport for client

radssh:
Draft implementation for #852 

Seems to work fine for normal open_session management, some quirks with port forwarding and global request/response handling, and catastrophic behavior if renegotiate_keys is called when in proxy mode :)

Some guesswork needed, as the OpenSSH client uses "passenger mode" instead of "proxy mode", but proxy mode seems a much better fit for paramiko, since the SSH protocol messages are already implemented, and the proxy mode only required a small set of changes to the packetizing (zero padding failed to work due to modulo calculation)

Supply a (new) controlpath parameter to `Transport()` or `SSHClient()`, and domain socket connection is attempted. If successful, it confirms that is is talking to a ControlMaster server, and switches the ControlMaster session into Proxy mode. The Transport then behaves as an authenticated transport, skipping the initial hostkey validation, key exchange, and user auth.

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -236,6 +236,7 @@ class SSHClient(ClosingContextManager):
         auth_timeout=None,
         gss_trust_dns=True,
         passphrase=None,
+        controlpath=None,
     ):
         """
         Connect to an SSH server and authenticate to it.  The server's host key
@@ -310,6 +311,8 @@ class SSHClient(ClosingContextManager):
             for the SSH banner to be presented.
         :param float auth_timeout: an optional timeout (in seconds) to wait for
             an authentication response.
+        :param string controlpath: path to unix domain socket of running
+            ControlMaster multiplexing agent.
 
         :raises:
             `.BadHostKeyException` -- if the server's host key could not be
@@ -328,6 +331,19 @@ class SSHClient(ClosingContextManager):
         .. versionchanged:: 2.4
             Added the ``passphrase`` argument.
         """
+        if controlpath:
+            try:
+                t = self._transport = Transport(
+                    (hostname, port), controlpath=controlpath)
+                # If this connection succeeds, most (but not all) of the
+                # SSHClient connect steps can be bypassed
+                if self._log_channel is not None:
+                    t.set_log_channel(self._log_channel)
+                t.start_client(timeout=timeout)
+                return
+            except SSHException as e:
+                # Quietly fallback to traditional connection steps
+                pass
         if not sock:
             errors = {}
             # Try multiple possible address families (e.g. IPv4 vs IPv6)
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -334,7 +334,8 @@ class SSHClient(ClosingContextManager):
         if controlpath:
             try:
                 t = self._transport = Transport(
-                    (hostname, port), controlpath=controlpath)
+                    (hostname, port), controlpath=controlpath
+                )
                 # If this connection succeeds, most (but not all) of the
                 # SSHClient connect steps can be bypassed
                 if self._log_channel is not None:
--- a/paramiko/common.py
+++ b/paramiko/common.py
@@ -174,6 +174,15 @@ CONNECTION_FAILED_CODE = {
     DISCONNECT_NO_MORE_AUTH_METHODS_AVAILABLE,
 ) = (7, 13, 14)
 
+(
+    MUX_MSG_HELLO,
+    MUX_C_PROXY,
+    MUX_S_PROXY,
+    MUX_C_ALIVE_CHECK,
+    MUX_S_ALIVE,
+    MUX_S_OK,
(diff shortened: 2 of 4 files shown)


mljli:
For others got "unsupported request" error, you need openssh-client 7.4p1 or newer to work.

radssh:
> For anyone who got "unsupported request" error, you need openssh-client 7.4p1 or newer to work.

Good catch - I did not realize this was a relatively recent addition to the mix protocol, and a bit surprised that the protocol version number didn't get bumped along with that change so it could be caught cleaner than that. Added that detail to the exception message.

mjtravers:
This works great. Any thought to adding the initialization/management of the ControlMaster connection and ControlPath file via paramiko rather than relying one that is already up and running?

radssh:
The mux server side implementation can be done, but is going to be a significant amount of work. It probably needs to be a full implementation covering both proxy mode (which I used in the client implementation here) as well as passenger mode, that makes use of directly passing file handles, that is supported by the Python3 standard library, but would need another dependency added for Python2 support.

t2y:
I confirmed this patch works for me. It saved me from annoyance. Great! 👍

mljli:
The shared connection closed periodically when I was using this PR. It does not happen every time but if you keep trying (for about 5 times or more), ssh will disconnect itself from the server. Here is the error message,
> debug3: mux_client_read_packet: read header failed: Broken pipe

I tried to find the cause but with no luck. My program exited normally and no exception was thrown.

FYI, I created a context manager to close the client automatically.
```
def ssh_to(host):
    import paramiko
    ssh_config = paramiko.config.SSHConfig()
    with open(osp.join(HOME, '.ssh', 'config')) as f:
        ssh_config.parse(f)
    host_config = ssh_config.lookup(host)

    controlpath = osp.expanduser(host_config['controlpath'])
    if not osp.exists(controlpath):
        controlpath = None

    ssh_client = paramiko.client.SSHClient()
    ssh_client.set_missing_host_key_policy(paramiko.client.AutoAddPolicy)
    ssh_client.connect(
            host_config['hostname'],
            port=host_config['port'],
            username=host_config['user'],
            controlpath=controlpath
    )
    yield ssh_client
    ssh_client.close()

# then use it
with ssh_to('x.x.x.x') as ssh_client:
    ssh_client.exec_command('ls')
```

mljli:
@radssh Please take a look at the log.
[paramiko.log](https://github.com/paramiko/paramiko/files/2959447/paramiko.log)

mljli:
@radssh Please take a look at the log.
[paramiko.log](https://github.com/paramiko/paramiko/files/2960659/paramiko.log)

radssh:
At a glance, the paramiko log (and the ssh client log for the ControlMaster) seem to indicate that the connection is getting dropped. I will do some code testing, based on the log, using the sftp for file transfers.

I tried your earlier code sample, and it looks like your context manager isn't really a context manager. I'm getting `AttributeError` about missing `__enter__`(Python3) or `__exit__` (Python2) methods. Since your log shows you have something working with the `SSHClient`, I'll skip trying to test with your context manager code for now.

mljli:
> At a glance, the paramiko log (and the ssh client log for the ControlMaster) seem to indicate that the connection is getting dropped. I will do some code testing, based on the log, using the sftp for file transfers.
> 
> I tried your earlier code sample, and it looks like your context manager isn't really a context manager. I'm getting `AttributeError` about missing `__enter__`(Python3) or `__exit__` (Python2) methods. Since your log shows you have something working with the `SSHClient`, I'll skip trying to test with your context manager code for now.

My bad, I forgot to paste some import statements. Yeah, I'm using sftp to transfer files because the bastion host says ssh2. I tried doing the same task using command line tools and the error never shows. So I guess there might be something wrong with paramiko. I've uploaded my real program. My shared connection get closed each time I run it. I changed its extension since Github does not accept .py. You have to modify some variables' values in the main function, otherwise the program won't work. Thanks for your help. I really appreciate it.

[sync.txt](https://github.com/paramiko/paramiko/files/2961496/sync.txt)

bauen1:
I would love to see this merged, even if only the ability to piggyback onto an existing socket.
This would make using [pyinfra](https://github.com/Fizzadar/pyinfra/) with ssh jump hosts that e.g. require multi factor authentication a lot easier and potentially faster.

rupertnash:
I just tried this and it works fine to connect, however when Paramiko disconnects it also hangs up the master.

Afraid I know little about ssh and networking so can't contribute to the patch

takluyver:
I have rebased this as #1857.

What do people think about having a separate method for connecting to a multiplexing socket, rather than an extra parameter to the `connect()` method? This would mean a couple of extra lines of code if you want to fallback to opening a new connection, but leaves the caller the choice of not falling back like that.

bitprophet:
Rolling into #1931.

Issue #1261, opened 2018-08-15
support KDF rounds for key generation


johnnybubonic:
if/when the (much more secure, by the way) new private key format is supported (as suggested in https://github.com/paramiko/paramiko/issues/1136#issuecomment-412232708 ), key generation should support # of rounds as well. per man page for ssh-keygen,

>      -a rounds
>              When saving a new-format private key (i.e. an ed25519 key or when the -o flag is set), this option specifies the number of KDF (key derivation function) rounds used.  Higher numbers result in slower passphrase verifica‐
>              tion and increased resistance to brute-force password cracking (should the keys be stolen).
> 
>              When screening DH-GEX candidates (using the -T command).  This option specifies the number of primality tests to perform.
>

Issue #1136, opened 2017-12-18
Ed25519 key generation support


max-wittig:
Nearly every key, exept the Ed25519 key offers a `generate` function in paramiko, which is used to generate a new key pair.

Is there a reason why the `ed25519` keytype doesn't have this class method?

From the documentation:
http://docs.paramiko.org/en/2.4/api/keys.html

![image](https://user-images.githubusercontent.com/6639323/34103452-0bdc38d4-e3ed-11e7-99ef-2f8cabf4e794.png)

dopaminereceptionist:
Also interested in this. Should be reasonably straight forward.

johnnybubonic:
for reference:

here's a python 2 implementation: https://github.com/pts/py_ssh_keygen_ed25519

here's the overview of the new "openssh-key-v1" format that ssh_ed25519 **always** uses: https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.key 
(which you should add support for for RSA and ECDSA keys as well, since openssh has announced that'll be the default at some time in the future anyways)

hope that helps

johnnybubonic:
@bitprophet looks like this should be much more simplified since the cryptography module [has now implemented ed25519](https://github.com/pyca/cryptography/issues/3509).

paprickar:
@bitprophet do you have a gut feeling when this can be implemented?

bitprophet:
I actually saw this independently the other day when digging into an unrelated key issue - I had need to run basic key generation and reached for Ed25519 out of "this is the current 'good key type' to use in a  vacuum". And yup. It's missing this!

I see I added it to the feature milestone some time ago. I have no firm ETA for it but see https://bitprophet.org/projects/#roadmap - this one absolutely qualifies for my next set of feature-adds. Expect I'd hope to get to this after the Python 2 removal release (again per that roadmap).

utkonos:
Could the cryptography `.generate()` just be wrapped/used directly using some of the code I wrote there? Unless I'm totally mistaken, this issue could be solved with a few lines of code and a few unit tests, no?

I can definitely write the code and PR to close this issue, I just want to make sure I'm on the right path before committing.

utkonos:
BTW: there's no `bits` or `curve` selection needed for generating `ed25519`. The method would just be `.generate()` and boom, it's done.

utkonos:
[...]

PR #1343, opened 2018-11-28
add support for new OpenSSH private key format

jaredhobbs:
This work is based off the work done in https://github.com/paramiko/paramiko/pull/618


jaredhobbs:
Probably. I'll investigate when I get some free time but I'm pretty swamped at the moment.

thomasd3:
Is there any ETA for when the new format could be supported?

carumusan:
Please merge this. Looks like since OpenSSH 7.8 (release 4-5 months ago) the new format is the default. https://www.openssh.com/txt/release-7.8

dgjustice:
Bump!  Fedora 29, openssh 7.9p1.  Paramiko can't read my private key files without this change.

bdrydyk:
Bumping this as well. This has broken my automation from OS X Mojave

mitom:
Is there anything blocking this that requires help? Lacking this is causing major issues for us.

kindlehl:
I'm currently having issues as well. Broken on Debian 9

jaredhobbs:
> Would it not be better to make it a raw string and drop flake comment?
> `r"^-{5}BEGIN (RSA|DSA|EC|OPENSSH) PRIVATE KEY-{5}\s*$"`
> https://lintlyci.github.io/Flake8Rules/rules/W605.html

Yes, fixed.

pgierz:
Any idea when this will go in?

paramiko/ecdsakey.py
```diff
@@ -283,12 +283,22 @@ def _from_private_key(self, file_obj, password):
...
+            try:
+                key = serialization.load_der_private_key(
+                    data, password=None, backend=default_backend()
+                )
+            except (ValueError, AssertionError) as e:
+                raise SSHException(str(e))
+        elif pkformat == self.PRIVATE_KEY_FORMAT_OPENSSH:
+            curve, verkey, sigkey = self._uint32_cstruct_unpack(data, "sss")
+            try:
+                key = ec.derive_private_key(sigkey, curve, default_backend())
```
ploxiln: this doesn't work: `sigkey` needs to be an integer, and `curve` needs to be translated into one of the Curve objects in `_ECDSA_CURVES`

noticed while working on my port and cleanup of this. here's a test to show the errors:
https://github.com/ploxiln/paramiko-ng/pull/13/commits/ab0389154ba4c083504fc97f3a973d09e08d5568
the fix is mixed into my large refactor, since I figured this out towards the end:
https://github.com/ploxiln/paramiko-ng/pull/13/commits/2a1b69e074897e0e08c6b3e2048b136f9a5a3fdd

paramiko/pkey.py
```diff
@@ -337,6 +382,140 @@ def _read_private_key(self, tag, f, password=None):
...
+        checkint1, checkint2, keytype, keydata = cstruct
+
+        if checkint1 != checkint2:
+            raise SSHException(
+                "OpenSSH private key file checkints do not match"
+            )
+
+        # Remove padding
+        padlen = byte_ord(keydata[len(keydata) - 1])
+        return keydata[: len(keydata) - padlen]
```
ploxiln: this unpad suffers from roughly the same bug I fixed for Ed25519Key in https://github.com/paramiko/paramiko/pull/1400

greenscar:
any movement? We all know Apple is the new Microsoft, doing everything their own way. However, It kinda sucks when I can't run Ansible / Testinfra from my mac box due to 
`paramiko.ssh_exception.SSHException: not a valid RSA private key file"`

joaquinclearmetal:
If this key is generated on macOS and the pub key is installed, it will break everywhere paramiko is used, not just from macOS.  Parmiko should support OpenSSH by support SSH key format document 13 years ago in RFC4716.

nickthorpe:
Bumping this as well...

newearthmartin:
Can someone please merge this? It has been open for months

joaquinclearmetal:
It would be nice if Paramiko supported SSH.  I have to recreate keys if they were created on macOS (and apparently this affects Fedora).

cybiere:
New format is also default in Debian, any update on this merge ?

ChrisEdgington:
I was able to fix this problem by converting my macos id_rsa to pem format via:

ssh-keygen -p -m PEM -f ~/.ssh/id_rsa

cbonitz:
OpenSSH formatted private keys seem to be the default on Ubuntu 19.04, just like on MacOS and Fedora as mentioned by previous commenters. This is not an edge case anymore, this will concern almost all users of unix-like operating systems very soon.

XVilka:
Would be nice to have this finally merged.

jamesharris-garmin:
Could we please just run black on pkey and resolve the tests on tests/test_pkey.py?

XVilka:
Sadly, it looks to me like the project is abandoned.

jamesharris-garmin:
I worry that this might be the case but it is worth a try anyway.

bitprophet:
Diving into reviewing/tweaking/merging this now. Thanks for everyone who's been keeping the patchset rolling, and for everyone's patience 🙏

bitprophet:
Merged to latest master & blackened, taking a look:

- No changelog? 😢  Added an entry and linked to most of the major related tickets (sob, there are so many) including crediting a pile of folks ❤️ 
- This code only seems to support `BEGIN OPENSSH PRIVATE KEY` style heads and not "real" RFC 4716 style headers which are `BEGIN SSH2 PRIVATE KEY`.
  - However, since `ssh-keygen` only appears to spit out the former header style, and given I'm not experienced with these formats yet myself, I'm going to leave this be.
    - Suspect OpenSSH departed from the RFC afterwards for whatever reason, and this problem domain (OpenSSH compatibility) is definitely one where practice trumps theory 😂 
  - Should be pretty easy to extend it later for the variant if folks need it.
- Renamed the new keys a bit, it wasn't clear to me offhand why they were named `<size>_o.key`, guessing o for OpenSSH, but I like obvious naming...
- Not a huge fan of the (as far as I can tell) internal-only format signifier constants (eg `PKey.PRIVATE_KEY_FORMAT_ORIGINAL`) but it's not a great use of my time to clean things up to that level, so I settled for:
  - making them private so they're not part of the public API and can be tweaked later
  - updating the exceptions raised when they go off the rails (they were not consistent or terrifically readable).
  - also I'm honestly not even sure how we can end up in that particular code branch (the code that spits out the pkformat value would raise its own exception first!) but again, going to avoid rocking the boat too much for now, in case I'm misreading this.
- More to be edited in as I go...

bitprophet:
This will be out in Paramiko 2.7 (only one more ticket after this one before that's cut!)

grant-humphries:
@bitprophet thank you for your work on this!!

jaredhobbs:
thanks so much!

johnnybubonic:
yay

please don't forget about #1136 and #1261 too! they should be able to be resolved now that this format's supported.

bitprophet:
@johnnybubonic Put 'em in a milestone, I can't say when they'll be gotten to but at least they're on my radar now, thank you!

PR #1354, opened 2018-12-15
pin pytest-xdist to < 1.25.0

reaperhulk:
1.25.0 requires pytest >=3.6, but pytest-relaxed doesn't support pytest >3.2.x. This temporary pin can work around that issue. Without this the test suite is broken (although ultimately pytest-relaxed should be updated to support newer pytest).

This affects the cryptography downstream tests right now so we'd like to get it resolved ASAP.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,7 +5,7 @@ invocations>=1.2.0,<2.0
 pytest>=3.2,<3.3
 pytest-relaxed==1.1.4
 # pytest-xdist for test dir watching and the inv guard task
-pytest-xdist>=1.22,<2.0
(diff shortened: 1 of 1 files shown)


bitprophet:
Grumble, I don't recall what in pytest 3.2 broke my dumb custom plugin, so this is doubtless faster for now. Thanks!

bitprophet:
Cherry-picked back to 2.0 and merged up.

bitprophet:
Travis still running but Works On My Machine™ ⭐️  – assuming you're testing against branches/master/HEAD and not tags, I assume this does it for ya? @ me in IRC otherwise.

PR #1359, opened 2018-12-21
request_port_forward should always send server allocated port

maximeborges:


pyhedgehog:
Oh, no! It should not!
`request_port_forward()` is logical equivalent of openssh `RemoteForward` option.
In practical cases I've seen you almost use specific port you want to listen on.

maximeborges:
I forgot the context for this (6 years old!) PR. 
I think it was for an attempt to make a service similar to ngrok, but no idea why I needed that change.
Anyway, closing.

PR #1360, opened 2018-12-23
handling of terminated proxy commands

p-sherratt:
The existing code will not robustly detect if the proxy command has terminated;

In `recv`, since we are usually just polling and not performing any operation on the pipe, no operation on the pipe is performed for it to detect/raise an error condition.  `send` is OK, this should raise a BrokenPipeError, but I believe we should poll the subprocess to check it's still alive within the `recv` loop.

Also, the `closed` property currently checks returncode, which would never be set without polling the subprocess anyway.


p-sherratt:
related: #1274, #962

PR #1366, opened 2019-01-07
Correcting documentation: Python 'b' flag has an effect on readline

martinprikryl:


hiratara:
I agree with this PR. I made the same mistake by reading documents.

PR #1388, opened 2019-02-14
Use pytest.mark.xfail instead of broken pytest_relaxed

mimi1vx:
pytest_relaxed is broken with newer versions on pytest. And pytest itself has needed funcionality

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,9 +1,7 @@
 # Invocations for common project tasks
 invoke>=1.0,<2.0
 invocations>=1.2.0,<2.0
-# NOTE: pytest-relaxed currently only works with pytest >=3, <3.3
-pytest>=3.2,<3.3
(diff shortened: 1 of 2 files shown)


mimi1vx:
@alex  travis fail looks like wrong setup for python 3.4

mimi1vx:
@bitprophet btw. I have teste paramiko with pytest-4.5.x and all PASS :D

bitprophet:
I was hoping to eventually migrate the tests to use pytest-relaxed's style of organization, so I'd rather go fix that to work with newer pytest instead (https://github.com/bitprophet/pytest-relaxed/issues/2). Thanks though! Gonna go poke that now because why not.

bitprophet:
Got that sorted, pytest-relaxed 1.1.5 is here and with it the latest pytest. Hope this is good on your end too after the update 👍

PR #1399, opened 2019-02-28
`Transport.connect` without password or key

MrMino:
This makes the method attempt a no authentication connect if no password or key is supplied.
This also makes the method behave as promised by `:raises:` in its docstring. Therefore the `.. note::` can be removed.


MrMino:
I now see that many things around `Transport` depend on `Transport.connect()` failing silently, which I didn't anticipate. Is this stemming from some conscious design decision, or is it more of a technical debt thing that would require a refactor and changes in the class interface?

Issue #1402, opened 2019-03-05
Does SSHClient really needs to have that complicated socket creation logic?


MrMino:
In `SSHClient.connect()`, the [first part of the subroutine](https://github.com/paramiko/paramiko/blob/master/paramiko/client.py#L331) is responsible for opening a socket. This takes about 50 lines and spans two methods.

Question: does it really have to be so complicated given the existence of [`socket.create_connection()`](https://docs.python.org/2/library/socket.html#socket.create_connection)?

`create_connection()` was added in python 2.6 and, if used in `connect()`, it would take care of finding the right address family for the socket and connecting it.

The only breaking change coming with this, is that the `NoValidConnectionsError` wouldn't be used anymore, but [that doesn't seem to mentioned in the documentation of `connect()` anyway](http://docs.paramiko.org/en/2.4/api/client.html#paramiko.client.SSHClient.connect), and from my own experience, seeing it in a traceback leads to more confusion than value.

Issue #22, opened 2011-06-24
doesn't fall back on ipv4 route when there is no ipv6 route


driesdesmet:
I can't connect to a ssh server that has an AAAA record from a client that doesn't have an ipv6 route:

Python 2.6.1 (r261:67515, Jun 24 2010, 21:47:49) 
[GCC 4.2.1 (Apple Inc. build 5646)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import paramiko
>>> client=paramiko.SSHClient()
>>> client.connect(hostname='ssh.alwaysdata.com')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Python/2.6/site-packages/paramiko/client.py", line 296, in connect
    sock.connect(addr)
  File "<string>", line 1, in connect
socket.error: [Errno 65] No route to host

I would expect paramiko to use IPv4 in this case. Connecting with os x's default ssh client works as expected.

tgoodlet:
Oh man. This still isn't fixed hehe?
ipv6 is like trying to push craft beer on people :D

bitprophet:
Whoa now, I'm actually drinking a craft beer right now! ([Proof](https://untappd.com/user/bitprophet/checkin/152588453))

Interestingly I could swear I remembered merging a fix related to this, but maybe it's just my memory of the comments I left above. Gonna poke & see.

bitprophet:
Confirmed that the current state of the code has the problem this is fixing (tries only 1st resolved address/family).

The latest state of https://github.com/Bluehorn/paramiko/compare/issue22 looks good. My only worry is if some platforms not under test raise socket exception types not included in `(ECONNREFUSED, EHOSTUNREACH)` (e.g. I remember other issues discovering fun subtle differences between Mac/BSD and Linux, where the same scenario yielded different exception numbers.) However that's not worth blocking a merge.

bitprophet:
As I merge I notice that my earlier concern still isn't fully addressed, if we encounter errors which *are* conn-refused/host-unreachable, and they persist for *all* address families, they will be hidden and then we'll just die farther down the file when `socket` is unset/None.

Will attempt to recreate a few scenarios to confirm that this is an issue & to verify a fix.

yaccz:
I think paramiko should also accept the connection hints (addrinfo) in a user controlled way so the user can for example say he wants just ipv4 addresses

bitprophet:
[...]

PR #1405, opened 2019-03-06
Simplify sock creation

MrMino:
See #1402


MrMino:
@bitprophet @bskinn I've refreshed this, it's no longer rotten.

Do you think this is viable?

bitprophet:
@MrMino argh, I hate to ask this after you just did that, but: can you repoint it at the 3.0 branch? might be worth including this in that release, since it'd be backwards incompatible (maybe, if anyone is in fact catching that exception; that was added during my tenure so presumably it fills a need _somewhere_...).

Otherwise, that would at least get you in good stead for inclusion in 3.1 or 4.x (which will come out a lot sooner than 3.0 did - I punted a lot of small/medium changes from this release to that one).

MrMino:
I added a changelog note, not sure if it's the right wording for ir.

bitprophet:
I dug into why that exception exists, and it's because of some very old problems raised both here and in Fabric. Eg #22 (which hilariously is such a small number it breaks Github's autocomplete) and fabric/fabric#1263 among others: when you try a bunch of possible IPv4/6 addresses in order and _none_ of them succeed, naïvely you will end up just raising the last error by itself.

Which ends up being terribly misleading for many users, in the same way that #387 is for "try all these keys to auth". You really want a specific exception signaling "we tried everything and it all failed". Thus, the custom exception.

What stinks is that the new stdlib socket creation method has _the exact same problem_ until Python 3.11!!

So I think I need to reject/hold this until we can safely drop support for Python <3.11, which will be some time I'm afraid. I'll leave it open for now in case I'm missing something. Thanks & sorry!

bskinn:
> @bitprophet: "What stinks is that the new stdlib socket creation method has the exact same problem until Python 3.11!!"

Exception groups to the rescue?

bitprophet:
> > @bitprophet: "What stinks is that the new stdlib socket creation method has the exact same problem until Python 3.11!!"
> 
> Exception groups to the rescue?

No, they just added an opt-in kwarg to the fancy connect method in 3.11 that sounds like it does what our custom code does today (raise an aggregate meta-exception).

MrMino:
> I dug into why that exception exists, and it's because of some very old problems raised both here and in Fabric. Eg #22 (which hilariously is such a small number it breaks Github's autocomplete) and [fabric/fabric#1263](https://github.com/fabric/fabric/issues/1263) among others: when you try a bunch of possible IPv4/6 addresses in order and _none_ of them succeed, naïvely you will end up just raising the last error by itself.

Uhh... yea, I hate that. I didn't realize that this is the case for the builtin. This PR was made out of confusion about `NoValidConnectionsError` - it is undocumented. The docs for state `socket.error`, which is the base class of the former.

I'm happy to hold off until 3.11, and add the kwarg then.

Or... we could create a `backport.*` of the 3.11 thing. That way the users get the 3.11 treatment without having to wait 2 major versions.

bitprophet:
I'll document the error class now, since yea, shouldn't have to do archaeology just to figure out why it exists. Otherwise, 🤔 

I don't hate the backport idea, OTOH I personally def don't have the bandwidth for yet another PyPI entry right now 😂  we could just vendor it but that's usually not ideal either unless it's a "get & go" tool like Invoke.

MrMino:
What I meant by "undocumented" is that `SSHClient.connect` docs do not mention it:

https://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient.connect

> I don't hate the backport idea, OTOH I personally def don't have the bandwidth for yet another PyPI entry right now joy we could just vendor it but that's usually not ideal either unless it's a "get & go" tool like Invoke.

I might try it, no promises though. I had creating a backport lib on my bucket list, for the sake of seeing what it takes.

bskinn:
The new feature in 3.11 _is_ backed by exception groups, I checked.

I don't know if a feature-matched backport is even technically possible. Not an expert, though, so don't let me stop you from trying.

bitprophet:
> What I meant by "undocumented" is that [`SSHClient.connect` docs](https://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient.connect) do not mention it.

Ah, gotcha. That's probably because it didn't "seem" necessary to update the docstring given that exception class was implemented backwards compatibly, but from this remove, I also don't see the point in not mentioning it. See https://github.com/paramiko/paramiko/commit/45de33414fbd253262d8d7135e58427255c5fc5f !

PR #1407, opened 2019-03-08
Add terminal window size resize on posix_shell

hughesxu:
Use system I/O interface to get real time Terminal Window size, then adjust window size.

### DIFF
--- a/demos/interactive.py
+++ b/demos/interactive.py
@@ -40,7 +40,8 @@ def interactive_shell(chan):
 
 def posix_shell(chan):
     import select
-
+    import struct
+    import fcntl
     oldtty = termios.tcgetattr(sys.stdin)
     try:
         tty.setraw(sys.stdin.fileno())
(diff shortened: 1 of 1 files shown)


demos/interactive.py
```diff
@@ -48,6 +50,11 @@ def posix_shell(chan):
         chan.settimeout(0.0)
 
         while True:
+            # Resize terminal size every loop to make change timely
+            s = struct.pack("HHHH", 0, 0, 0, 0)
+            t = fcntl.ioctl(sys.stdout.fileno(), termios.TIOCGWINSZ, s)
+            winsize = struct.unpack("hhhh", t)
+            chan.resize_pty(width=winsize[1], height=winsize[0])
```
ploxiln: This is just a demo but ... this is a lot of redundant traffic to send over the channel. This should only be sent if width or height changes. But also, it's a lot of ioctl() syscalls to check every loop too, I think you can add a signal handler for SIGWINCH to get notifications that terminal size changed.

bitprophet:
Thanks for submitting this! However I'm afraid I can't merge it for a few reasons:

* In general I am phasing these demos out. They're an antipattern (people keep copy-pasting them even though they are not tested or upkept code in the main part of the tree). Code needs to be living and tested.
* The living tested real world "demo" of Paramiko is Fabric, and I just added this feature over there! It ended up being essentially what ploxiln suggested: https://github.com/fabric/fabric/commit/c22a34942f1a3f0201d4f2f2f09adb9d39cb881a - a signal handler is a very efficient way of handling this and should work on most POSIX platforms.

PR #1408, opened 2019-03-10
add a missing comma

fchapoton:


MrMino:
I don't think there should be a comma there. It's a concatenation of two strings, not two separate items.

fchapoton:
oh, indeed, sorry

Issue #1149, opened 2018-01-11
Feature request: accept a list of keys in SSHClient.connect()'s `pkey` argument


MrMino:
As `key_filename` accepts a string *or* list of filenames, for the sake of consistency, it would be nice if `pkey` also accepted a list.

As I understand it, it would require changing the `pkey` part in `SSHClient._auth()`, am I right?

PR #1411, opened 2019-03-12
Accept a list in `pkey` parameter of `SSHClient.connect()`

MrMino:
See #1149

This one is done by analogy to `key_filenames` logic - both parts are very similar. Kinda smells like code duplication to me, I think it could be refactored to make `SSHClient._auth()` take just a list of pkeys.


MrMino:
Might be useful to #387

XVilka:
Any updates on this one? Why not yet merged?

PR #1318, opened 2018-10-12
Highlight python syntax in readme

ocaballeror:
So we go from this:

screenshot_2018-10-12 ocaballeror paramiko 1

to this:

screenshot_2018-10-12 ocaballeror paramiko

PR #1415, opened 2019-03-19
Fix python syntax highlighting

ZhukovAlexander:


GPHemsley:
Superseded by #1504.

ZhukovAlexander:
@GPHemsley personally, I don't mind. But I feel that would probably be a bit more engaging for contributors if their changes were merged in order in which they were created, rather then prioritizing "bigger" pull requests that incorporate their diffs.

GPHemsley:
> @GPHemsley personally, I don't mind. But I feel that it would probably be a bit more engaging for contributors if their changes were merged in order in which they were created, rather then prioritizing "bigger" pull requests that incorporate their diffs.

I don't necessarily disagree with you, but there is a large backlog of PRs waiting to merge and it doesn't help having duplicates.

And besides, #1318 offered the same patch first.

Issue #1068, opened 2017-09-19
AttributeError: 'module' object has no attribute 'GSSException'


dtrizna:
Hi, guys!

Facing the Exception mentioned in title simply by importing the latest version (2.3.0) of paramiko:

`>>> import paramiko`
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "/usr/lib/python2.7/site-packages/paramiko/__init__.py", line 31, in <module>
>     from paramiko.transport import SecurityOptions, Transport
>   File "/usr/lib/python2.7/site-packages/paramiko/transport.py", line 38, in <module>
>     from paramiko.auth_handler import AuthHandler
>   File "/usr/lib/python2.7/site-packages/paramiko/auth_handler.py", line 48, in <module>
>     from paramiko.ssh_gss import GSSAuth, GSS_EXCEPTIONS
>   File "/usr/lib/python2.7/site-packages/paramiko/ssh_gss.py", line 54, in <module>
>     GSS_EXCEPTIONS = (gssapi.GSSException,)
> AttributeError: 'module' object has no attribute 'GSSException'

Problem appeared after upgrade this evening:
`pip install --upgrade paramiko`

Re-installation of modules doesn't help:
`pip unistall paramiko; pip install paramiko`
neither
`pip install --force-reinstall --upgrade paramiko`

Version of pip:
`[root@hostname]# pip --version
pip 9.0.1 from /usr/lib/python2.7/site-packages (python 2.7)`

Python:
> Python 2.7.5 (default, Nov  6 2016, 00:28:07)
> [GCC 4.8.5 20150623 (Red Hat 4.8.5-11)] on linux2

It this fault in my installation?

ploxiln:
That `GSSException` thing was introduced in #1061. But the other cause of your problem might be that you have the "wrong" gssapi installed (whereas having no gssapi module installed would be OK) - see #861 

cc @akruis

b2sweety:
Can someone advise on how to uninstall gssapi module?  The playbooks are failing for  Aruba and IOS network modules. 

e<1x.xx.xx.x1> socket_path: None
fatal: [Aruba-IAP]: FAILED! => {
"changed": false,
"failed": true,
"msg": "unable to open shell. Please see: https://docs.ansible.com/ansible/network_debug_troubleshooting.html#unable-to-open-shell"
}
Please see: https://docs.ansible.com/ansible/network_debug_troubleshooting.html#unable-to-open-shell"}

[...]

Issue #1335, opened 2018-11-13
python gssapi support is written against dead experimental version of gssapi


mimi1vx:
https://github.com/pythongssapi/python-gssapi don't have this exception.

version of gssapi with this Exceptions (https://github.com/sigmaris/python-gssapi) has noted from 2015 is dead and please use version above

```
Traceback (most recent call last):
  File "/usr/bin/repose", line 11, in <module>
    load_entry_point('repose==1.4.0', 'console_scripts', 'repose')()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 484, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2725, in load_entry_point
    return ep.load()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2343, in load
    return self.resolve()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2349, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/lib/python3.6/site-packages/repose/main.py", line 5, in <module>
    from .argparsing import get_parser
  File "/usr/lib/python3.6/site-packages/repose/argparsing.py", line 9, in <module>
    from .host import ParseHosts
  File "/usr/lib/python3.6/site-packages/repose/host.py", line 5, in <module>
    from .target import Target
  File "/usr/lib/python3.6/site-packages/repose/target/__init__.py", line 6, in <module>
    from ..connection import Connection, CommandTimeout
  File "/usr/lib/python3.6/site-packages/repose/connection.py", line 11, in <module>
    import paramiko
  File "/usr/lib/python3.6/site-packages/paramiko/__init__.py", line 22, in <module>
    from paramiko.transport import SecurityOptions, Transport
  File "/usr/lib/python3.6/site-packages/paramiko/transport.py", line 38, in <module>
    from paramiko.auth_handler import AuthHandler
  File "/usr/lib/python3.6/site-packages/paramiko/auth_handler.py", line 72, in <module>
    from paramiko.ssh_gss import GSSAuth, GSS_EXCEPTIONS
  File "/usr/lib/python3.6/site-packages/paramiko/ssh_gss.py", line 55, in <module>
    GSS_EXCEPTIONS = (gssapi.GSSException,)
AttributeError: module 'gssapi' has no attribute 'GSSException'
[...]

PR #1439, opened 2019-05-05
If the 'wrong' gssapi is present don't use it

joelnb:
See #1068 for details of the issue. This ensures that having the 'wrong' module is treated the same as not having the module at all.

On Ubuntu 16.04 I have installed freeipa-client package which pulls in the [python-gssapi](https://packages.ubuntu.com/xenial/python-gssapi/) package. When I try to import paramiko on this system I get this error (same as the issue linked above):

```console
# python
Python 2.7.12 (default, Nov 12 2018, 14:36:49) 
[GCC 5.4.0 20160609] on linux2
>>> import paramiko
  File "/usr/local/lib/python2.7/dist-packages/paramiko/__init__.py", line 22, in <module>
    from paramiko.transport import SecurityOptions, Transport
  File "/usr/local/lib/python2.7/dist-packages/paramiko/transport.py", line 38, in <module>
    from paramiko.auth_handler import AuthHandler
  File "/usr/local/lib/python2.7/dist-packages/paramiko/auth_handler.py", line 72, in <module>
    from paramiko.ssh_gss import GSSAuth, GSS_EXCEPTIONS
  File "/usr/local/lib/python2.7/dist-packages/paramiko/ssh_gss.py", line 55, in <module>
    GSS_EXCEPTIONS = (gssapi.GSSException,)
AttributeError: 'module' object has no attribute 'GSSException'
```

I think the reason for this is because the package installs the ['gssapi' module](https://pypi.org/project/gssapi/) which is no longer maintained, while paramiko expects the ['python-gssapi' module](https://pypi.org/project/python-gssapi/) which is maintained. Because these modules both use the name 'gssapi' paramiko doesn't know which it is actually getting until it tries to access something which doesn't exist in the module I have.

The suggested workaround is to `pip uninstall gssapi` but as I mentioned I cannot do that because it is a system package which is required by other packages which I definitely need.

This PR just changes the behaviour so that if the imported 'gssapi' module does not have a GSSException member then it is treated the same as if the module was not found at all. This works as a solution for me because I do not ever need paramiko to do GSS authentication.

### DIFF
--- a/paramiko/ssh_gss.py
+++ b/paramiko/ssh_gss.py
@@ -48,7 +48,10 @@ _API = "MIT"
 try:
     import gssapi
 
-    GSS_EXCEPTIONS = (gssapi.GSSException,)
+    try:
+        GSS_EXCEPTIONS = (gssapi.GSSException,)
+    except AttributeError:
+        raise ImportError("Incorrect gssapi module available")
 except (ImportError, OSError):
     try:
         import pywintypes


ploxiln:
It's actually the opposite - the "python-gssapi" module which paramiko is compatible with is the longer-unmaintained one.

See #1311 #1335

joelnb:
Ah, thanks for the information. #1311 seems like a proper fix rather than just hiding the issue as it seems I have done here.

I will try to just use a distro version of paramiko for now and it will probably be compatible with the gssapi module I currently have.

PR #1460, opened 2019-06-19
clean up SSHException subclass args and string methods

### LINKED ISSUE #1440: SSHException should define _str_ method
Most `paramiko.ssh_exception.SSHException` subclasses call `SSHException.__init__(self, message)` with some nicely crafted message.
However, since they overwrite `self.args = (...)` just a few lines later, the `Exception.args` used in `Exception.__str__` don't return the `message` but basically `str(self.args)`.

```python
class BadHostKeyException(SSHException):
    def __init__(self, hostname, got_key, expected_key):
        message = (
            "Host key for server {} does not match: got {}, expected {}"
        )  # noqa
        message = message.format(
            hostname, got_key.get_base64(), expected_key.get_base64()
        )
        SSHException.__init__(self, message)
        self.hostname = hostname
        self.key = got_key
        self.expected_key = expected_key
        # for unpickling
        self.args = (hostname, got_key, expected_key)
```

For example:

```python
In [1]: from paramiko.ssh_exception import BadHostKeyException                                                                                                            

In [2]: from paramiko.ed25519key import Ed25519Key                                                                                                                        

# Assume this key exists
In [3]: key = Ed25519Key.from_private_key_file('ssh_host_ed25519_key')                                                                            

In [4]: exc = BadHostKeyException('somehost', key, key)                                                                                                                   

In [5]: exc.args                                                                                                                                                          
Out[5]: 
('somehost',
 <paramiko.ed25519key.Ed25519Key at 0x7f6dcc122b00>,
 <paramiko.ed25519key.Ed25519Key at 0x7f6dcc122b00>)

In [6]: str(exc)                                                                                                                                                          
Out[6]: "('somehost', <paramiko.ed25519key.Ed25519Key object at 0x7f6dcc122b00>, <paramiko.ed25519key.Ed25519Key object at 0x7f6dcc122b00>)"
```

ploxiln:
Setting "self.args" is equivalent to calling the parent class
`__init__()` method with those args, and also overrides the only
thing that parent class `__init__()` method did. It is confused.

What is actually desired is to override the `__str__()` method with
a custom formatted exception message. The default for Exception
with zero args is blank, with one arg is the str of that arg,
with multiple args is just the repr of the args tuple.
(That never includes the class name. So, we want custom str.)

fixes #1440

### DIFF
--- a/paramiko/ssh_exception.py
+++ b/paramiko/ssh_exception.py
@@ -56,20 +56,16 @@ class BadAuthenticationType(AuthenticationException):
     .. versionadded:: 1.1
     """
 
-    #: list of allowed authentication types provided by the server (possible
-    #: values are: ``"none"``, ``"password"``, and ``"publickey"``).
     allowed_types = []
 
     def __init__(self, explanation, types):
-        AuthenticationException.__init__(self, explanation)
+        AuthenticationException.__init__(self, explanation, types)
         self.allowed_types = types
-        # for unpickling
-        self.args = (explanation, types)
 
     def __str__(self):
-        return "{} (allowed_types={!r})".format(
-            SSHException.__str__(self), self.allowed_types
-        )
+        msg = "Bad authentication type, allowed types: "
+        msg += ",".join(self.allowed_types)
+        return msg
 
 
 class PartialAuthentication(AuthenticationException):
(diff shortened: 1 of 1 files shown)


paramiko/ssh_exception.py
```diff
@@ -114,18 +114,17 @@ class BadHostKeyException(SSHException):
...
         self.hostname = hostname
         self.key = got_key
         self.expected_key = expected_key
-        # for unpickling
-        self.args = (hostname, got_key, expected_key)
+
+    def __str__(self):
+        return "Host key for server %s does not match: got %s, expected %s" % (
+            self.hostname,
+            self.got_key.get_base64(),
```
ploxiln: just spotted a bug here (confusing arg vs prop name), hang on a sec

ploxiln: squashed in the fix

bitprophet: Argh I'm literally working on this myself right now, believe I caught this tho.

bitprophet:
Should be good now, put my own changes on top. T. Hanks!

PR #1471, opened 2019-07-02
add auth_interactive_dumb_handler to client

mrgum:
Hello

This is change to client to add a auth_interactive_dumb_handler to connect, this is passed to _auth and used in the two_factor_auth

I've not added any tests but have tested it for my user case with the handler below (my own exceptions etc)

Anything more I need to do?

                def mfaduohandler(title, instructions, prompt_list):
                    answers = []
                    if instructions is None:
                        self.logger.error('instructions from server is None - send help')
                    elif instructions == "":
                        self.logger.info('no instructions from server, assuming wait for user to accept')
                        answers.append("ok")
                        return [echo and raw_input(prompt) or getpass(prompt) for (prompt, echo) in prompt_list]
                    elif instructions.strip().startswith('Please enroll at '):
                        duoenrolurl = re.search("(?P<url>https?://[^\s]+)", instructions.strip()).group("url")
                        self.logger.info('user does not have a enroled duo device sending them to {}'.format(duoenrolurl))
                        raise DuoEnrolException(duoenrolurl)
                    else:
                        self.logger.info('do not know what to do with instructions {}'.format(instructions.strip()))
                    return answers


mrgum:
I've stopped using paramiko so I'll close

PR #1478, opened 2019-07-16
Proxy Protocol support

jlutran:
Here's a patch to support [Proxy protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt). Both protocol versions are currently supported but I can work on a cleaner version including tests if you're interested ;-)


MrMino:
Is there any documentation for this?

jlutran:
> Is there any documentation for this?

Documentation to use this patch or documentation on proxy protocol ?

- For this patch, you just need to add `t.proxy_protocol = True` to [demo_server example](https://github.com/paramiko/paramiko/blob/master/demos/demo_server.py) anywhere between L140 and L150 to add proxy protocol support.

- For the proxy protocol itself, I didn't find any better doc than [HAproxy's](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt).

Issue #1141, opened 2017-12-26
Paramiko considerably slower in Python 3 than Python 2


dopaminereceptionist:
I have a very basic test script.

```
import sys

import aaargh
from paramiko import SSHClient, AutoAddPolicy

cli = aaargh.App()

@cli.cmd
@cli.cmd_arg('hostname')
@cli.cmd_arg('command')
@cli.cmd_arg('port')
def ssh(hostname, command, port=22):
    with SSHClient() as ssh_client:
        ssh_client.set_missing_host_key_policy(AutoAddPolicy())
        ssh_client.connect(hostname=hostname,
                           port=port,
                           username='root',
                           allow_agent=False,
                           look_for_keys=False,
                           key_filename='/etc/ssh/id_rsa')
        stdin, stdout, stderr = ssh_client.exec_command(command)
        return stdout.read()

if __name__ == '__main__':
    output = cli.run()
    try:
        sys.stdout.buffer.write(output)
    except Exception:
        # For python 2
        sys.stdout.write(output)
```

Never had an issue with this using Python 2. However, with Python 3 there is a non-linear slowdown based on the size of the returned data.

```
# First, verify the length of the outputted data.
python3 ssh.py ssh 127.0.0.1 'dd if=/dev/zero bs=10M count=1' 4000 | wc -c
10485760
# Second, time it in Python 3.
$ time python3 ssh.py ssh 127.0.0.1 'dd if=/dev/zero bs=10M count=1' 4000 | sha256sum -
e5b844cc57f57094ea4585e235f36c78c1cd222262bb89d53c94dcb4d6b3e55d  -

real    0m3.912s
user    0m1.612s
sys     0m2.252s
Third, time it in Python 2.
$ time python2 ssh.py ssh 127.0.0.1 'dd if=/dev/zero bs=10M count=1' 4000 | sha256sum -
e5b844cc57f57094ea4585e235f36c78c1cd222262bb89d53c94dcb4d6b3e55d  -

real    0m0.565s
user    0m0.364s
sys     0m0.072s
Now, try 20MB instead.
$ time python3 ssh.py ssh 127.0.0.1 'dd if=/dev/zero bs=20M count=1' 4000 > /dev/null

real    0m14.449s
user    0m6.032s
sys     0m8.356s
$ time python2 ssh.py ssh 127.0.0.1 'dd if=/dev/zero bs=20M count=1' 4000 > /dev/null

real    0m0.886s
user    0m0.456s
[...]

PR #1479, opened 2019-07-18
issue #1141: avoid quadratic buffer management on Python 3.x

dw:
When size is negative or None, it is possible for the result of the read to grow without bound. Since result is managed through repeat string concatenation, runtime is quadratic in the number of successful calls to
_read().

In Python 2 this had reasonable runtime because Python/ceval.c [contained a special case for appending to a string that could be safely mutated](https://github.com/python/cpython/blob/v2.7.11/Python/ceval.c#L5109-L5188), as the only reference to it existed on the stack, however during the great Python 3.x transition this special case was [rewritten to only support Unicode](https://github.com/python/cpython/blob/v3.5.1/Python/ceval.c#L5211-L5259), causing a huge amount of heap churn and memory copies on 3.x.

This was diagnosed using a combination of `/usr/bin/time`, Linux `perf` and `py-spy` run against the reproduction from #1141 to isolate the offending function.

Perf output showing 87% CPU spent in `memcpy()` and soft faults:

Screenshot from 2019-07-18 11-16-38

`/usr/bin/time` output confirming many soft faults due to repeatedly allocated heap having no backing pages on each loop iteration:

````
	Command being timed: "perf record -o perf.3 -g python3 ssh.py ssh 127.0.0.1 dd if=/dev/zero bs=20M count=1 22"
	User time (seconds): 4.68
	System time (seconds): 7.04
	Percent of CPU this job got: 96%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:12.14
	Maximum resident set size (kbytes): 68800
	Minor (reclaiming a frame) page faults: 6569001
````

`/usr/bin/time` with fix applied:

````
	Command being timed: "perf record -o perf.3 -g python3 ssh.py ssh 127.0.0.1 dd if=/dev/zero bs=20M count=1 22"
	User time (seconds): 0.49
	System time (seconds): 0.29
	Percent of CPU this job got: 65%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:01.20
	Maximum resident set size (kbytes): 69272
	Minor (reclaiming a frame) page faults: 21368
	Voluntary context switches: 5391
````

### DIFF
--- a/paramiko/file.py
+++ b/paramiko/file.py
@@ -192,9 +192,9 @@ class BufferedFile(ClosingContextManager):
             raise IOError("File is not open for reading")
         if (size is None) or (size < 0):
             # go for broke
-            result = self._rbuffer
+            result = [self._rbuffer]
+            self._pos += len(self._rbuffer)
             self._rbuffer = bytes()
-            self._pos += len(result)
             while True:
                 try:
                     new_data = self._read(self._DEFAULT_BUFSIZE)
@@ -202,10 +202,10 @@ class BufferedFile(ClosingContextManager):
                     new_data = None
                 if (new_data is None) or (len(new_data) == 0):
                     break
-                result += new_data
+                result.append(new_data)
                 self._realpos += len(new_data)
                 self._pos += len(new_data)
-            return result
+            return bytes().join(result)
         if size <= len(self._rbuffer):
             result = self._rbuffer[:size]
             self._rbuffer = self._rbuffer[size:]


lsh-0:
@bitprophet , can we get your attention on this one, please? This issue with paramiko has been preventing us from moving off of python2 for a long long time.

dw:
This patch is bad:

>              self._pos += len(result) 

Needs to be 

> self._pos += len(result[-1])

or similar. Will resent later in the week

lsh-0:
> Will resent later in the week

I appreciate your work

lsh-0:
bump. What is blocking this? I apologize for being so demanding but paramiko and it's terrible performance under python3 has been blocking us from moving off of python2 for years now. 

https://pythonclock.org/

jimmyharris:
My guess as to the block here is that this change isn’t Python27 compatible (which needs to be supported for a couple of months) additionally @dw mentioned something about needing to update the patch but I don’t see evidence of the change? That might be my inability to parse the GitHub mobile interface though.

lsh-0:
I've decided to purge Fabric and Paramiko from our dependencies instead.

https://github.com/elifesciences/threadbare

ploxiln:
The author amended his commit and force-pushed the changed commit. The fix he mentioned is applied in this pull request.

This change is compatible with python-2.7 (the test failures are unrelated - the invoke task tool is importing tasks and there's and error far down the chain in docutils - this branch probably just needs a rebase).

I happen to be maintaining my own fork of paramiko ... and I happen to have included this change in a recent release: https://github.com/ploxiln/paramiko-ng/releases

But I think this might not help much with the specific slowness you experience. For example, I don't think this helps with SFTPClient slowness, nor with the overhead/latency of running a series of exec_command()s. There are a few different causes of slowness, it is complicated :)

PR #1515, opened 2019-10-02
Support reading binary file-like objects using from_private_key

### LINKED ISSUE #1513: RSAKey.from_private_key doesn't work with io.BytesIO object in python3
when using io.BytesIO to create file like object and pass than object to RSAKey.from_private_key in python3 like this

```
ssh_key = '-----BEGIN DSA PRIVATE KEY-----\n'\
'<keydata>'\
'-----END DSA PRIVATE KEY-----\n'

file_obj = io.BytestIO(ssh_key)
pkey = RSAKey.from_private_key(file_obj)
```
it raises an error

the issue is here 
```
    def _read_private_key(self, tag, f, password=None):
        lines = f.readlines()
        start = 0
        beginning_of_key = "-----BEGIN " + tag + " PRIVATE KEY-----"
        while start < len(lines) and lines[start].strip() != beginning_of_key:
            start += 1
        if start >= len(lines):
            raise SSHException("not a valid " + tag + " private key file")
```
readline returns a list of binary and this comparison fails
```
lines[start].strip() != beginning_of_key
```

Overv:
This PR fixes #1513 by automatically decoding bytes when a binary file-like object is passed to `from_private_key`.

Example code:

```python
from paramiko import RSAKey
import io

with open('id_rsa', 'rb') as f:
    ssh_key = f.read()
    file_obj = io.BytesIO(ssh_key)
    pkey = RSAKey.from_private_key(file_obj)
```

### DIFF
--- a/paramiko/pkey.py
+++ b/paramiko/pkey.py
@@ -281,6 +281,11 @@ class PKey(object):
 
     def _read_private_key(self, tag, f, password=None):
         lines = f.readlines()
+
+        # Support binary file-like objects
+        if type(lines[0]) is bytes:
+            lines = [line.decode('ascii') for line in lines]
+
         start = 0
         beginning_of_key = "-----BEGIN " + tag + " PRIVATE KEY-----"
         while start < len(lines) and lines[start].strip() != beginning_of_key:


paramiko/pkey.py
```diff
@@ -281,6 +281,11 @@ def _read_private_key_file(self, tag, filename, password=None):
 
     def _read_private_key(self, tag, f, password=None):
         lines = f.readlines()
+
+        # Support binary file-like objects
+        if type(lines[0]) is bytes:
+            lines = [line.decode('ascii') for line in lines]
```
XVilka: Seems like flake8 is complaining:
```
--- paramiko/pkey.py	2019-10-02 19:40:00.632415 +0000
+++ paramiko/pkey.py	2019-10-02 19:40:32.048944 +0000
@@ -282,11 +282,11 @@
     def _read_private_key(self, tag, f, password=None):
         lines = f.readlines()
 
         # Support binary file-like objects
         if type(lines[0]) is bytes:
-            lines = [line.decode('ascii') for line in lines]
+            lines = [line.decode("ascii") for line in lines]

         start = 0
         beginning_of_key = "-----BEGIN " + tag + " PRIVATE KEY-----"
         while start < len(lines) and lines[start].strip() != beginning_of_key:
             start += 1
would reformat paramiko/pkey.py
```

Overv: Completely missed that, sorry! I've pushed a fix.

PR #1520, opened 2019-10-10
[RSA keys] Add support for rsa-sha2-256 and rsa-sha2-512

Terrtia:
RSA keys: Add support for rsa-sha2-256 and rsa-sha2-512

RFC: https://tools.ietf.org/rfc/rfc8332.txt

```
This memo updates RFCs 4252 and 4253 to define new public key
   algorithms for use of RSA keys with SHA-256 and SHA-512 for server
   and client authentication in SSH connections.

```

```
3.  New RSA Public Key Algorithms

   This memo adopts the style and conventions of [RFC4253] in specifying
   how use of a public key algorithm is indicated in SSH.

   The following new public key algorithms are defined:

     rsa-sha2-256        RECOMMENDED    sign    Raw RSA key
     rsa-sha2-512        OPTIONAL       sign    Raw RSA key

   These algorithms are suitable for use both in the SSH transport layer
   [RFC4253] for server authentication and in the authentication layer
   [RFC4252] for client authentication.

   Since RSA keys are not dependent on the choice of hash function, the
   new public key algorithms reuse the "ssh-rsa" public key format as
   defined in [RFC4253]:

   string    "ssh-rsa"
   mpint     e
   mpint     n

   All aspects of the "ssh-rsa" format are kept, including the encoded
   string "ssh-rsa".  This allows existing RSA keys to be used with the
   new public key algorithms, without requiring re-encoding or affecting
   already trusted key fingerprints.

   Signing and verifying using these algorithms is performed according
   to the RSASSA-PKCS1-v1_5 scheme in [RFC8017] using SHA-2 [SHS] as
   hash.

   For the algorithm "rsa-sha2-256", the hash used is SHA-256.
   For the algorithm "rsa-sha2-512", the hash used is SHA-512.
```

### DIFF
--- a/paramiko/rsakey.py
+++ b/paramiko/rsakey.py
@@ -123,7 +123,14 @@ class RSAKey(PKey):
         return m
 
     def verify_ssh_sig(self, data, msg):
-        if msg.get_text() != "ssh-rsa":
+        key_algorithm = msg.get_text()
+        if key_algorithm == "ssh-rsa":
+            algorithm = hashes.SHA1()
+        elif key_algorithm == "rsa-sha2-256":
+            algorithm = hashes.SHA256()
+        elif key_algorithm == "rsa-sha2-512":
+            algorithm = hashes.SHA512()
+        else:
             return False
         key = self.key
         if isinstance(key, rsa.RSAPrivateKey):
--- a/paramiko/rsakey.py
+++ b/paramiko/rsakey.py
@@ -138,7 +138,7 @@ class RSAKey(PKey):
 
         try:
             key.verify(
-                msg.get_binary(), data, padding.PKCS1v15(), hashes.SHA1()
+                msg.get_binary(), data, padding.PKCS1v15(), algorithm
             )
         except InvalidSignature:
             return False
--- a/paramiko/rsakey.py
+++ b/paramiko/rsakey.py
@@ -137,9 +137,7 @@ class RSAKey(PKey):
             key = key.public_key()
 
         try:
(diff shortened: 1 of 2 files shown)


tiran:
Thanks for the patch!

Paramiko is currently not compatible with systems with strict crypto policy. Fedora's FUTURE and FIPS [crypto policy](https://gitlab.com/redhat-crypto/fedora-crypto-policies/blob/master/policies) have RSA-SHA1 disabled. Only RSA with SHA256 (rsa-sha2-256) and SHA512 (rsa-sha2-512) are supported RSA pubkey auth algorithms.

Your patch is not complete and doesn't enable rsa-sha2-256 for me.

tiran:
- You need to modify the ``RSAKey`` class and replace all occurrences of ``ssh-rsa`` with ``self.get_name()``: ``key_type="ssh-rsa",``, `` cert_type="ssh-rsa-cert-v01@openssh.com"``, ``m.add_string("ssh-rsa")``.
- Then you need to figure out how to make RSAKey respond to the algorithm identifier. ``ECDSAKey`` gets the key type from ``msg.get_text()``.
- Add the new algorithms to ``Transport._preferred_keys``. I suggest before ``sha-rsa``
- Add ``"rsa-sha2-256-cert-v01@openssh.com"`` and ``"rsa-sha2-512-cert-v01@openssh.com"`` mappings to ``Transport._key_info``.

bitprophet:
Closing in favor of #1643 - thanks!

Issue #1475, opened 2019-07-09
SSH connection still alive on channel.close


yaroslavvb:
The following doesn't seem to close connection on the server side.
```
    chan: paramiko.Channel = transport.open_session()
    chan.exec_command(long_running_command)
    chan.close()
```

IE, repeating this will eventually end up with `paramiko.ssh_exception.ChannelException: (1, 'Administratively prohibited')` as the server rejects additional SSH connections.

Looking at debug log, the number of channels stays constant while number of active SSH connections grows

```
 <paramiko.Transport at 0x151d4320 (cipher aes128-ctr, 128 bits) (active; 2 open channel(s))>]
```

My workaround has been to manually kill the command in addition to channel close
```chan.exec_command('killall long_running_command')```

paramiko version '2.4.2'

**Comment by mhaoli** [comment]
I'm also facing this problem. 

I used paramiko to ssh from a linux device to a Windows device, and ran a long running command. After calling the `channel.close`, the chanel was still alive. I checked that the channel object exists in `ssh_client.get_transport()._channels._map` and the pid was still alive on Windows device. 

My paramiko version: 2.7.1

**Comment by azertyfun** [comment]
Just ran into this issue, but in my case it's because I had the following workaround for an unrelated issue:

```python
try:
    self.connection.connect(...)
except paramiko.ssh_exception.AuthenticationException:
    self.connection.connect(..., disabled_algorithms={'pubkeys': ['rsa-sha2-256', 'rsa-sha2-512']})
```

Turns out I had to close that connection even if it had failed:

```python
try:
    self.connection.connect(...)
except paramiko.ssh_exception.AuthenticationException:
    self.connection.close()
    self.connection.connect(..., disabled_algorithms={'pubkeys': ['rsa-sha2-256', 'rsa-sha2-512']})
```

PR #1535, opened 2019-10-29
Add send_signal method to Channel

MrMino:
Implemented as per [section 6.9 of RFC 4254](https://tools.ietf.org/html/rfc4254#section-6.9) ("The Secure Shell (SSH)
Connection Protocol"). Support for this request was added in OpenSSH
7.9. Support in other implementations may vary.


MrMino:
@bitprophet any update on this? As people are moving away from older OpenSSH versions now, it would be a very nice-to-have. The PR is relatively simple.

MrMino:
Note to self: should this ever be pursued further, exit code fetching would also have to be improved, see #823

greezybacon:
Any reason on why this cannot be merged? I ended implementing this myself. And, when I looked through the pull requests prior to creating my own, I found this one. Great work, @MrMino. 

The only thing I can think to add would be a translation between the Python `signal` library and the text signal name to send:
```python
import signal

class Channel(...):
   ...

    def send_signal(self, signal):
        if type(signal) is signal.Signals:
            signal = signal.name[3:]
```

bskinn:
This PR only includes the code change; if augmented with docs and tests, that would facilitate a potential merge.

MrMino:
@baskinn I thought the docs are generated via apidoc? I'll add some tests today / tomorrow.

greezybacon:
@MrMino I would be happy to help in any capacity. Please advise/delegate if you'd like a teammate on this.

bskinn:
@MrMino Oh, right, yeah -- the docstring should do it. Sorry!

travisdowns:
I still have my fingers crossed this will be merged one day!

bskinn:
Coming back to this after the mention in #1475.

Is this something that could easily have tests added?

Also, this would need an entry in `/sites/www/changelog.rst`.

MrMino:
@bskinn I'd like to revive this.

If anyone can point me to an existing example of how the tests for low-level protocol code like this should look like, I'd be happy to add them.
Other `Channel` methods such as `resize_pty()`, `invoke_subsystem()`, or `get_pty()` doesn't seem to be tested on this level either, and I'm afraid the required test infrastructure might not exist in the repo. My only idea for testing this is to check whether the channel sends the right bytes (i.e. some double-bookkeeping that ensures that the Right Thing™ isn't changed to the Wrong Thing™ by mistake).

If you are thinking about an end-to-end tests with a stub Paramiko SSH server, the server-side code would need to be implemented first, which I believe is a bit higher toll.

MrMino:
@greezybacon Integrating the `signal` module into this is would be a mistake.

The signals exposed in stdlib are platform dependent. By supporting these enums here, we would be inadvertently inviting misuse by code that only works when the platforms of the server and the client match.

MrMino:
This code is now happily running in a larger environment. I also tested it myself on one of OpenSSH versions from 2023.

bskinn:
@MrMino, I'll bump this up to bitprophet again; he may think that specific tests aren't needed here, in which case I would think this is close to merge-able.

Please add an initial draft of an entry to `sites/www/changelog.rst` now, I don't think that should need bitprophet's input.

Also, please expand the docstring of the new function -- at minimum, replicate (with suitable edits if appropriate) the comment in the up-top PR description, including the link to the RFC. Additionally, if there are any context-specific reasons for the specific calls to the `.add_` methods of `m`, it would be good to add comments explaining those also. (E.g., is each value added to the message for the purposes of satisfying paramiko's machinery, or for meeting the RFC spec, or for both?)

Thanks!

MrMino:
@bskinn done.

> Additionally, if there are any context-specific reasons for the specific calls to the .add_ methods of m, it would be good to add comments explaining those also.

There are none. It is taken straight from the RFC.

MrMino:
(force pushed again to fix a linter warning)

bskinn:
Sorry for the delay, @MrMino -- yes, I think CHANGELOG/docs are good now, I've pulled that label.

I'm going to leave the 'needs tests' on for now as a reminder. bitprophet has continued to be tight on OSS time lately, but I've got this PR on the list for him to consider the next time he has opportunity.

MrMino:
FTR test failures are unrelated, happened after a rebase.

The following MR adds handlers for `exit-signal` messages that come from the server if the process was killed by a signal: #2517

PR #1543, opened 2019-11-09
Remove any trace of pytest_relaxed from paramiko

kmosiejczuk:
paramiko turns off pytest_relaxed, but if it is installed, pytest will
use it, and it is not harmless if you are depending on normal pytest
behavior. Had various python porting projects of mine have all their
tests start exploding until I uninstalled pytest_relaxed.

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -33,7 +33,6 @@ import warnings
 import weakref
 from tempfile import mkstemp
 
-from pytest_relaxed import raises
 from mock import patch, Mock
 
 import paramiko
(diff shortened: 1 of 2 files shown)


tests/test_client.py
```diff
@@ -682,13 +681,6 @@ def test_password_kwarg_works_for_password_auth(self):
         # Straightforward / duplicate of earlier basic password test.
         self._test_connection(password="pygmalion")
 
-    # TODO: more granular exception pending #387; should be signaling "no auth
-    # methods available" because no key and no password
-    @raises(SSHException)
```
ploxiln: This can be accomplished in a reasonable way without `pytest_relaxed.raises`, look for examples in this file using `self.assertRaises()`

PR #1550, opened 2019-11-12
Use isinstance instead of issubclass to test for types.

atombrella:


tests/test_auth.py
```diff
@@ -172,7 +172,7 @@ def test_bad_password(self):
             self.assertTrue(False)
         except:
             etype, evalue, etb = sys.exc_info()
-            self.assertTrue(issubclass(etype, AuthenticationException))
+            self.assertIsInstance(etype, AuthenticationException)
```
ploxiln: the only problem with this pull request is that you also had to replace `etype` with `evalue`. or even better:

```python
except Exception as e:
    self.assertIsInstance(e, AuthenticationException)
```

PR #1555, opened 2019-11-19
dict_keys can't be indexed in python3

jleake-tripwire:


ploxiln:
`HostKeys.get()` returns a custom `SubDict` class which has an explicit `keys()` methods which always returns a list.

https://github.com/paramiko/paramiko/blob/2.6.0/paramiko/hostkeys.py#L182-L187

jleake-tripwire:
> `HostKeys.get()` returns a custom `SubDict` class which has an explicit `keys()` methods which always returns a list.
> 
> https://github.com/paramiko/paramiko/blob/2.6.0/paramiko/hostkeys.py#L182-L187

Aha, that'll do it. Thanks!

PR #1556, opened 2019-11-20
Use isinstance instead of issubclass to test for types.

atombrella:
Update to #1550 Tests are now passing.


tests/test_auth.py
```diff
@@ -266,7 +264,7 @@ def test_auth_non_responsive(self):
         self.tc.connect()
         try:
             self.tc.auth_password("unresponsive-server", "hello")
-        except:
-            etype, evalue, etb = sys.exc_info()
-            self.assertTrue(issubclass(etype, AuthenticationException))
+        except Exception as e:
+            _, evalue, _ = sys.exc_info()
+            self.assertIsInstance(e, AuthenticationException)
             self.assertTrue("Authentication timeout" in str(evalue))
```
intgr: While you're touching this, you might also convert this to use `with self.assertRaises(AuthenticationException)` instead.

PR #1558, opened 2019-11-22
Add functions to auto-detect and load any private key type

intgr:
Previously there was no way to auto-load any key type. Many downstream
users re-implemented this logic themselves, which could easily become
outdated.

It's also likely that they will implement it badly, like this: https://github.com/tokoyami/pySCP/blob/c4c3ac7ec39965e0fe4274e8e43428a9210988c7/connector.py#L13-L29 :)


intgr:
Fixed line length and code style issues using `black`. Sorry, didn't notice those errors earlier.

paramiko/pkey.py
```diff
@@ -31,11 +32,87 @@
...
+        key_str,
+        re.MULTILINE,
+    )
+    if match is None:
+        raise SSHException("private key file not supported or invalid")
+
+    keytype = match.group(1)
+
+    if keytype == "OPENSSH":
+        from paramiko import Ed25519Key
```
ploxiln: this could be any type of key ... it just happens that paramiko does not implement support for the new openssh private key format for any other key type (until #1343 is merged)

intgr: Ah thanks for pointing that out. But seems like there is nothing I can do about that right now? PR #3143 does not seem to be in a hurry. :(
Should I add a code comment here?

This PR seems to me like a step in the right direction at least, users shouldn't care about key types (unless they specifically want to). And having to specify key type in advance seems awkward for supporting the new-type keys too.

ploxiln: yeah, for now a code comment is probably the best that can be done

intgr:
@ploxiln Is this PR something you would be interested in merging into `paramiko-ng`?

ploxiln:
Yes ... but my fork has new-openssh-format support, and that new format has the key type string sort of in the middle. I may want more refactoring in order to avoid copying a significant chunk of the new-format parsing code.

(There is also a bit of round-about-ness converting between a file-object and a list of lines, which I might prefer some refactoring of... for example, `load_private_key()` wraps the `key_str` into a `StringIO`, and then `load_private_key_from_file()` calls `.read()` to get the whole thing as a `str` in memory again for type detection, and then the first thing `_read_private_key()` does is `.readlines()` which gets another full read in memory but into a list of line strings ... seems like the the list-of-lines-in-memory should be the native/internal type passed around and used directly in most places. Or something like that.)

intgr:
It seems that for this feature to work without doing redundant work (doing multiple passes over the file content), it's necessary to split out significant parts of logic from the `PKey` class into stand-alone functions.

While reading the code, I didn't much like the ping-ponging of logic between `PKey` base class and its subclasses, so I think that's an improvement in general. Just making sure that you agree that's the way to go?

ploxiln:
The helper methods on the PKey class are very similar to just having helper functions in pkey.py ... but they also organize and reference constants like `_CIPHER_TABLE` on `self` ... so to reduce the amount of churn, I think you could continue to use pretty-much-static methods. I don't think a _big_ refactor is needed, just the bits at the beginnings of the functions, but the details are already different in the other fork so maybe the discussion should be there :)

intgr:
I rebased this on paramiko master; merging #1343 invalidated the old patch.

The new work is mostly based on https://github.com/ploxiln/paramiko-ng/pull/57 but still has some rough edges. In particular, #1566 and #1567 should be merged before this.

bitprophet:
Thanks for this, I like the overall idea here, though I haven't yet read the entire patch in depth. Basic thoughts:

- I've become a fan of classmethod alt constructors, so we might want to make the module level functions into those (alongside the newly classmethod-ized subroutines). Eg `PKey.autoload()`.
  -  and maybe combine the 'string vs filepath' versions into a single call w/ kwargs, though I'm less sure about that as it's a tradeoff (more complexity in arg handling).
  - but I've never been a fan of the original codebase's naming here (why is one method implicit about what type it takes, and why is the other ambiguous about file paths vs FLOs?) so I figure we should take this opportunity to not perpetuate it.
- I'd like to make it clear to readers of the API docs (and changelog) that the autoload does incur the cost of parsing the key twice (even though it does not load from disk twice).
  - The majority of users won't care but there will always be some that do, especially with this particular library...
- Might be nice to have some basic design-level docs about the key classes and this autoloading, probably as prose at the top of the API doc stub that currently just collates the key class' module autodocs.
  - This is something I can definitely do as I'm super picky about prose 😩 
  - But this can probably wait for #387 so we'll see.

intgr:
My initial thoughts:

> I've become a fan of classmethod alt constructors, so we might want to make the module level functions into those (alongside the newly classmethod-ized subroutines). Eg PKey.autoload().

The reason I didn't want to do it like that is because that would create a circular dependency between `PKey` and each of its subclasses. That just seems like a bad design.

I'm open to renaming this API but I don't particularly like the name `autoload`, since that puts emphasis on the "auto". I think this new API should be the one and only right way of loading private keys going forward. There is little reason for a user to instantiate a key class directly. Even if they want to apply some policy (e.g. only allow RSA and ECDSA keys), that can be just as nicely done *after* the key is loaded.

(I also dislike how paramiko mixes private and public keys in one class. Lumping more stuff into one class only makes that worse. Although the breakage from "fixing" that is probably not worth it)

> the autoload does incur the cost of parsing the key twice (even though it does not load from disk twice).

It shouldn't. The `load_private_key` function does the parsing and then passes the parsed data to the correct class via the newly-introduced `_raw=` argument. Can you point out anything that's done twice?

intgr:
I created a separate PR #1571 to unify duplicated logic in Ed25519Key, that refactor doesn't really belong in this PR.

Apreche:
I see this hasn't been worked on in a few months. Wondering what can be done to move it along. Right now I'm going to have to ask users of our application to manually input the type of key they are using. Not an ideal solution.

intgr:
This PR depends on the refactoring in #1571 -- getting that merged would be the first step. @bitprophet Have you had a chance to look at it?

intgr:
@Apreche I was recently made aware of this post by bitprophet (maintainer of Paramiko): http://bitprophet.org/blog/2020/07/02/help-wanted/

So maybe if you or someone else steps up to review #1571 and this PR, it can be sped up. Although few people are familiar with the project's coding style, I'm not sure how helpful code review by independent developers could be.

Apreche:
@intgr I'd love to help with this, but I'm not sure how much help I can be. The PRs look fine to me from a Python perspective, but I have no familiarity with the projects themselves or their architecture to make any further judgements.

PR #1559, opened 2019-11-26
Refactor discovery of SSH private keys

intgr:
Previously, scanning of private keys was done even if `look_for_keys=False` was used.

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -706,30 +706,8 @@ class SSHClient(ClosingContextManager):
                 except SSHException as e:
                     saved_exception = e
 
-        if not two_factor:
-            keyfiles = []
-
-            for keytype, name in [
-                (RSAKey, "rsa"),
-                (DSSKey, "dsa"),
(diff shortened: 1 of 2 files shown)


ploxiln:
see #1243 :)

intgr:
Doh! :)

Well certainly in hindsight, it seems silly that a PR that was ready, has been blocked for over 1 year, because it "may clash" with something that has not materialized (if I understand right, has been in the works since 2014?).

Sorry to be blunt, and I understand that it wasn't your decision personally, but I think this is a sign of a bad upstream. In my opinion, changes should be merged on a basis of "is this a step in the right direction". The justification "won't merge this small improvement because we will re-implement it perfectly in the future" is a bad one for many reasons.

If a project is responsive to first-time contributors, then a few of them will return to make other improvements. If their MRs languish, they most likely won't. And looking at the codebase, there are plenty of things that need improvement.

ploxiln:
generally agreed :)
see also https://github.com/ploxiln/paramiko-ng/releases

Issue #1584, opened 2019-12-29
EOF sent for already closed channel


bentsi:
For some reason (probably related to #1115) paramiko sends EOF for an already closed Channel, causing the Transport to close the whole SSH session.

paramiko log
```
2019-12-26 23:10:06,019 f:channel.py   p:DEBUG > [chan 8129] Max packet in: 32768 bytes
2019-12-26 23:10:07,014 f:channel.py   p:DEBUG > [chan 8129] Max packet out: 32768 bytes
2019-12-26 23:10:08,050 f:channel.py   p:DEBUG > [chan 8129] Sesch channel 8129 request ok
2019-12-26 23:10:08,085 f:channel.py   p:DEBUG > [chan 8129] EOF received (8129)
2019-12-26 23:10:08,087 f:channel.py   p:DEBUG > [chan 8129] EOF sent (8129)
2019-12-26 23:10:08,692 f:transport.py p:INFO  > Disconnect (code 2): Received ieof for nonexistent channel 11.
```

SSHD log
```
Dec 26 23:10:06 ip-10-0-109-19 sshd[1936]: debug1: server_input_channel_open: ctype session rchan 8129 win 2097152 max 32768
Dec 26 23:10:06 ip-10-0-109-19 sshd[1936]: debug1: input_session_request
Dec 26 23:10:06 ip-10-0-109-19 sshd[1936]: debug1: channel 11: new [server-session]
Dec 26 23:10:06 ip-10-0-109-19 sshd[1936]: debug1: session_new: session 11
Dec 26 23:10:06 ip-10-0-109-19 sshd[1936]: debug1: session_open: channel 11
Dec 26 23:10:06 ip-10-0-109-19 sshd[1936]: debug1: session_open: session 11: link with channel 11
Dec 26 23:10:06 ip-10-0-109-19 sshd[1936]: debug1: server_input_channel_open: confirm session
Dec 26 23:10:07 ip-10-0-109-19 sshd[1936]: debug1: server_input_channel_req: channel 11 request exec reply 1
Dec 26 23:10:07 ip-10-0-109-19 sshd[1936]: debug1: session_by_channel: session 11 channel 11
Dec 26 23:10:07 ip-10-0-109-19 sshd[1936]: debug1: session_input_channel_req: session 11 req exec
Dec 26 23:10:07 ip-10-0-109-19 sshd[1936]: Starting session: command for centos from 10.0.226.244 port 51038 id 11
[...]

PR #1585, opened 2019-12-30
EOFError Race condition when closing a transport thread

hybridczar:
Adding exception handling for packetizer race condition when the server ends the session before the client. This can occur when utilizing fabric2 2.5.0 and sends a high volume of commands through a Connection() object, then trying to close the object.

The EOFError is being raised when trying to write out the final bytes to close the socket, but the socket is already closed. This does not prevent the rest of the transport object shutdown properly, so a warning log is written out instead of raising the exception higher.

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -2151,7 +2151,10 @@ class Transport(threading.Thread, ClosingContextManager):
                 chan._unlink()
             if self.active:
                 self.active = False
-                self.packetizer.close()
+                try:
+                    self.packetizer.close()
+                except EOFError:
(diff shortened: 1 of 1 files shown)


hybridczar:
The tests seem to pass, but are failing to report they passed for some reason.

fruch:
sounds related to #1584 and #1115 
@bitprophet would love to hear your thoughts on how to approach fixing those issues ?

ploxiln:
Those other issues seem to be about closing one channel of the transport, while this is about closing the whole transport and all channels.

poleguy:
It's failing flake8. Same thing just happened to me. It seems like flake8 is enforcing formatting rules.

PR #1592, opened 2020-01-08
Remove redundant ed25519 extras_require

felixonmars:
Both of the libs are already listed as install_requires so there is no need to write them in extras_require again.


ploxiln:
This looks intentional, see below in the file:

```python
    # TODO 3.0: remove bcrypt, pynacl and update installation docs noting that
    # use of the extras_require(s) is now required for those
    install_requires=["bcrypt>=3.1.3", "cryptography>=2.5", "pynacl>=1.0.1"],
```

felixonmars:
Indeed... Sorry for overlooking it.

PR #1615, opened 2020-02-04
fixing starting point of key for PKey._read_private_key_pem

htol:
Hi!
When file key or variable with key not started directly from beginning of file we will get base64 padding error because we will not strip begging of key. Here is an example:
Context which comes to 
`[u'\n',
 u'-----BEGIN RSA PRIVATE KEY-----\n', 
 u'<skipped key lines>\n',
 u'-----END RSA PRIVATE KEY-----\n']`

As you can see there is empty line at beginning. Without fix there will be an error about base64 padding and data which was provided to data = decodebytes(b("".join(lines[start:end]))) at _read_private_key_pem() will be lokked like this.
`[u'-----BEGIN RSA PRIVATE KEY-----\n', 
 u'<skipped key lines>\n',]`
As you can see there is obviously an error.

### DIFF
--- a/paramiko/pkey.py
+++ b/paramiko/pkey.py
@@ -331,7 +331,7 @@ class PKey(object):
             m = self.END_TAG.match(lines[end])
 
         if keytype == tag:
-            data = self._read_private_key_pem(lines, end, password)
+            data = self._read_private_key_pem(lines[start-1:], end, password)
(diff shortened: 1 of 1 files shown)


ploxiln:
see also #1604 #1601 #1580

PR #1618, opened 2020-02-05
SSHClient.connect: support two factor callback

g-braeunlich:
This commit adds a new argument `two_factor_callback` to SSHClient.connect.
This way, it is possible to call this function in a gui, when you do not know in advance (cannot use the password argument), if there a 2nd factor is required and you want to get the users input from something different than the terminal.

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -237,6 +237,7 @@ class SSHClient(ClosingContextManager):
         gss_trust_dns=True,
         passphrase=None,
         disabled_algorithms=None,
+        two_factor_callback=None,
     ):
         """
         Connect to an SSH server and authenticate to it.  The server's host key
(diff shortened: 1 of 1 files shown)


paramiko/client.py
```diff
@@ -383,7 +388,7 @@ def connect(
         )
         if self._log_channel is not None:
             t.set_log_channel(self._log_channel)
-        if banner_timeout is not None:
+            if banner_timeout is not None:
```
ploxiln: bad indent here

g-braeunlich: thx for the careful review!
Should be fixed now

paramiko/client.py
```diff
@@ -314,6 +315,10 @@ def connect(
         :param dict disabled_algorithms:
             an optional dict passed directly to `.Transport` and its keyword
             argument of the same name.
+        :param two_factor_callback:
+            an optional callable with no arguments. If specified, it will be
+            called when a 2nd factor is needed. Usefull when calling connect
+            from a gui which needs to open a dialog
```
allanlewis: A few minor things here:

- `s/Usefull/Useful/`
- Please surround "connect" with backticks as it's a code object.
- `s/gui/GUI/`
- Please append a full stop/period.

g-braeunlich: Thx. Done

g-braeunlich:
Anything else to improve here?

g-braeunlich:
anything I can do to get this merged?

PR #1628, opened 2020-02-25
Local address and local port bind support for connections.

peuris:
Support for binding the IP address and port to predefined values. Required in NAT environments where connections are expected from certain addresses and ports.

:param local_bind_ip:
    Set the default local interface for the connection. (defaults to 0.0.0.0)
:param local_bind_port:
    Set the default local port of the interface. (defaults to 0)

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -305,6 +305,8 @@ class Transport(threading.Thread, ClosingContextManager):
     def __init__(
         self,
         sock,
+        local_bind_ip = None,
+        local_bind_port = None,
         default_window_size=DEFAULT_WINDOW_SIZE,
         default_max_packet_size=DEFAULT_MAX_PACKET_SIZE,
         gss_kex=False,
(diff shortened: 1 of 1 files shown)


ploxiln:
This is not strictly needed, because the caller can make the tcp connection itself in whatever way is needed, and pass the connected `sock` to the `Transport`. The option of passing a tuple or string for sock is just for convenience.

peuris:
When using paramiko.SFTPClient in the simplest way, it would be pythonic(**beautiful**) approach not to import socket and use the binding methods only for binding the connection to an interface (ip, port), at least when the other basic socket connection parameters are provided for Transport class in a tuple.


`import paramiko

# use key
pkey = paramiko.RSAKey.from_private_key_file("mykey")

# create transport
transport = paramiko.SSHClient()
transport.set_missing_host_key_policy(paramiko.AutoAddPolicy())
transport = paramiko.Transport(("127.0.0.1", "22")) # add bind
transport.connect(username = "usr", pkey = pkey)

# connect
sftp = paramiko.SFTPClient.from_transport(transport)

# close
sftp.close()
transport.close()`

paramiko/transport.py
```diff
@@ -405,6 +412,20 @@ def __init__(
...
                     # addr = sockaddr
                     sock = socket.socket(af, socket.SOCK_STREAM)
+                    # bind local socket to custom ip and port
+                    if not local_bind_ip or type(local_bind_ip) is not str:
+                        local_bind_ip = "0.0.0.0"
+                    if not local_bind_port:
+                        local_bind_port = 0
+                    if type(local_bind_port) is str:
+                        try:
+                            local_bind_port = int(local_bind_port)
```
pyhedgehog, line 422: Shouldn't we call `socket.getservbyname()` if its not decimals?

PR #1643, opened 2020-03-13
Add rsa-sha2-256 and rsa-sha2-512 algorithms

krisztian-kovacs:
These are specified in RFC 8332 (https://tools.ietf.org/html/rfc8332) and proposed by recent OpenSSH versions as a drop-in replacement for the deprecated ssh-rsa algorithm.
The advantage is that the same RSA keys can be used without relying on the SHA-1 digest now considered insecure.

### DIFF
--- a/paramiko/rsakey.py
+++ b/paramiko/rsakey.py
@@ -221,3 +221,12 @@ class RSASHA256Key(RSAKey):
 
     _signature_digest_algorithm = hashes.SHA256
     _signature_algorithm_name = "rsa-sha2-256"
+
+
+class RSASHA512Key(RSAKey):
+    """
+    A special RSAKey that uses SHA-256 digest for sign/verify.
+    """
+
+    _signature_digest_algorithm = hashes.SHA512
+    _signature_algorithm_name = "rsa-sha2-512"
--- a/paramiko/rsakey.py
+++ b/paramiko/rsakey.py
@@ -37,6 +37,9 @@ class RSAKey(PKey):
     data.
     """
 
+    _signature_digest_algorithm = hashes.SHA1
+    _signature_algorithm_name = "ssh-rsa"
+
     def __init__(
         self,
         msg=None,
@@ -104,7 +107,7 @@ class RSAKey(PKey):
         )
 
     def get_name(self):
-        return "ssh-rsa"
+        return self._signature_algorithm_name
 
     def get_bits(self):
         return self.size
@@ -114,16 +117,18 @@ class RSAKey(PKey):
 
     def sign_ssh_data(self, data):
         sig = self.key.sign(
-            data, padding=padding.PKCS1v15(), algorithm=hashes.SHA1()
+            data,
+            padding=padding.PKCS1v15(),
+            algorithm=self._signature_digest_algorithm(),
         )
 
         m = Message()
-        m.add_string("ssh-rsa")
+        m.add_string(self._signature_algorithm_name)
         m.add_string(sig)
         return m
 
     def verify_ssh_sig(self, data, msg):
-        if msg.get_text() != "ssh-rsa":
+        if msg.get_text() != self._signature_algorithm_name:
             return False
         key = self.key
         if isinstance(key, rsa.RSAPrivateKey):
@@ -131,7 +136,10 @@ class RSAKey(PKey):
 
         try:
             key.verify(
-                msg.get_binary(), data, padding.PKCS1v15(), hashes.SHA1()
+                msg.get_binary(),
+                data,
+                padding.PKCS1v15(),
+                self._signature_digest_algorithm(),
             )
         except InvalidSignature:
             return False
@@ -204,3 +212,12 @@ class RSAKey(PKey):
             self._got_bad_key_format_id(pkformat)
         assert isinstance(key, rsa.RSAPrivateKey)
         self.key = key
+
+
+class RSASHA256Key(RSAKey):
+    """
+    A special RSAKey that uses SHA-256 digest for sign/verify.
+    """
+
+    _signature_digest_algorithm = hashes.SHA256
+    _signature_algorithm_name = "rsa-sha2-256"
--- a/paramiko/__init__.py
+++ b/paramiko/__init__.py
@@ -47,7 +47,7 @@ from paramiko.ssh_exception import (
     SSHException,
 )
 from paramiko.server import ServerInterface, SubsystemHandler, InteractiveQuery
-from paramiko.rsakey import RSAKey, RSASHA256Key
+from paramiko.rsakey import RSAKey, RSASHA256Key, RSASHA512Key
 from paramiko.dsskey import DSSKey
 from paramiko.ecdsakey import ECDSAKey
 from paramiko.ed25519key import Ed25519Key
--- a/paramiko/__init__.py
+++ b/paramiko/__init__.py
@@ -47,7 +47,7 @@ from paramiko.ssh_exception import (
     SSHException,
 )
 from paramiko.server import ServerInterface, SubsystemHandler, InteractiveQuery
-from paramiko.rsakey import RSAKey
+from paramiko.rsakey import RSAKey, RSASHA256Key
 from paramiko.dsskey import DSSKey
 from paramiko.ecdsakey import ECDSAKey
 from paramiko.ed25519key import Ed25519Key
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -99,7 +99,7 @@ from paramiko.message import Message
 from paramiko.packet import Packetizer, NeedRekeyException
 from paramiko.primes import ModulusPack
 from paramiko.py3compat import string_types, long, byte_ord, b, input, PY2
-from paramiko.rsakey import RSAKey
+from paramiko.rsakey import RSAKey, RSASHA256Key
 from paramiko.ecdsakey import ECDSAKey
 from paramiko.server import ServerInterface
 from paramiko.sftp_client import SFTPClient
@@ -173,6 +173,7 @@ class Transport(threading.Thread, ClosingContextManager):
         "ecdsa-sha2-nistp256",
         "ecdsa-sha2-nistp384",
         "ecdsa-sha2-nistp521",
+        "rsa-sha2-256",
         "ssh-rsa",
         "ssh-dss",
     )
@@ -261,6 +262,8 @@ class Transport(threading.Thread, ClosingContextManager):
     _key_info = {
         "ssh-rsa": RSAKey,
         "ssh-rsa-cert-v01@openssh.com": RSAKey,
+        "rsa-sha2-256": RSASHA256Key,
+        "rsa-sha2-256-cert-v01@openssh.com": RSASHA256Key,
         "ssh-dss": DSSKey,
         "ssh-dss-cert-v01@openssh.com": DSSKey,
         "ecdsa-sha2-nistp256": ECDSAKey,
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -99,7 +99,7 @@ from paramiko.message import Message
 from paramiko.packet import Packetizer, NeedRekeyException
 from paramiko.primes import ModulusPack
 from paramiko.py3compat import string_types, long, byte_ord, b, input, PY2
-from paramiko.rsakey import RSAKey, RSASHA256Key
+from paramiko.rsakey import RSAKey, RSASHA256Key, RSASHA512Key
 from paramiko.ecdsakey import ECDSAKey
 from paramiko.server import ServerInterface
 from paramiko.sftp_client import SFTPClient
@@ -174,6 +174,7 @@ class Transport(threading.Thread, ClosingContextManager):
         "ecdsa-sha2-nistp384",
         "ecdsa-sha2-nistp521",
         "rsa-sha2-256",
+        "rsa-sha2-512",
         "ssh-rsa",
         "ssh-dss",
     )
@@ -264,6 +265,8 @@ class Transport(threading.Thread, ClosingContextManager):
         "ssh-rsa-cert-v01@openssh.com": RSAKey,
         "rsa-sha2-256": RSASHA256Key,
         "rsa-sha2-256-cert-v01@openssh.com": RSASHA256Key,
+        "rsa-sha2-512": RSASHA512Key,
+        "rsa-sha2-512-cert-v01@openssh.com": RSASHA512Key,
         "ssh-dss": DSSKey,
         "ssh-dss-cert-v01@openssh.com": DSSKey,
         "ecdsa-sha2-nistp256": ECDSAKey,
--- a/tests/test_pkey.py
+++ b/tests/test_pkey.py
@@ -29,6 +29,7 @@ from hashlib import md5
 from paramiko import (
     RSAKey,
     RSASHA256Key,
+    RSASHA512Key,
     DSSKey,
     ECDSAKey,
     Ed25519Key,
@@ -57,6 +58,7 @@ FINGER_ECDSA_384 = "384 c1:8d:a0:59:09:47:41:8e:a8:a6:07:01:29:23:b4:65"
 FINGER_ECDSA_521 = "521 44:58:22:52:12:33:16:0e:ce:0e:be:2c:7c:7e:cc:1e"
 SIGNED_RSA = "20:d7:8a:31:21:cb:f7:92:12:f2:a4:89:37:f5:78:af:e6:16:b6:25:b9:97:3d:a2:cd:5f:ca:20:21:73:4c:ad:34:73:8f:20:77:28:e2:94:15:08:d8:91:40:7a:85:83:bf:18:37:95:dc:54:1a:9b:88:29:6c:73:ca:38:b4:04:f1:56:b9:f2:42:9d:52:1b:29:29:b4:4f:fd:c9:2d:af:47:d2:40:76:30:f3:63:45:0c:d9:1d:43:86:0f:1c:70:e2:93:12:34:f3:ac:c5:0a:2f:14:50:66:59:f1:88:ee:c1:4a:e9:d1:9c:4e:46:f0:0e:47:6f:38:74:f1:44:a8"  # noqa
 SIGNED_RSA_SHA2_256 = "cc:06:60:e0:00:2c:ac:9e:26:bc:d5:68:64:3f:9f:a7:e5:aa:41:eb:88:4a:25:05:9c:93:84:66:ef:ef:60:f4:34:fb:f4:c8:3d:55:33:6a:77:bd:b2:ee:83:0f:71:27:41:7e:f5:07:05:00:a9:4c:07:80:6f:be:76:67:cb:58:35:b9:2b:f3:c2:d3:3c:ee:e1:3f:59:e0:fa:e4:5c:92:ed:ae:74:de:0d:d6:27:16:8f:84:a3:86:68:0c:94:90:7d:6e:cc:81:12:d8:b6:ad:aa:31:a8:13:3d:63:81:3e:bb:05:b6:38:4d:02:0d:1b:5b:70:de:83:cc:3a:cb:31"  # noqa
+SIGNED_RSA_SHA2_512 = "87:46:8b:75:92:33:78:a0:22:35:32:39:23:c6:ab:e1:06:92:ad:bc:7f:6e:ab:19:32:e4:78:b2:2c:8f:1d:0c:65:da:fc:a5:07:ca:b6:55:55:31:83:b1:a0:af:d1:95:c5:2e:af:56:ba:f5:41:64:0f:39:9d:af:82:43:22:8f:90:52:9d:89:e7:45:97:df:f3:f2:bc:7b:3a:db:89:0e:34:fd:18:62:25:1b:ef:77:aa:c6:6c:99:36:3a:84:d6:9c:2a:34:8c:7f:f4:bb:c9:a5:9a:6c:11:f2:cf:da:51:5e:1e:7f:90:27:34:de:b2:f3:15:4f:db:47:32:6b:a7"  # noqa
 FINGER_RSA_2K_OPENSSH = "2048 68:d1:72:01:bf:c0:0c:66:97:78:df:ce:75:74:46:d6"
 FINGER_DSS_1K_OPENSSH = "1024 cf:1d:eb:d7:61:d3:12:94:c6:c0:c6:54:35:35:b0:82"
 FINGER_EC_384_OPENSSH = "384 72:14:df:c1:9a:c3:e6:0e:11:29:d6:32:18:7b:ea:9b"
@@ -244,6 +246,21 @@ class KeyTest(unittest.TestCase):
         self.assertEqual(sig, msg.get_binary())
         msg.rewind()
         pub = RSASHA256Key(data=key.asbytes())
+        self.assertTrue(pub.verify_ssh_sig(b"ice weasels", msg))
+
+    def test_sign_rsa_sha2_512(self):
+        # verify that the rsa-sha2-512 private key can sign and verify
+        key = RSASHA512Key.from_private_key_file(_support("test_rsa.key"))
+        msg = key.sign_ssh_data(b"ice weasels")
+        self.assertTrue(type(msg) is Message)
+        msg.rewind()
+        self.assertEqual("rsa-sha2-512", msg.get_text())
+        sig = bytes().join(
+            [byte_chr(int(x, 16)) for x in SIGNED_RSA_SHA2_512.split(":")]
+        )
+        self.assertEqual(sig, msg.get_binary())
+        msg.rewind()
+        pub = RSASHA512Key(data=key.asbytes())
         self.assertTrue(pub.verify_ssh_sig(b"ice weasels", msg))
 
     def test_sign_dss(self):
--- a/tests/test_pkey.py
+++ b/tests/test_pkey.py
@@ -26,7 +26,15 @@ import os
 from binascii import hexlify
 from hashlib import md5
 
-from paramiko import RSAKey, DSSKey, ECDSAKey, Ed25519Key, Message, util
+from paramiko import (
+    RSAKey,
+    RSASHA256Key,
+    DSSKey,
+    ECDSAKey,
+    Ed25519Key,
+    Message,
+    util,
+)
 from paramiko.py3compat import StringIO, byte_chr, b, bytes, PY2
 
 from .util import _support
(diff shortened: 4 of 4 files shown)


ploxiln:
Cool. This looks like a more complete alternative to https://github.com/paramiko/paramiko/pull/1520

bitprophet:
Dropped in rapidly growing p0 milestone. I need to put some attention into the other two legs of my big project triad, but when I circle back here this will get looked at. Thanks!

ChristopherRabotin:
Hey there. Is there anything I can do to help get this PR reviewed & merged and released in the next version of paramiko?

bitprophet:
> Hey there. Is there anything I can do to help get this PR reviewed & merged and released in the next version of paramiko?

As long as it still merges cleanly, nope, it's just waiting on my time to roll through the milestone and merge stuff for release (as noted). Thanks for your patience!

archoversight:
This is going to be key to continue functioning against RHEL 8 systems in FIPS mode where `ssh-rsa` is disabled completely.

ChristopherRabotin:
Have you tested this code for authentication against a server? I checked out your branch and tried connecting using a `3072 SHA256` key, and I still have "authentication failure." I'm sure the password provided to the key is correct because if it isn't, then the `RSAKey` object cannot be created (bit mismatch error or something).

Any idea what could be happening?

Thanks

Screenshot_2020-11-16 Kubernetes Dashboard

bitprophet:
> **Edit:** Let me add that the remote server I'm trying to connect only supports SFTP connections and does not allow for pure SSH connections. When trying to connect in that way, the server will close the connection after saying that only SFTP is supported. Could this confuse paramiko into thinking that the auth failed?

Off the top of my head, this may be plausible, in which case you may find yourself pending on the auth reorganization listed under #387 (though I'd also consider a more targeted PR if it turns out that your analysis is correct, and if there's a reasonably simple way to fix it w/o upending regular ssh behavior. Big if, sadly.)

krisztian-kovacs:
@ChristopherRabotin Yes, this works for us with OpenSSH. (And do test our own third party implementation of rsa-sha2-256 and rsa-sha2-512 with this.) Do you have logs from the server that could maybe in diagnosing the issue?

ChristopherRabotin:
Okay. I don't have access to the server, it's a third party and there is no
way they'll provide me any logs.

Christopher Rabotin.

krisztian-kovacs:
In this case maybe you could get the logs from paramiko itself. You can enable logging to a file like this:

```
paramiko.util.log_to_file("demo_server.log")
```

(From https://github.com/paramiko/paramiko/blob/master/demos/demo_simple.py#L39)

Feel free to remove any potential sensitive content, the SSH message sequence is what's interesting to help debug the issue.

Dith3r:
There is also list in client.Client._auth where it test filename and select only one type of RSAKey (ssh-rsa).

kuwv:
Isn't this a duplicate of: https://github.com/paramiko/paramiko/pull/1103

zdlhappy:
When to support rsa-sha2-512 and rsa-sha2-256? In progressing?

paramiko/rsakey.py
```diff
@@ -204,3 +212,21 @@ def _decode_key(self, data):
...
+    A special RSAKey that uses SHA-256 digest for sign/verify.
+    """
+
+    _signature_digest_algorithm = hashes.SHA256
+    _signature_algorithm_name = "rsa-sha2-256"
+
+
+class RSASHA512Key(RSAKey):
+    """
+    A special RSAKey that uses SHA-256 digest for sign/verify.
```
quarckster, line 228: Is it a typo here? Should it be `SHA-512`?

zdlhappy: Yes, I think so.

Alphadelta14:
> Have you tested this code for authentication against a server? I checked out your branch and tried connecting using a `3072 SHA256` key, and I still have "authentication failure." I'm sure the password provided to the key is correct because if it isn't, then the `RSAKey` object cannot be created (bit mismatch error or something).
> 
> Any idea what could be happening?
> 
> **Edit:** Let me add that the remote server I'm trying to connect only supports SFTP connections and does not allow for pure SSH connections. When trying to connect in that way, the server will close the connection after saying that only SFTP is supported. Could this confuse paramiko into thinking that the auth failed?
> 
> Thanks
> 
> Screenshot_2020-11-16 Kubernetes Dashboard

I ran into the same sort of issue when tracking this issue down. The auth banner isn't particularly related. I did find in the server logs a `userauth_pubkey: key type ssh-rsa not in PubkeyAcceptedKeyTypes [preauth]` rejection message that corresponded to it.
Wish the logs were better about that, but I guess we shouldn't expect servers to provide sensitive information about why they were rejected.
Switching `pkey=RSAKey()` to `pkey=RSA512Key()` from this PR fixed it for me though.

Alphadelta14:
> I ran into the same sort of issue when tracking this issue down. The auth banner isn't particularly related. I did find in the server logs a userauth_pubkey: key type ssh-rsa not in PubkeyAcceptedKeyTypes [preauth] rejection message that corresponded to it.
> Wish the logs were better about that, but I guess we shouldn't expect servers to provide sensitive information about why they were rejected.
Actually, it might be a good idea to check the key.get_name() in Transport.auth_publickey() to make sure that it inside of server_key_algo_list?

Dreamsorcerer:
Any movement on this? ssh-rsa has now actually been deprecated and is getting removed from openssh. This PR is starting to become a notable security issue. Would be great to ensure this is merged and released in time for inclusion into the Ubuntu 22.04 archives.

sktomer:
Hello... we really need this patch so that paramiko plays nice with the latest default crypto policies in fedora34 or later.
Is there a blocker for this patch or can it be qualified and merged asap ?

Dreamsorcerer:
> @Dreamsorcerer, what is the deadline for inclusion in Ubuntu 22.04?

I suspect the deadline is around February, but they probably just pull the package from Debian, so a Debian maintainer will need to package and release it before then. So, I'd say before the end of this year to be safe.

sktomer:
> @sktomer, is there any specific deadline associated with fedora crypto compat? Or just desirable ASAP because desirable (and secure)?

I'm from AWS (Amazon). We have a tooling infrastructure based on Python which uses Paramiko to ssh into EC2 instances and run specific tests. We are blocked on Fedora 33 and newer versions as their default crypto policies do not support ssh-rsa algorithm. I have tried proposing to patch our Paramiko locally using this PR, however the AWS Crypto experts believe the best path forward would be for Paramiko to officially support this.

bskinn:
Got it, thanks to you both for the info and context.

I've flagged this for @bitprophet's attention in triage.

bitprophet:
Apologies for missing this over the last month+, life's been hell. I'll make sure this goes out in the next bugfix rel, which should be soon (want to get CI working again first but that's mostly done now).

Dreamsorcerer:
Any chance this is going to happen soon? Going to become a bigger issue if this doesn't get packaged for Ubuntu before the next release. I think openssh has actually disabled ssh-rsa by default now.

bskinn:
Noted, @Dreamsorcerer, I've reraised the issue to @bitprophet's attention.

Does the end-of-Dec-2021 deadline you stated above still hold?

Dreamsorcerer:
It's just an estimate, I'm not a debian/ubuntu maintainer. But, I doubt it will make it in time if later than this year. The earlier it's released, the more likely it will get in.

bitprophet:
Planning on putting this out in 2.9.0 in the very near future. Not sure why I had it pegged as a bugfix; it's clearly a feature. Not like it matters a ton, but if I'm gonna semver, I'm gonna semver.

I just successfully released 2.8.1 so my dev pipeline has been repaired sufficiently - unless my dayjob catches serious fire this week I'm hoping to take an extra OSS day to do this and probably a few other new features.

manfred-kaiser:
I have tested the changes, but the changes does not work with a default installation of OpenSSH 8.8 or when SHA1 signatures are disabled.

According to RFC-4252, publickey authentication is split in 2 steps. during the first step the client can ask the server, if a specific public key is known. Following message is sent to the server: 

```
      byte      SSH_MSG_USERAUTH_REQUEST
      string    user name in ISO-10646 UTF-8 encoding [RFC3629]
      string    service name in US-ASCII
      string    "publickey"
      boolean   FALSE
      string    public key algorithm name
      string    public key blob
```
The field "public key algorithm name" should be the name of the signature algorithm name. If you are using SHA1 signatures the name "ssh-rsa" is used, but for SHA512 the name "rsa-sha2-512" should be used. 

The problem is that "ssh-rsa" is also used for RSA Keys.

The PublicBlob class parses a publickey and the key type is stored in the `key_type` property.

https://github.com/paramiko/paramiko/blob/13f28d65894e7737b01999a7475dccc71078c734/paramiko/pkey.py#L648-L672

When performing publickey authentication, the information from the `PublicBlob` class is used to gernerate the message, which is sent to the server performing the public key lookup step.

https://github.com/paramiko/paramiko/blob/d43ed76a44c451b2b9031883a1a7e55de59fdf79/paramiko/auth_handler.py#L285-L294

The problem is, when using the publickey information from `PublicBlob`, the keytype is sent to the server:

https://github.com/paramiko/paramiko/blob/d43ed76a44c451b2b9031883a1a7e55de59fdf79/paramiko/auth_handler.py#L290

When using a RSA key, the keytype is `ssh-rsa` and when this information is added to the message, the server interprets this as `SHA1` hash, because SHA1 is also defined as `ssh-rsa` and authentication with disabled SHA1 signatures are not possible.

The message, which is sent to the server should also changed and the signature name and not the key type must be sent to the server.

manfred-kaiser:
You have defined the classes `RSAKey`, `RSASHA256Key` and `RSASHA512Key`. While this is possible, it makes the usage of paramiko more complicated. The reason is, that the developber/user must decide, which class should be used during authentication.

For example a SSH server is configured to allow only SHA512 signatures for RSA keys, the user must specify the correct class. The problem is, that the user does not know, why the authentication fails. Trying out all possibilities is also not an option, because each login attempt with the wrong signature triggers a failed login. If the server is configured with fail2ban it's likely that the user get's blocked.

OpenSSH converts all publickeys to SHA512 if possible. This can be tested with following commands:

```
ssh-keygen -t ssh-rsa # create a rsa key with SHA1 signature
```
Copy that key in you authorized keys file.

```
ssh -vvv remotehost
...
debug3: authmethod_is_enabled publickey
debug1: Next authentication method: publickey
debug1: Offering public key: /home/manfred/.ssh/id_rsa RSA SHA256:G4t3DdQG+P4SjIJp7X8HnfL6Nj9yym7QGr8Q4h0RzOU
debug3: send packet: type 50
debug2: we sent a publickey packet, wait for reply
debug3: receive packet: type 60
debug1: Server accepts key: /home/manfred/.ssh/id_rsa RSA SHA256:G4t3DdQG+P4SjIJp7X8HnfL6Nj9yym7QGr8Q4h0RzOU
debug3: sign_and_send_pubkey: RSA SHA256:G4t3DdQG+P4SjIJp7X8HnfL6Nj9yym7QGr8Q4h0RzOU
debug3: sign_and_send_pubkey: signing using rsa-sha2-512 SHA256:G4t3DdQG+P4SjIJp7X8HnfL6Nj9yym7QGr8Q4h0RzOU
debug3: send packet: type 50
debug3: receive packet: type 52
debug1: Authentication succeeded (publickey).
```

The reason, why this is working is, that the OpenSSH client support [signature flags](https://datatracker.ietf.org/doc/html/draft-miller-ssh-agent-04#section-4.5.1).

```
4.5.1.  Signature flags

   Two flags are currently defined for signature request messages:
   SSH_AGENT_RSA_SHA2_256 and SSH_AGENT_RSA_SHA2_512.  These two flags
   are only valid for "ssh-rsa" keys and request that the agent return a
   signature using the "rsa-sha2-256" or "rsa-sha2-512" signature
   methods respectively.  These signature schemes are defined in
   [I-D.ietf-curdle-rsa-sha2].
``` 
When those flags are present, the OpenSSH client/agent uses sha256/sha512 instead of sha1.

Dreamsorcerer:
Is #1925 related to that problem? I'm under the impression that both PRs are needed.

manfred-kaiser:
> Is #1925 related to that problem? I'm under the impression that both PRs are needed.

I think #1925 is also relevant for OpenSSH 8.8, but should not affect this pull request.  At the moment, I have only testes a forwarded agent, and it worked without any problems, even with older paramiko versions.

**Publickey authentication - hashtype must be used instead of keytype**

Another problem is the usage of the keytype when sending a publickey request to ask the server if a specific key is known (1. step of publickey authentication). According to the RFC, this must be the hashtype. For SHA1, this is "ssh-rsa" and for SHA2 this is rsa-sha2-256 or rsa-sha2-512.
In older Pararmiko versions, this was not a problem, because he RSAKey class used a sha1 hash, which is defined as "ssh-rsa". The keytype for RSA keys is also "ssh-rsa". This was the reason, why this worked without any errors.

https://github.com/paramiko/paramiko/blob/d43ed76a44c451b2b9031883a1a7e55de59fdf79/paramiko/auth_handler.py#L290

There is also a good blog entry about this topic, which makes it more clear: https://levelup.gitconnected.com/demystifying-ssh-rsa-in-openssh-deprecation-notice-22feb1b52acd

I my opinion the change from keytype to hashtyp is a bugfix and not a new feature.

bitprophet:
> Would this work with: #1103

As far as I can tell, the two are entirely orthogonal; this is about adding newer RSA hostkey and client-key signature hashing, whereas #1103 is about fingerprint emission (eg as seen in known_hosts or when printing info about keys).

That said, that ticket is on my "merge soon" list and looks pretty small so I may try to cram it into this release anyways.

> Any chance this is going to happen soon? Going to become a bigger issue if this doesn't get packaged for Ubuntu before the next release. I think openssh has actually disabled ssh-rsa by default now.

I haven't looked yet to see what version of OpenSSH various repos package up, but they only just dropped support for ssh-rsa on their end in 8.8 which is about 2-3 months old. FWIW.

----

I spent Saturday reading all the relevant RFCs (which was arguably `REQUIRED`, `MAY` have given me a headache and certainly is not usually `RECOMMENDED`) so I think I see the way forward here, which is to grab this PR and then make some hopefully minor adjustments re: when to offer the string `ssh-rsa` and when to offer eg `ssh-rsa2-sha256` (touched on in some of the comments above). I'll then try testing against OpenSSH with and without `ssh-rsa` in `HostKeyAlgorithms` and `PubkeyAcceptedAlgorithms`.

Dreamsorcerer:
> I haven't looked yet to see what version of OpenSSH various repos package up, but they only just dropped support for ssh-rsa on their end in 8.8 which is about 2-3 months old. FWIW.

Well, Ubuntu will likely end up with whatever is packaged in Debian next month. But, the main issue is that as this rolls out in other distros, which will happen before the next (24.04) release, users will not be able to use the packaged version of paramiko to connect to servers running any of those other releases (and they already can't connect to servers with sensible secure configurations). In short, there are plenty of servers users already can't connect to, and it will likely be the vast majority of them before the next LTS releases in 2+ years.

bitprophet:
After more reading, the more recent comments are semi accurate re: edge cases with this feature.

- First, yes, as-is these new classes at least need to tweak what key type they report; they should always report "as" `ssh-rsa` in the first part of the auth submission, regardless of what signature hash will be used. (As noted by @manfred-kaiser)
- Also as noted by Manfred, there's a deeper problem here of how to know which class to use for an arbitrary RSA-type key (mostly regardless of source) when acting as a client submitting pubkey auth. 
  - RSA keys themselves do not change for this support; we cannot know just from the key data itself what algorithm "should" be used for signing.
  - Paramiko currently just tries all key types in arbitrary order when loading keys. We can't simply extend this because, as noted, any RSA key "could be" used for any of the 3 possible signing options. And any of the 3 will "work" to load an RSA key; then the other two would never get tried, even if the server only accepts one of the later options.
  -  Which brings us to what OpenSSH client does to figure this out, namely: compare its client config for `PubkeyAcceptedAlgorithms` (a list whose default value is identical to the default of, but is otherwise distinct from, `HostkeyAlgorithms`) with what the server advertises as acceptable signatures. It finds the matching subset, if any, similar to what happens during key exchange, and then picks the 'highest' ranked such signature algorithm.
  - There are multiple problems here for Paramiko!
    - We don't yet support `PubkeyAcceptedAlgorithms`; if anything, the abovementioned list of key types we use to load a given key filename is our de-facto such list, but ... not really.
    - We don't yet support the mechanism OpenSSH uses to send that signature list, which is `MSG_EXT_INFO`, submitted near the end of the key exchange process (and which OpenSSH client stores in its big struct of server related data).
  - My gut instinct, since this has waited so long already & is time-sensitive, is to do the obviously expedient thing and reuse our existing list of "happy little algorithms" (our equiv of `HostkeyAlgorithms`) as the list of preference order.
- At this point I'm not sure we do in fact want 3 separate classes, since as noted by others and myself here, it's not necessarily sensible to decide at key loading time which "flavor" we want, but only at authentication time. I'll workshop this on my end as I adapt the rest of the patch.

bitprophet:
> > I haven't looked yet to see what version of OpenSSH various repos package up, but they only just dropped support for ssh-rsa on their end in 8.8 which is about 2-3 months old. FWIW.
> 
> Well, Ubuntu will likely end up with whatever is packaged in Debian next month. But, the main issue is that as this rolls out in other distros, which will happen before the next (24.04) release, users will not be able to use the packaged version of paramiko to connect to servers running any of those other releases (and they already can't connect to servers with sensible secure configurations). In short, there are plenty of servers users already can't connect to, and it will likely be the vast majority of them before the next LTS releases in 2+ years.

Totally, I wasn't meaning to imply this is NOT time sensitive, was just dropping a detail I hadn't seen specifically mentioned. Quite right that even if most servers in the wild are still capable of supporting `ssh-rsa`, it's now bad practice.

manfred-kaiser:
>  * First, yes, as-is these new classes at least need to tweak what key type they report; they should always report "as" `ssh-rsa` in the first part of the auth submission, regardless of what signature hash will be used. (As noted by @manfred-kaiser)

the first part during public key algorithm must contains the signature algorithm. The keytype is defined in the publickey blob.

```c
      byte      SSH_MSG_USERAUTH_REQUEST
      string    user name in ISO-10646 UTF-8 encoding [RFC3629]
      string    service name in US-ASCII
      string    "publickey"
      boolean   FALSE
      string    public key algorithm name  // this should be the signature algorithm
      string    public key blob  // public key blob is the publickey and the key type is defined in this blob
 ```

>  * And we can potentially use the server's submitted list of hostkey algos 
>    as the counter-check, depending, though that's not currently preserved 
>    after figuring out which server key to care about (if it sent >1).

The hostkey algos can have "ssh-rsa" enabled, but `PubkeyAcceptedAlgorithms` has disabled "ssh-rsa". This check does not work, because in my testsetup with disabled "ssh-rsa" (SHA1) signatures, the host key algorithms still includes "ssh-rsa".

>  * At this point I'm not sure we do in fact want 3 separate classes, since as noted by others and myself here, it's not necessarily sensible to decide at key loading time which "flavor" we want, but only at authentication (or hostkey verification) time. I'll workshop this on my end as I adapt the rest of the patch.

I would prefere on RSAKey class because there is only one RSA Key Type. Which signature algorithm should be used must be defined in a different place.

There is a simple workarround, to be compatible with servers, which have SHA1 signatures (ssh-rsa) disabled. We can define another default algorithm like SHA512, but this could break compatibility with servers, which only supports SHA1 signatures. (ssh-rsa)

Without information from the server, it's not possible to know which signature algorithm should be used.

**Comment by bitprophet** [comment]
> The hostkey algos can have "ssh-rsa" enabled, but `PubkeyAcceptedAlgorithms` has disabled "ssh-rsa". This check does not work, because in my testsetup with disabled "ssh-rsa" (SHA1) signatures, the host key algorithms still includes "ssh-rsa".

Correct, the change I mentioned would only be a "gets you part of the way there" workaround. as you noted later:

> Without information from the server (`MSG_EXT_INFO`), it's not possible to know which signature algorithm should be used.

This is the only 100% solution. I'll take a stab at it, probably won't be _that_ difficult, but I'm willing to release something that's a partial solution otherwise & add this later.

> I would prefere one RSAKey class because there is only one RSA Key Type. Which signature algorithm should be used must be defined in a different place.

Yes, this is in line with my current thinking, as noted. I'm writing my own patch and will reference this PR where necessary.

**Comment by bitprophet** [comment]
Quick side note that we cannot yet get 100% OpenSSH parity regarding the behavior of sha2 without _also_ adding config support for `PubkeyAcceptedAlgorithms`, but I'm not super worried about that as regrettably our support for honoring (vs simply parsing & making available) ssh_config has always been a bit weak, and/or just not feasible w/o the stronger framework-style support found in Fabric.

**Comment by manfred-kaiser** [comment]
> Quick side note that we cannot yet get 100% OpenSSH parity regarding the behavior of sha2 without _also_ adding config support for `PubkeyAcceptedAlgorithms`, but I'm not super worried about that as regrettably our support for honoring (vs simply parsing & making available) ssh_config has always been a bit weak, and/or just not feasible w/o the stronger framework-style support found in Fabric.

`PubkeyAcceptedAlgorithms` can be used on the client side and on the server side. Using the client configuration does not solve the problem, because the client can allow sha1 signatures and the server side can have sha1 disabled.

Without information from the server side, it's not possible to know which algorithms will be accepted by the server.

**OpenSSH documentation:**

**Client side (ssh_config):** Specifies the signature algorithms that will be **used** for public key authentication...
**Server side (sshd_config):** Specifies the signature algorithms that will be **accepted** for public key authentication...

**Comment by manfred-kaiser** [comment]
@bitprophet What do you think about changing the default algorithm for RSA keys to SHA512 and make this configurable, so that users can define another hash algorithm?

This would solve the problem for most OpenSSH versions >= 7.2 and if a user needs to connect to a very old SSH server, the user can enable a legacy mode, which uses SHA1 for RSA keys.

**Comment by bitprophet** [comment]
Making the various accepted algorithms/ciphers/key types configurable is a longterm desire, but I'm hoping to get this out w/o adding too many dependencies. FWIW I am intending to follow OpenSSH here by having the sha2 based algorithms appear earlier in those lists, thus any connection to servers implementing sha2 would end up preferring that method.

I am still rummaging around in the code on my end and writing tests; intending to have this done by end of week unless something else blows up on me.

**Comment by bitprophet** [comment]
Have the tests passing for hostkey exchange, which also necessitated doing part of the work for pubkey signatures as well. Hoping to wrap that and agent stuff up by early next week, if not sooner, and get this published.

**Comment by JackCurtin-eaton** [comment]
I'd love to be able to have this for my regression tests this week so I can update our product to open SLS 1.1.1l :)

**Comment by bitprophet** [comment]
(still poking away, @JackCurtin-eaton ...)

Looking at pubkey auth here and hitting a possible wrinkle, namely that Paramiko's design makes trying >1 pubkey auth step per actual key, a bit fraught - partly due to the underlying network aspect of "send message, wait for reply, act on reply, continue",  and partly due to the use of threading events (though I suspect explicitly given events is not widely used and it already doesn't work if one is using eg 2FA).

Approaches open to me seem to be:

1. Naively wrap all calls to `Transport.auth_pubkey` in a loop over the possible RSA algorithm types. Gross, copypasta. Also may just not work well in some spots without even more awful logic.
1. New method on `Transport` that does the above for you. A bit better, still have to tweak the other calls to use this method but that's not awful. *However* anybody calling the old method directly will not know to do this w/o reading the changelog & may still be all "wah! why my RSA SHA2 not working?". And this adds more API complexity where we really shouldn't have to.
1. Push the loop into the existing method. Might be fine, but anybody trying to pass in `event` will be in big trouble. Is anybody really doing that? No way to tell!
1. Copy OpenSSH somewhat and just pray selecting the first possible algorithm (honoring `disabled_algorithms`, so users can control this somewhat) will be the right one. Yea, not great.
1. Copy OpenSSH fully and implement `server-sig-algs` support after all, which as noted upthread allows almost-complete removal of guessing (outside of some weird legacy server that might support SHA2 but not server-sig-algs).

Leaning towards the last option there, actually - I think if I have to do weird funky logic and/or mess with the network events, I'd rather just...add the new feature anyways.

**Comment by bitprophet** [comment]
Got MSG_EXT_INFO working, and most of what else is needed for pubkey auth (hitting some weird snags near the end). Hopefully not a lot more from here, we'll see how tomorrow goes.

**Comment by bitprophet** [comment]
Finally figured out why SHA2 based pubkey auth wasn't working with a real OpenSSH target but was working vs ourselves. Some very old code turned out to have multiple subtly distinct copies floating around and I'd only found/updated one of them.

Remember kids, friends don't let friends copy and paste.

Need to wrap this present in some more tests (& run a few more real world tests) but hopefully putting it on PyPI soon.

**Comment by bitprophet** [comment]
For anyone wanting to do some last minute testing, the branch as it stands is here: https://github.com/paramiko/paramiko/tree/rfc8832-sha2-key-algo - though it is still a feature branch getting force pushed to often so buyer beware. I expect to squash it into only a few commits before merging.

**Comment by bitprophet** [comment]
OK, I ended up adding serverside EXT_INFO support too as I needed it for testing the client side (derp).

As far as I can tell right now, this works reasonably well and is pretty well tested too (and I made sure that it passes on Python 2 locally as well, thank you containers!) - Circle is currently happy as well: https://app.circleci.com/pipelines/github/paramiko/paramiko/140/workflows/de9d64c8-eb42-48c4-86e4-9a07b86db929

Only thing left is agent. I intend to tackle that tomorrow morning and release to PyPI.

**Comment by bitprophet** [comment]
Going to merge my work to main and close this ticket since agent is arguably separate tickets. Thanks all for the patches/feedback/discussion.

**Comment by TTimo** [comment]
How can third parties (me) track when this fix will be available in a new official paramiko release?

**Comment by bitprophet** [comment]
@TTimo I'll report back here when it's released, but otherwise you can always bookmark the changelog (https://paramiko.org/changelog.html) and look there to see if the entry for this ticket is under a tagged release or still under "next release".

**Comment by bitprophet** [comment]
Cool, the changes required for agent support are very minor, though once again I had to write my own bits as the existing PRs weren't quite lining up with the latest state of the code. Putting finishing touches on now and a few tests (agent support is not tested at all, argh, wtf folks - I'm only gonna add a few very quick ones but better than nothing!).

**Comment by bitprophet** [comment]
2.9 now up on PyPI!

**Comment by bskinn** [comment]
👏👏👏

Marathon effort. Thank you!

**Comment by bitprophet** [comment]
https://github.com/paramiko/paramiko/issues/1955 spawned from this thread. Already have a proving test and fixed enough that it passes, just need to fix regressions in other tests now 🙃

**Comment by Kami** [comment]
Thanks for contributing this change. 

Here is just some info / context on how we handle "backward incompatible" nature of this change as consumers in Libcloud - https://github.com/apache/libcloud/pull/1685.

In short, if we detect paramiko >= 2.9.0 and we get auth error back from the server (afaik, there is no easy way to catch and only retry on a more granular exception, but please correct me if I'm wrong), we try to fall back to the previous approach (disable sha-2 variants). We confirmed this workaround works correctly in various setups.

Sadly this does have a potential to decrease security (think "downgrade" attack) in case user doesn't require backward compatibility layer and that's why we noticed how to disable it in docs / upgrade notes / deployment page.

Feedback and comments are of course welcome.

**Comment by perlun** [comment]
@bitprophet I know you are incredibly busy so sorry for disturbing you. I have trouble getting this to work with my (RSA-based) key though. :thinking: Here's what it logs on the (Ubuntu 22.04) server side:

```
Feb 21 10:58:02 ubuntu-2204-centre-test-c4d5124d sshd[19345]: userauth_pubkey: key type ssh-rsa not in PubkeyAcceptedAlgorithms [preauth]
Feb 21 10:58:02 ubuntu-2204-centre-test-c4d5124d sshd[19345]: userauth_pubkey: key type ssh-rsa not in PubkeyAcceptedAlgorithms [preauth]
Feb 21 10:58:02 ubuntu-2204-centre-test-c4d5124d sshd[19347]: userauth_pubkey: key type ssh-rsa not in PubkeyAcceptedAlgorithms [preauth]
Feb 21 10:58:02 ubuntu-2204-centre-test-c4d5124d sshd[19347]: userauth_pubkey: key type ssh-rsa not in PubkeyAcceptedAlgorithms [preauth]
```

Connecting via OpenSSH (OpenSSH_8.9p1 Ubuntu-3ubuntu0.1, OpenSSL 3.0.2 15 Mar 2022) to the same host with the same key works fine. Paramiko 2.12.0 running from pytest-testinfra 7.0.1.

`ssh host -v` gives me this on connecting. Any ideas? :thinking: 

```
debug1: Offering public key: /home/plundberg/.ssh/id_rsa RSA SHA256:nEecUUWt8//qdNKTcQ5nPAGpTLXmWQm5jBt/Qzx4JvQ agent
debug1: Server accepts key: /home/plundberg/.ssh/id_rsa RSA SHA256:nEecUUWt8//qdNKTcQ5nPAGpTLXmWQm5jBt/Qzx4JvQ agent
```

bskinn:
@perlun It'll be best if you post this as a new issue, linking back to this PR. That way it will be more visible and also I can tag it for `Support`.

perlun:
> @perlun It'll be best if you post this as a new issue, linking back to this PR. That way it will be more visible and also I can tag it for `Support`.

Thanks, good point @bskinn. :+1: Here we go: https://github.com/paramiko/paramiko/issues/2191

Kache:
Ran into this "slightly backwards incompatible" issue, as summarized in [Paramiko fails ... when upgrading to 2.9.0+](https://stackoverflow.com/questions/75836730/paramiko-fails-connecting-w-private-key-file-is-encrypted-when-upgrading-to-2)

Aren't the [paramiko changelog docs for 2.9.0](https://www.paramiko.org/changelog.html#2.9.0) wrong? I wouldn't have been able to solve the incompatibility without finding [this post](https://github.com/paramiko/paramiko/pull/1643#issuecomment-1120234137) by @Kami with the correct workaround.

bskinn:
Thanks for reporting this, @Kache -- could you repost this as a new issue so that it can be independently addressed?

PR #1644, opened 2020-03-19
AgentSSH - sign_ssh_data can take optional flags to indicate hashalg

commonism:


paramiko/agent.py
```diff
@@ -407,12 +410,22 @@ def __str__(self):
     def get_name(self):
         return self.name
 
-    def sign_ssh_data(self, data):
+    def sign_ssh_data(self, data, flags=0):
+        """
+        Sign a blob of data with this private key, and return a `.Message`
+        representing an SSH signature message.
+
+        Use flags to sign using sha2-256 or sha2-512, default is sha1
```
ploxiln: default is sha1 only for rsa keys, not for ecdsa or ed25519, right?

commonism: Yes. flags are not defined for anything but ssh-rsa.

ploxiln:
How might this be used? This seems like just one needed part, at least for a client with an rsa key for user auth.

To auth with key, one calls `Transport.auth_publickey()` (or `SSHClient` calls that for you), and I think that eventually ends up in `AuthHandler._parse_service_accept()` where `sign_ssh_data()` is called (and the key type name is embedded in the message there too). So that would need to know that for "ssh-rsa" keys, if it is using an AgentKey, it can call `sign_ssh_data()` with flag, and change the name from "ssh-rsa" to the appropriate sha2 variant?

Maybe the easier way to use this in paramiko, is to have the agent create 3 keys when it sees an rsa key, one for each hash algo variant. Each AgentKey reports its variant name, and automatically passes the appropriate flag to the agent when `sign_ssh_data()` is called, so `sign_ssh_data()` can keep the same method signature as on every other key type.

Just thinking out-loud ...

commonism:
I'm not looking for ssh authentication with this one. I just need to sign.
Sign data without having key and data in a single place by forwarding access to the key via ssh.

Used here:
https://github.com/m32/endesive/pull/43

ploxiln:
I see. Anyway, this is a useful reference/hint for implementing SSH auth, thanks.

bitprophet:

PR #1647, opened 2020-03-24
add filename in ssh key errors

anarcat:
Just getting an error like "Private key file is encrypted" is really
unhelpful: which keyfile? How do I tell which file paramiko tripped
on? I could not figure it out short of running an strace on the
program or injecting a debugger or trace in the source.

I tried to cover all the cases where a private key file is read, which
seem to be where most of the pain is (e.g. in #387).

With this patch, the filename is passed to the read functions, which
is rather unfortunate: it would be better if a file pointer would be
passed into those functions, so that we could just access the `.name`
parameter from there. But it seems that ship has sailed.


idlesign:
@bitprophet Hi, Jeff. 
Would you mind looking at this one and possibly merging?

ssbarnea (approved):

vsoch:
+1 this would be very helpful to me as well. Has it been replaced by another PR?

Issue #1593, opened 2020-01-08
BytesWarning with py3 -bb


justinpryzby:
Similar to #726 Python 3.6.8 -bb causes this.  I'm using -bb to try to catch issues while migrating from py2.  I didn't realize packages didn't all handle that already.

  File "/usr/local/lib/python3.6/site-packages/paramiko/transport.py", line 2109, in run
    handler(self.auth_handler, m)
  File "/usr/local/lib/python3.6/site-packages/paramiko/auth_handler.py", line 656, in _parse_userauth_banner
    self._log(INFO, "Auth banner: {}".format(banner))
BytesWarning: str() on a bytes instance

Issue #1074, opened 2017-09-25
BytesWarning: str() on a bytes instance


Rezney:
While running pytest I am getting the following two warnings:

```
  /usr/lib/python3.6/site-packages/paramiko/sftp_client.py:344: BytesWarning: str() on a bytes instance
    'open(%r, %r) -> %s' % (filename, mode, hexlify(handle)))
  /usr/lib/python3.6/site-packages/paramiko/sftp_file.py:86: BytesWarning: str() on a bytes instance
    self.sftp._log(DEBUG, 'close(%s)' % hexlify(self.handle))

-- Docs: http://doc.pytest.org/en/latest/warnings.html
```

which I believe is a bug in the code as:

```
# python3
Python 3.6.2 (default, Sep 22 2017, 08:28:09) 
[GCC 7.2.1 20170915 (Red Hat 7.2.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 
>>> from binascii import hexlify
>>> 
>>> hexlify(b'test')
b'74657374'
>>> 
>>> hexlify('test')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: a bytes-like object is required, not 'str'
>>> 
```

Thanks a lot for taking a look!

bitprophet:
Don't recall offhand what the handle objects are implemented as but this may just be a minor oversight from the Python 3 overhaul & hopefully fixable by using `paramiko.py3compat.b()` or similar. Thanks for the report!

Rezney:
Thanks for taking a look! PR #1075 submitted.

bitprophet:
Rolling into the PR.

Issue #726, opened 2016-04-19
BytesWarning triggered in SSHClient.connect()


johnthagen:
If the following code:

```python
import paramiko

client = paramiko.SSHClient()
client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
client.connect('127.0.0.1', username='user', password='password')
```
Is run in Python 3.5 with ``BytesWarnings`` enabled (`-bb` flag), an error is triggered:

```bash
$ python3 -bb para.py
Traceback (most recent call last):
  File "para.py", line 5, in <module>
    client.connect('127.0.0.1', username='user', password='ironmanmachine')
  File "/usr/local/lib/python3.5/site-packages/paramiko/client.py", line 348, in connect
    server_key)
  File "/usr/local/lib/python3.5/site-packages/paramiko/client.py", line 623, in missing_host_key
    (key.get_name(), hostname, hexlify(key.get_fingerprint())))
BytesWarning: str() on a bytes instance
```

bitprophet:
Is this causing an actual, real-world problem for you somewhere? I don't normally see people running with `BytesWarning` enabled.

----

I dug briefly out of curiosity and did find that e.g. `RSAKey.__str__` returns `self.asbytes()` and so a simple `print(key)` spits out a nice `TypeError: __str__ returned non-string (type bytes)`.

But that may be orthogonal since we're never actually `str`'ing the key itself, only the results of `.get_name` (returns a str obj), `hostname` (hopefully a str) and `hexlify` (oh hey! *this*, and the result of `PKey.get_fingerprint` that it operates on, *are* bytes. Huh.)

Docstring for `get_fingerprint` implies it returns a string, but the entire method's blame output shows it untouched by the big Python 3 upgrade, so I don't think anybody has yet had to confront whether this should legitimately be spitting out a string or bytes object. (Of course, for debugging purposes, we still need *some* string representation of it.)

`get_fingerprint` is implemented as simply `md5(self.asbytes()).digest()`. Under Python 2, `md5(xxx).digest()` prints a string representation of bytes; under Python 3, it's a bytes object.

Problem is it's not bytes representing encoded text, so the usual "turn a `bytes` into a `unicode`" suspects will all get pissy with us. Asking for a `repr` instead is kinda fugly because it doubles up all the backslashes, etc (vs what a Python 2 user sees natively).

----

[...]

PR #1661, opened 2020-04-11
Remove BytesWarnings in client.py by putting u() around every hexlify, other BytesWarnings fixes

aggieNick02:
I run python in an environment where BytesWarnings are enabled and promoted to errors. This was really helpful during py2->3 transitioning, and still helps prevent misuse of bytes as strings.

I'd like to keep doing this while using Paramiko, but I hit BytesWarnings when doing so. I looked at related issues for paramiko, and they've previously been fixed by putting a u() around the offending bytes. This is appropriate for calls to hexlify that are then passed to format(), since hexlify returns the bytes for the ASCII-encoded hex rep of the input.

This is what was done for the same kind of issue previously: https://github.com/paramiko/paramiko/issues/1074

I also fixed the same kind of issue, in auth_client.py, reported at:
https://github.com/paramiko/paramiko/issues/1593 

This should also close the issue 728 linked below. The issue requested a string-ish hex-digest, which would be more dramatic and risky for a wholesale transition, since there could be clients that still need the bytes-ish one. But the u() around the hexlifys going to format safely resolves being able to run with -bb.:
https://github.com/paramiko/paramiko/issues/726

I went ahead and found the rest of the uses of hexlify that were exclusively passed to format/print, and fixed similarly. I left other uses of hexllify alone, as there are definitely still places, e.g. tests, that fully expect bytes back.

### DIFF
--- a/demos/demo.py
+++ b/demos/demo.py
@@ -28,7 +28,7 @@ import socket
 import sys
 import time
 import traceback
-from paramiko.py3compat import input
+from paramiko.py3compat import u,input
 
 import paramiko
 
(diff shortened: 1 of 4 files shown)


aggieNick02:
Sorry, didn't realize there were whitespace checks. Do you want me to fix and supply a new PR?

PR #1665, opened 2020-04-16
Replace pytest-relaxed with plain pytest.raises

mgorny:
There is really no technical reason to bring pytest-relaxed to call
@raises as a decorator while plain pytest works just fine.  Plus,
pytest.raises() is used in test_sftp already.

pytest-relaxed causes humongous breakage to other packages
on the system.  It has been banned from Gentoo for this reason.

This is alternative to #1543 that doesn't remove tests.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -2,7 +2,6 @@
 invoke==1.6.0
 invocations==2.6.0
 pytest==4.4.2
-pytest-relaxed==1.1.5
 # pytest-xdist for test dir watching and the inv guard task
 pytest-xdist==1.28.0
 mock==2.0.0
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,7 +1,4 @@
 [pytest]
-# We use pytest-relaxed just for its utils at the moment, so disable it at the
-# plugin level until we adapt test organization to really use it.
-addopts = -p no:relaxed
 # Loop on failure
 looponfailroots = tests paramiko
 # Ignore some warnings we cannot easily handle.
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -33,7 +33,7 @@ import warnings
 import weakref
 from tempfile import mkstemp
 
-from pytest_relaxed import raises
+import pytest
 from mock import patch, Mock
 
 import paramiko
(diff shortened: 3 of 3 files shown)


bnavigator:
Any plans to merge this?

pytest-relaxed is broken with pytest > 5.4, even with https://github.com/bitprophet/pytest-relaxed/pull/10, it is broken with pytest > 6.1

https://github.com/bitprophet/pytest-relaxed/issues/12

radez (commented):
+1 to remove this dependency

mgorny:
Ping. Two years have passed already.

bskinn:
Thanks for pinging to bring this back up.

> pytest-relaxed causes humongous breakage to other packages on the system. It has been banned from Gentoo for this reason.

Other packages on what system? 

Are you saying that paramiko might be packaged into Gentoo, if not for the `pytest-relaxed` dependency? In other words, is there strong interest in paramiko *being* packaged in Gentoo, but it's blocked?

bskinn:
Got it - thanks for the context!

bitprophet:
> Basically, when pytest-relaxed was installed in order to run paramiko's tests, we've seen huge fallout from other Python packages. Please bear in mind that pytest loads all installed plugins by default, and pytest-relaxed changes its behavior quite a lot.

pytest-relaxed needs some love, but it still seems a bit odd that this is _my_ problem here in paramiko or pytest-relaxed, instead of it being a configuration issue when packaging pytest-using kit such that the test phase only uses its specific deps (or such that the prod install facing stage isn't including test-only dependencies).

Surely there are other pytest plugins out there that cause breaking changes - I find it hard to believe that a build/test regime that boils down to "install a very large subset of all pytest plugins all at the same time" was working swimmingly until relaxed came along?

----

I am actually intending to use _more_ pytest-relaxed in Paramiko's suite in the near/mid future, not less, so I can't merge this. However, I would definitely like to give pytest-relaxed the love it needs to work with newer pytests at the very least, and I'm certainly open to discussions about any truly bad behavior that remains after that point.

mgorny:
I'm not claiming that everything works "swimmingly". I'm talking about the difference between 5% of packages sometimes failing, and 95% of packages being totally broken.

PR #1666, opened 2020-04-16
Prefer built-in unittest.mock over external package

mgorny:
Prefer and use built-in unittest.mock in Python 3.3+ instead
of unnecessarily requiring the external mock package.  This helps
distributions that are phasing out Python 2 to remove redundant
packages.


mgorny:
Ping.

PR #1667, opened 2020-04-16
Skip tests requiring invoke if it's not installed

mgorny:
Since invoke is an optional dependency and only one group of tests
require it, skip them gracefully rather than failing if it's not
present.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -6,7 +6,11 @@ from socket import gaierror
 
 from paramiko.py3compat import string_types
 
-from invoke import Result
+try:
+    from invoke import Result
+except ImportError:
+    Result = None
+
 from mock import patch
 from pytest import raises, mark, fixture
 
@@ -742,6 +746,7 @@ def _expect(success_on):
     return inner
 
 
+@mark.skipif(Result is None, reason="requires invoke package")
 class TestMatchExec(object):
     @patch("paramiko.config.invoke", new=None)
     @patch("paramiko.config.invoke_import_error", new=ImportError("meh"))


mgorny:
Ping!

bskinn:
Seems like an easy and sensible add -- flagged as such.

bitprophet:
Thanks! I'll cherry pick this back a few lines for good measure.

@mgorny since you're in the set of folks who don't have/want invoke installed, I'd love your feedback on [this TODO](https://github.com/paramiko/paramiko/blob/690243665dee43d5f7cb280c4cd29eff2367433b/setup.py#L90-L92) - for the next big release I'm thinking the default install would include "arguable extras" like invoke, ed25519 etc, and users who care a lot about minimalism would need to `pip install paramiko[minimal]` or similar. (I would probably still leave the gssapi flavor as a yet-more-optional extra though - it's not nearly as common as ed25519, in my travels).

mgorny:
> @mgorny since you're in the set of folks who don't have/want invoke installed, I'd love your feedback on [this TODO](https://github.com/paramiko/paramiko/blob/690243665dee43d5f7cb280c4cd29eff2367433b/setup.py#L90-L92) - for the next big release I'm thinking the default install would include "arguable extras" like invoke, ed25519 etc, and users who care a lot about minimalism would need to `pip install paramiko[minimal]` or similar. (I would probably still leave the gssapi flavor as a yet-more-optional extra though - it's not nearly as common as ed25519, in my travels).

Hmm, is that even possible? I always thought that extras can only add more dependencies, not remove them.

mgorny:
Hmm, I suppose [PEP 508](https://peps.python.org/pep-0508/) technically permits using `extra != "minimal"` but I'm not sure if 1) you won't hit buggy package managers that don't account for this, and 2) if any of the existing build systems let's you specify this cleanly.

bskinn:
Yeah, I'll be following your progress on this, @bitprophet -- I have a similar desire for a `minimal`-like "subtractive extra" on one of my projects, and I couldn't see any way to achieve it. 

> Hmm, I suppose [PEP 508](https://peps.python.org/pep-0508/) technically permits using `extra != "minimal"`

I don't think it does, @mgorny -- the PEP [says pretty directly](https://peps.python.org/pep-0508/#extras) that extras always introduce additional dependencies, and that all specified extras result in a union of the dependencies of the named extras.

bitprophet:
Yea that's a good point, isn't it, that extras go only one way 🤔 wonder where I got the idea that it didn't, unless I've just spent too long around regular operating system style packages lately...

Will have to think on it some more. I still like the idea of ed25519 being in the default install (wonder if cryptography itself has support for it internally now?) though invoke is still in a bit of a gray zone, and as noted, I still don't love the idea of the average user being hit with gssapi.

mgorny:
…or just add `all` and `default` extras, so people can `pip install paramiko[default]` or `paramiko[all]`.

PR #1678, opened 2020-05-06
Set timeout to prevent transport hanging during initialization

kvancamp:
This is to address issue 1677 which I opened (Paramiko Transport hangs during initialization)

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -406,6 +406,7 @@ class Transport(threading.Thread, ClosingContextManager):
                     # addr = sockaddr
                     sock = socket.socket(af, socket.SOCK_STREAM)
                     try:
+                        sock.settimeout(self._active_check_timeout)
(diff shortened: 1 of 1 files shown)


paramiko/transport.py
```diff
@@ -406,6 +406,7 @@ def __init__(
                     # addr = sockaddr
                     sock = socket.socket(af, socket.SOCK_STREAM)
                     try:
+                        sock.settimeout(self._active_check_timeout)
```
ploxiln, line 409: `_active_check_timeout` is 0.1 seconds, seems not quite enough for a reasonable connect() timeout

I'd use `banner_timeout` but that's set-up later and can't be configured by the user until after `Transport.__init__()` completes. Maybe best to just use the default `banner_timeout` value of 15 seconds here ...

kvancamp: Thanks! I'll give it a try.

PR #1687, opened 2020-05-15
Do not send several sshauth-service requests

### LINKED ISSUE #1390: 'SSHException: No existing session' when multiple ssh keys in ssh-agent
Although similar ticket exists #1097 
I don't have exactly the same environment/setting, when this exception occurs.

**Example**

>import paramiko
>ssh = paramiko.SSHClient()
>ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
>ssh.connect(ciscodevice, username=user, allow_agent=True)

**Setup** 

- Debian 9, Python 2.7.13, paramiko==2.4.2

- ssh-agent running, with 2 rsa-2048 keys (neither one is passphrase protected)

- Trying to connect to Cisco device (so far i tried 1941, 4500X, CX2960)

**Expected behavior**

Paramiko should try to use every key available from ssh-agent, to find the right one. This works if i try to connect to any linux server.

**Result**
```

  File "/home/me/.virtualenvs/xyz/local/lib/python2.7/site-packages/paramiko/client.py", line 437, in connect
    passphrase,
  File "/home/me/.virtualenvs/xyz/local/lib/python2.7/site-packages/paramiko/client.py", line 749, in _auth
    raise saved_exception
paramiko.ssh_exception.SSHException: No existing session
```

**Notice**

If i remove unnecessary key from ssh-agent, and only leave the right one, it connects successfully.

vincentbernat:
Some SSH servers do not expect to receive more than one
sshauth-service. The RFCs doesn't seem to say anything definitive about
this, but RFC 4251 says in the introduction:

> The client sends a service request once a secure transport layer
> connection has been established. A second service request is sent
> after user authentication is complete. This allows new protocols to
> be defined and coexist with the protocols listed above.

So, it may be valid to not accept subsequent service request during
authentication.

To avoid sending a sshauth-service request for each auth attempt, we
reuse the AuthService() object for each authentication request to keep
track if we are already in the sshauth service.

This may fix #1390. However, alone, this is not enough because in my case, the remote SSH server is also sending a type 60 message/`MSG_USERAUTH_PK_OK`, despite the presence of the signature and Paramiko refuses to handle it because it confuses it for a `MSG_USERAUTH_INFO_REQUEST`.


vincentbernat:
I have added an additional commit to handle message type 60 with the following rationale:

> While Paramiko always attach a signature when proposing a public key,
> some servers, notably SSH-2.0-Cisco-1.25 may not expect such a
> signature and send unconditionally a `MSG_USERAUTH_PK_OK` (same number
> as `MSG_USERAUTH_INFO_REQUEST`). In this case, we just send again our
> userauth request and it gets accepted.

This fixes #1390 for me, but it is not directly related as I hit this specific bug ("Illegal info request from server") even with only one key in my agent. It could be a separate PR but it would require the first commit of this PR. So, I am bundling everything here.

sushilkar:
Might be related https://github.com/paramiko/paramiko/issues/1707

electrocnic:
```
/path/to/.venv/lib/python3.8/site-packages/paramiko/kex_ecdh_nist.py:39: CryptographyDeprecationWarning: encode_point has been deprecated on EllipticCurvePublicNumbers and will be removed in a future version. Please use EllipticCurvePublicKey.public_bytes to obtain both compressed and uncompressed point encoding.
  m.add_string(self.Q_C.public_numbers().encode_point())
/path/to/.venv/lib/python3.8/site-packages/paramiko/kex_ecdh_nist.py:91: CryptographyDeprecationWarning: Support for unsafe construction of public numbers from encoded data will be removed in a future version. Please use EllipticCurvePublicKey.from_encoded_point
  self.Q_S = ec.EllipticCurvePublicNumbers.from_encoded_point(
/path/to/.venv/lib/python3.8/site-packages/paramiko/kex_ecdh_nist.py:103: CryptographyDeprecationWarning: encode_point has been deprecated on EllipticCurvePublicNumbers and will be removed in a future version. Please use EllipticCurvePublicKey.public_bytes to obtain both compressed and uncompressed point encoding.
  hm.add_string(self.Q_C.public_numbers().encode_point())
Unknown custom user-made host key policy (ignore)
  File "main/main.py", line 148, in <module>
    main(args)
  File "main/main.py", line 129, in main
    client.connect(args.host, vms[current_vm], username, password)
  File "/path/to/.venv/lib/python3.8/site-packages/paramiko/client.py", line 421, in connect
    self._auth(
  File "/path/to/.venv/lib/python3.8/site-packages/paramiko/client.py", line 720, in _auth
    raise saved_exception
  File "/path/to/.venv/lib/python3.8/site-packages/paramiko/client.py", line 707, in _auth
    self._transport.auth_password(username, password)
  File "/path/to/.venv/lib/python3.8/site-packages/paramiko/transport.py", line 1370, in auth_password
    raise SSHException('No existing session')
paramiko.ssh_exception.SSHException: No existing session
```

I am not sure whether I installed the branch correctly in my project's virtualenv though. (I just replaced the official release of paramiko in .venv/lib/.../paramiko with the inner paramiko/paramiko of your branch)

bob-schumaker:
You can remove some "duplicated" code:

def __init__(...):
    ...
    self._auth_handler = None

@property
def auth_handler(self) -> AuthHandler:
    if self.auth_handler is None:
        self._auth_handler = AuthHandler(self)
    return self._auth_handler

louiecn:
I’ve reproduced this issue with an sshd server and multi-step auth - the first USERAUTH_REQUEST is normal then paramiko sends a second SERVICE_REQUEST which seems to be causing the server to forget the partial auth.

Haven’t tested the changes but the authservice_accepted stuff looks like it should work.

PR #1693, opened 2020-05-20
Correct processing of "match" types in config

SphericalHam:
Without lower-casing the match type, we miss detecting the special-case
match types of "canonical" and "all".

This same change also addresses another bug in config processing, where
an unkonwn match type will always be evaluated to be matching.


SphericalHam:
Whether or not this is a correct change really depends on whether the match conditions are considered to be keywords. It's reasonable to interpret it either way.

Openssh doesn't require them to be lower-case.

PR #1697, opened 2020-05-22
Add socket.MSG_PEEK support to Channel.recv

calebstewart:
I recently added an issue for this feature, but decided to just try to implement it. I added docstring updates for the added parameter. Currently, I used a `raise NotImplementedError` if flags besides `socket.MSG_PEEK` were used.

I have tested this, and it works. I added a line to the `demos/demo.py` which tested the functionality, and I am also currently using this functionality in my personal project (by referencing my fork explicitly). If there are any issues, please let me know.

An example of the line to add to `demo.py` to show this change in action (starting at line 174):

```python
    chan.invoke_shell()
    print("*** Here we go!\n")

    print("The shell is about to say: ", chan.recv(4096, flags=socket.MSG_PEEK))

    interactive.interactive_shell(chan)
```

### DIFF
--- a/paramiko/buffered_pipe.py
+++ b/paramiko/buffered_pipe.py
@@ -124,7 +124,7 @@ class BufferedPipe(object):
         finally:
             self._lock.release()
 
-    def read(self, nbytes, timeout=None, flags: int = 0):
+    def read(self, nbytes, timeout=None, flags = 0):
         """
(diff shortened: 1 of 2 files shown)


JohnHammond:
Now that this has been reviewed, will this be merged?

Issue #1700, opened 2020-06-02
channel close/feed race condition


rousku:
close metho in channel.py locks on channel lock. _feed method doesn't acquire channel lock. Following can happen:
# Channel.close()
# self._pipe.close()
# self._rsock.close()
# self._wsock.close()
# Channel._feed()
# self.in_buffer.feed(s)
# self._event.set()
# self._pipe.set()
# self._wsock.send(b"*") <- Crash

PR #1701, opened 2020-06-03
WindowsPipe/PosixPipe close() and set() are now thread-safe #1700

rousku:
Fix for #1700


ploxiln:
this appears similar to https://github.com/paramiko/paramiko/pull/1235

ploxiln:
(not the same pipes ... just similar ... nvm)

Issue #1631, opened 2020-02-26
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb1 in position 84: invalid start byte


paragonjh:
paramiko version: 2.6.0
When I read the stdout of Korean alphabet in code get this error:

Traceback (most recent call last):
  File "gtest_runner.py", line 305, in execute_zephyr_test
    runResults = executor.runTest()
  File "gtest_runner.py", line 65, in runTest
    for line in self.stdout:
  File "/vol/users/june1103.kim/.local/lib/python3.6/site-packages/paramiko/file.py", line 125, in __next__
    line = self.readline()
  File "/vol/users/june1103.kim/.local/lib/python3.6/site-packages/paramiko/file.py", line 334, in readline
    return line if self._flags & self.FLAG_BINARY else u(line)
  File "/vol/users/june1103.kim/.local/lib/python3.6/site-packages/paramiko/py3compat.py", line 147, in u
    return s.decode(encoding)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb1 in position 84: invalid start byte

please help me :).
I want to read the Korean alphabet in code using parameko.

hakaishi:
Fix in progress. There is a pull request at #1995

Issue #1443, opened 2019-05-11
Hability to ignore unidcode decode errors.


andredriem:
I ran on a issue today where I as trying to read a File object from a malformed utf-8 string with an invalid character. I managed to fix my issue creating my own readlines() function with a try/catch block, but I don't think this solution is resistant to future updates from this libary since I cant' guarantee that the file cursor will move to the next line before this error occur.
 
```python
def readlines(file):
  lines = []
  eof = False
  while not eof:
    try:
      line = file.readline()
      if line == '':
        eof = True
      else:
        lines.append(line)
    except UnicodeDecodeError as e:
      lines.append(e.object.decode('utf-8',errors='replace'))
  return lines
```

One simple solution to this would be adding a default error handling in line 68.  

https://github.com/paramiko/paramiko/blob/cf7d49d66f3b1fbc8b0853518a54050182b3b5eb/paramiko/py3compat.py#L60-L71

changing to 

```python
return s.decode(encoding, errors='replace')
```

or even better giving the user the choice of how she/he would handle the errors via an optional argument.

hakaishi:
continuation of previous pull request at #1995

Issue #1365, opened 2019-01-04
 'utf-8' codec can't decode


Doubleuncle:
There is an error reading stdout if contains any line with non utf-8 characters. 
I wanna get the all thing about the path(uid,size,permission……）,and the files contains utf-8,gbk, ansi characters.When i use 'stdout.readlines()' it always except error  " 'utf-8' codec can't decode……“

stdin, stdout, stderr = client.exec_command('ls -al')
res_list = stdout.readlines()

So has anyone encountered such problem can tell how to solve it?

martinloiseau:
Smae issue here, did you find any fix?

hakaishi:
Fix in progress. Pull request at #1995

louisgoodier-metservice:
I just lost an hour looking at this to realise a workaround... 

stdout._set_mode('b') #Binary mode
print(stdout.readlines())

Issue #546, opened 2015-06-27
Crashes on filenames that are not UTF-8 (listdir, maybe others)


larswirzenius:
In Obnam, my backup program, I use Paramiko to access files over SFTP. Mostly, Obnam users use this to access the backup repository, but some use it to access their live data. Live data files have all sorts of nasty filenames, and someone has just reported problems with non-UTF-8 filenames.

To reproduce this with Obnam 1.9 on Debian 8 (jessie):

    $ mkdir data
    $ touch data/"$(printf '\xff')"
    $ obnam backup --no-default-config -r repo sftp://localhost$(pwd)/data
    00h00m00s 1 files 0 B scanned: scanning for files in sftp://localhost/tmp/x/dataTraceback (most recent call last):
      File "/usr/lib/python2.7/dist-packages/cliapp/app.py", line 190, in _run
        self.process_args(args)
    ...
      File "/usr/lib/python2.7/dist-packages/obnamlib/plugins/sftp_plugin.py", line 455, in listdir2
        attrs = self.sftp.listdir_attr(pathname)
      File "/usr/lib/python2.7/dist-packages/paramiko/sftp_client.py", line 208, in listdir_attr
        filename = msg.get_text()
      File "/usr/lib/python2.7/dist-packages/paramiko/message.py", line 199, in get_text
        return u(self.get_bytes(self.get_size()))
      File "/usr/lib/python2.7/dist-packages/paramiko/py3compat.py", line 51, in u
        return s.decode(encoding)
      File "/usr/lib/python2.7/encodings/utf_8.py", line 16, in decode
        return codecs.utf_8_decode(input, errors, True)
    UnicodeDecodeError: 'utf8' codec can't decode byte 0xff in position 0: invalid start byte

What Obnam does is to get the list of files using listdir_attr, and that one fails, if there's a non-UTF8 filename on the remote side.

larswirzenius:
Program to exhibit the problem.

```
#!/usr/bin/python2

import os
import tempfile

import paramiko

filename = os.path.abspath('non-utf-filename-\xff')
with open(filename, 'w') as f:
    f.write('this file has a non-UTF8 filename\n')
print 'Created file with non-UTF8 filename:', repr(filename)

print 'Connecting to localhost over SSH'
remote = ('localhost', 22)
transport = paramiko.Transport(remote)
transport.connect()

print 'Authenticating using keys in SSH agent'
agent = paramiko.Agent()
for key in agent.get_keys():
    try:
[...]

PR #1716, opened 2020-07-13
add optional parameter 'encoding' and 'errors' to most functions in sftp_client.py

hakaishi:
related to #546 #1631 #1365 #1443

This adds encoding and errors for most sftp client functions.

file.py also needed to be adjusted.

Why did I prefer adding the optional parameter "encoding" instead of only adding "errors"?
Well, this way the user can choose the encoding if he wants. If the remote doesn't support utf-8 or the folder includes file names using another encoding than utf-8, you would not be able to work with these files in any way. Only adding encoding will ensure that.
Most users will know the encoding used by the remote system, so I would not see any difficulties there either.

Rewriting the code to only use binary internally would be a rather big change to the code, I guess. So this is why I would prefer to simply add an optional parameter defaulting to the current usage inside paramiko.

I ran some local tests and also pytest to ensure that no code has been broken by my changes.

Best regards,
Hakaishi


aopom:
This feature would be very much appreciated!

hakaishi:
I guess this is too old already. I'll try to merge the changes to my fork in a few days.

PR #1723, opened 2020-07-18
fix RSA key loading: p and q were being swapped

alex:
This currently works, because OpenSSL simply re-computes iqmp when it doesn't match the p & q values. However a future pyca/cryptography patch enforces this.

### DIFF
--- a/paramiko/rsakey.py
+++ b/paramiko/rsakey.py
@@ -189,7 +189,7 @@ class RSAKey(PKey):
             except ValueError as e:
                 raise SSHException(str(e))
         elif pkformat == self._PRIVATE_KEY_FORMAT_OPENSSH:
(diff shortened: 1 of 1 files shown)


alex:
fwiw, even though this currently runs without error, it's causing OpenSSL to constantly recompute iqmp, which makes it much slower. Twisted used to have a similar issue: https://github.com/twisted/twisted/commit/737db21084359e30e339d6cad38dcac487fb5051

bitprophet:
Cherry picked back to Paramiko 2.7. Thanks!

alex:
Thanks Jeff!

nathanielmanistaatgoogle:
I have some work affected by this as well. Excited to see that it is already fixed; is there an estimate for when a version containing the fix will appear on PyPI? 🙂

PR #1724, opened 2020-07-23
handle non-ascii usernames in _parse_userauth_request

graingert:


graingert:
Imho log lines should use the lazy `_log(DEBUG, "%(kw)r", {"kw": kw})` form

graingert:
also you can do this to lazily evaluate repr/formats when using modern f-string/.format:

```python

class _LazyLog(object):
    def __init__(self, fn):
        self.__fn = fn

    def __str__(self):  # python log system calls str(msg) just before logging it
        return self.__fn()


class Transport:
    ...
    def _log(self, level, fn, **kwargs):
        return self.logger.log(level, _LazyLog(fn), **kwargs)
```

then you can use it like this on py3:


```python
    self._log(
        DEBUG,
        lambda: f"Auth request ({method=}) {service=},{username=}",
        extra={"eg": "eg},
    )
```

or


```python
    self._log(
        DEBUG,
        lambda: "Auth request (method={method!r}) service={service!r},username={username!r}".format(
            method=method,
            service=service,
            username=username,
        ),
        extra={"eg": "eg},
    )
```

on py2

graingert:
@bitprophet ^

PR #1733, opened 2020-08-07
Demo of a server providing a shell

teeks99:
A full-functional example of a OS-like shell that a user can modify for their needs.

### DIFF
--- a/demos/demo_console_server.py
+++ b/demos/demo_console_server.py
@@ -1,0 +1,479 @@
+#!/usr/bin/env python
+
+# Copyright (C) 2020 Thomas Kent <tom@teeks99.com>
+#
+# This file is part of paramiko.
+#
+# Paramiko is free software; you can redistribute it and/or modify it under the
+# terms of the GNU Lesser General Public License as published by the Free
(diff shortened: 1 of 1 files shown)


goldstar611:
I haven't tested the code but this would be a great start for a DIY honeypot.

teeks99:
It would probably be a pretty naive honey pot, I strongly suspect there are aspects of the initial parts of the connection that differ from most SSH servers that an attacker could easily check for. 

I put this together as a part of a test harness for a client we have that needs to login to a server. This is far easier to setup and much more focused, with better telemetry coming out, than setting up a real service to login to.

goldstar611:
> I strongly suspect there are aspects of the initial parts of the connection that differ from most SSH servers that an attacker could easily check for.

Of course :) But if you're hosting an "internal" honey pot then all you might need is just to know that somebody connected, their IP and if they tried to authenticate with a key or password. Then you've got some good info to send to Info Sec ;)

hejiazhen1203:
> A full-functional example of a OS-like shell that a user can modify for their needs.

I want some help，how could  change the SimpleShell to /bin/bash , i want to get a terminal with shell after login.

PR #1781, opened 2020-11-25
Debug get-function of sftp_client.py

wolfgangmichaelkraus:
The get-function of the 'sftp_client.py' was being modified using file-object based approach in order to avoid size-mismatching in get-chains (for example using 'pysftp.get_d').

'IOError' was being thrown due to synchronicity issues between Python interpreter and operating-system using `os.stat(localpath)` to gather file-size of local file.

Commits tested in local OSX-Docker environment (macOS Catalina version 10.15.6 and Docker Engine version 19.03.13) using Python 3.9 image accessing remote SFTP-server to get files without any issue.

### DIFF
--- a/paramiko/sftp_client.py
+++ b/paramiko/sftp_client.py
@@ -800,10 +800,10 @@ class SFTPClient(BaseSFTP, ClosingContextManager):
         """
         with open(localpath, "wb") as fl:
             size = self.getfo(remotepath, fl, callback)
-        s = os.stat(localpath)
-        if s.st_size != size:
+            localsize = fl.tell()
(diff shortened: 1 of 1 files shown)


wolfgangmichaelkraus:
It does seem that Travis CI is failing due to a general issue regarding config for pypy and pypy3 and I cannot see how this is related to my pull request.

Both failed checks indicating the following error:
`ImportError while loading conftest '/home/travis/build/paramiko/paramiko/tests/conftest.py'.`

I'm not quite sure how to fix this and hope that someone could support with this issue.

PR #1785, opened 2020-12-03
Convert hard-coded block size to constant

### LINKED ISSUE #1784: Terrible performance of getfo and putfo with certain file-like objects
I am using Paramiko in conjunction with [S3Path](https://github.com/liormizr/s3path) to copy files from a SFTP server to an S3 bucket. (S3Path implements Python's pathlib interface for S3 buckets.) In theory, this is an excellent combination where I can simply do

```python
destination_file_handle = s3_object.open("wb")
sftp.getfo(sftp_source_file_location, destination_file_handler)
```
In practice, performance begins well but gradually decreases to about 50 kb/s. I would expect performance at least on the order of 5 MB/s, so 100× faster.

Investigation shows that the problem lies with [a hard-coded small block size of 32768](https://github.com/paramiko/paramiko/blob/677166b4411e86029226f3d093701fb3b999a082/paramiko/sftp_client.py#L678). This leads to a huge number of write operations, and presumably I'm being throttled by AWS as a result.

While for many purposes the 32kb block size may work well, I think one should be allowed to adjust it if needed.

With the below pull-request, the block size is no longer hard-coded in the function. This makes the following workaround possible:
```python
import paramiko.sftp_client as sftp_client_module
sftp_client_module.SFTP_FILE_OBJECT_BLOCK_SIZE = 2 ** 23  # 8 MiB
```
Then I get the desired 5MB/s transfer rate.

maresb:
Closes #1784


maresb:
The errors on the failed pypy Travis CI builds are:

> E   RuntimeError: You are linking against OpenSSL 1.0.2, which is no longer supported by the OpenSSL project. To use this version of cryptography you need to upgrade to a newer version of OpenSSL. For this version only you can also set the environment variable CRYPTOGRAPHY_ALLOW_OPENSSL_102 to allow OpenSSL 1.0.2.

I doubt this has anything to do with my PR.

PR #1792, opened 2020-12-24
Fix typo in demo scripts

ksaifullah:


demos/demo_sftp.py
```diff
@@ -79,7 +79,7 @@
     try:
         # try ~/ssh/ too, because windows can't have a folder named ~/.ssh/
         host_keys = paramiko.util.load_host_keys(
-            os.path.expanduser("~/ssh/known_hosts")
+            os.path.expanduser("~/.ssh/known_hosts")
```
ploxiln, line 82: this path is attempted above on line 76, the explanation for trying again without this dot is on line 80

PR #1806, opened 2021-01-30
Remove pynacl dependency.

adiroiban:
Scope
=====

The pynacl is only used for ED keys.

cryptography starting with 2.6 supports ED signing.

I think that by removing the pynacl and libsodium deps, it will be much easier to deploy paramiko.
I saw a few issues about failing to use paramiko due to the pynacl deps.

Changes
=======

Update requirements from cryptography 2.5 to 2.6 for ED support.

I think that bcrypt will remain a main deps as it is used in the new OpenSSH v1 private key format.

I kept the ed25519 extra requirement for backward compatibility.

Use cryptography to sign and validate SSH data.

Update the automated tests to include a quick sign / verify test.

Update the www docs to remove `paramiko[ed25519]` or pynacl deps.
Update bcrypt to informa that it's used for OpenSSH v1 private key format.

How to test
========

In the dev environment install cryptography 2.6 and make sure pynacl is removed and run the tests

```
pip install -r dev-requirements.txt
pip install cryptography==2.6
pip uninstall pynacl
invoke test
flake8
```

Check extra requirement still works for backward compatibility:

```
pip install .[ed25519]
```

### DIFF
--- a/paramiko/ed25519key.py
+++ b/paramiko/ed25519key.py
@@ -16,10 +16,12 @@
 
 import bcrypt
 
+from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric import ed25519
 from cryptography.hazmat.primitives.ciphers import Cipher
 
-import nacl.signing
 
 from paramiko.message import Message
 from paramiko.pkey import PKey, OPENSSH_AUTH_MAGIC, _unpad_openssh
@@ -29,7 +31,11 @@ from paramiko.ssh_exception import SSHException, PasswordRequiredException
 
 class Ed25519Key(PKey):
     """
-    Representation of an `Ed25519 <https://ed25519.cr.yp.to/>`_ key.
+    Representation of an `Ed25519 <https://ed25519.cr.yp.to/>`_ public or
+    private key.
+
+    When `msg` or `data` arguments are provided for the constructor, a
+    public key is generated.
 
     .. note::
         Ed25519 key support was added to OpenSSH in version 6.5.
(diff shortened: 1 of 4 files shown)


adiroiban:
Tests are failing, but I got the same failures when I run `inv sites` on master.

I think that paramiko automated tests should be migrated to GitHub Actions first.

Let me know if you need help with the GitHub Actions migration.

ash-cryptocom:
Any updates? Because we are trying to install paramiko with ARM64 chip but encountered issue with `pynacl`: https://github.com/pyca/pynacl/issues/654

adiroiban:
@ash-cryptocom you can try to the patch and see if it works for you.
I am using a forked paramiko with this patch and I have not observed any issues... but I am using paramiko only as tool to support the testing process and not for production.

jzaleski:
@ash-cryptocom / @adiroiban I can confirm success w/ this patch.

---

In addition to the patch I needed to get `cffi` to build the correct native-extension for my M1 mac

**You can confirm you're having the same issue with the following:**
```
> file .venv/lib/python3.9/site-packages/_cffi_backend.cpython-39-darwin.so
.venv/lib/python3.9/site-packages/_cffi_backend.cpython-39-darwin.so: Mach-O 64-bit bundle x86_64
```

**To fix the issue run the following:**
```
> pip[3] uninstall cffi -y && arch -arm64 pip install cffi --no-binary :all:
```

**Verify the native-extension architecture** (_it should be_ `arm64`):
```
> file .venv/lib/python3.9/site-packages/_cffi_backend.cpython-39-darwin.so
.venv/lib/python3.9/site-packages/_cffi_backend.cpython-39-darwin.so: Mach-O 64-bit bundle arm64
```

I hope this helps.

FelixSchwarz:
Probably I'm misunderstanding something but I think the minimum version of `cryptography` should be 2.7 as that is the first release which contains pyca/cryptography@a453d8709763badde325568c13961bb6745f333c.

ploxiln:
This doesn't use `cryptography` for key serialization, it only uses the ed25519 signing functions added in cryptography-2.6

ploxiln:
(... more accurately: this uses Raw serialization, not openssh format serialization, from cryptography, so cryptography version 2.6 should be sufficient)

PR #1814, opened 2021-02-11
Update .travis.yml

Jpfonseca:
Refer to #1813

Needed by #1811


Jpfonseca:
@bitprophet  you this can be an temporary fix until you move to Circle CI. I believe this build will finish correctly

PR #1820, opened 2021-02-28
paramiko/client.py: Fix invoke_shell to work with environment

jwessel:
When calling invoke_shell() with environment set to something
other than None it is not passed on to the ssh server.

The invoke_shell should pass the environment the same
manner as exec_command().

Signed-off-by: Jason Wessel <jason.wessel@windriver.com>


bskinn:
Closing in favor of the most recent #2129

Issue #418, opened 2014-10-10
Dropbear and Paramiko SSH server


slaff:
Latest versions, beyond version 2013.57, of the dropbear ssh client do not work with paramiko SSH server. Most probably the following patch from libssh (http://www.libssh.org/archive/libssh/2014-03/0000044.html) can be used as a base for a patch for the kex negotiation.

manfred-kaiser:
I tested a more recent version with the latest version of paramiko, but the client was not able to connect.

**Used Versions:**

* Dropbear client: v2017.75
* Paramiko: 2.7.2

**Paramiko Exception message:**

I got an exception that that packet 21 is expected, but got 30.

```
[transport.py:1819]  Exception: Expecting packet from (21,), got 30
[transport.py:1817]  Traceback (most recent call last):
[transport.py:1817]    File "/pythonenv/lib/python3.6/site-packages/paramiko/transport.py", line 2070, in run
[transport.py:1817]      self._expected_packet, ptype
[transport.py:1817]  paramiko.ssh_exception.SSHException: Expecting packet from (21,), got 30
```

**Dropbear client:**

`first_kex_packet_follows = True`


Kex sent by the dropbear: `['curve25519-sha256@libssh.org', 'ecdh-sha2-nistp521', 'ecdh-sha2-nistp384', 'ecdh-sha2-nistp256', 'diffie-hellman-group14-sha1', 'diffie-hellman-group1-sha1', 'kexguess2@matt.ucc.asn.au']`


**OpenSSH Client**

`first_kex_packet_follows = True`

Kex sent by openssh: `['curve25519-sha256', 'curve25519-sha256@libssh.org', 'ecdh-sha2-nistp256', 'ecdh-sha2-nistp384', 'ecdh-sha2-nistp521', 'diffie-hellman-group-exchange-sha256', 'diffie-hellman-group16-sha512', 'diffie-hellman-group18-sha512', 'diffie-hellman-group-exchange-sha1', 'diffie-hellman-group14-sha256', 'diffie-hellman-group14-sha1', 'ext-info-c']`

manfred-kaiser:
I have created a draft pull request #1823 because it needs more investigation an more checks

manfred-kaiser:
as stated in rcf 4253:

```
After receiving the SSH_MSG_KEXINIT packet from the other side,
 each party will know whether their guess was right.  If the
 other party's guess was wrong, and this field was TRUE, the
[...]

PR #1823, opened 2021-03-05
workaround for dropbear clients

manfred-kaiser:
added workaround for dropbear clients as mentioned in #418 

This is only a first test, but I created this pull request for others to follow my work and help to improve

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -2065,6 +2065,11 @@ class Transport(threading.Thread, ClosingContextManager):
                         continue
                     if len(self._expected_packet) > 0:
                         if ptype not in self._expected_packet:
+                            if ptype == 30:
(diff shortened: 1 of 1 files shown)


paramiko/transport.py
```diff
@@ -2065,6 +2065,11 @@ def run(self):
                         continue
                     if len(self._expected_packet) > 0:
                         if ptype not in self._expected_packet:
+                            if ptype == 30:
```
ploxiln, line 2068: there should probably be a name defined for this in `common.py`, just like `MSG_DISCONNECT`, `MSG_IGNORE`, `MSG_DEBUG` etc

manfred-kaiser: I have added MSG_KEXSKIP. I have to check, if 30 is the right value but I can't find information in the rfc.

Issue #1509, opened 2019-09-18
Agent support for OpenSSH on Windows (Powershell/OpenSSH)


benj56:
The OpenSSH that comes with Windows is not supported.
In `agent.py`, if the platform is Windows, it checks for a running Putty `pageant` agent, if not found it fails. 

Is there any chance this can be implemented? I believe the OpenSSH agent on Windows communicates via a named pipe, while `pageant` seems to use a mapped file. 

I would need this for agent forwarding.

maxslimmer:
It does use named pipes. I have a decent start on this. There are some parts missing (such as detecting the openssh agent is running, error handling) and it currently depends on pywin32. Unsure where to go from here, do I submit a PR?

bitprophet:
Let's roll this into #1868.

PR #1837, opened 2021-03-29
feat(agent): support Windows OpenSSH

lewgordon:
Addresses #1509


kdabci:
You can totally simplify this by simply using open(), read() and write() on a common file object instead of the win32 api.

sengels:
You can simplify this by using file api (open, close, read, write) instead of win32 api, as the underlying filesystem will handle this.

bitprophet:
Let's consolidate this into #1868

PR #1846, opened 2021-04-26
Make prefetching optional in get() and getfo()

h3ll0r:
In some cases prefetching can lead to problems when downloading larger files. The connection to the SFTP server is disconnected due to the query behavior. Proprietary implementations in particular tend to behave that way. Switching off prefetching is a slow but resilient solution to this problem.

### DIFF
--- a/tests/test_sftp.py
+++ b/tests/test_sftp.py
@@ -560,25 +560,27 @@ class TestSFTP(object):
         using a lager file.
         """
 
+        sftp_filename = sftp.FOLDER + "/dummy_file"
+        num_chars = 1024 * 1024 * 4
+
         fd, localname = mkstemp()
         os.close(fd)
 
         with open(localname, 'wb') as f:
-            num_chars = 1024 * 1024 * 4
             f.write(b'0' * num_chars)
 
-        sftp.put(localname, sftp.FOLDER + "/dummy_file")
+        sftp.put(localname, sftp_filename)
 
         os.unlink(localname)
         fd, localname = mkstemp()
         os.close(fd)
 
-        sftp.get(sftp.FOLDER + "/dummy_file", localname, prefetch=False)
+        sftp.get(sftp_filename, localname, prefetch=False)
 
-        assert os.stat(localname).st_size == 4194304
+        assert os.stat(localname).st_size == num_chars
 
         os.unlink(localname)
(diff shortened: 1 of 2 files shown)


bskinn:
Thanks for the PR! Could you add a test or two to the PR that exercises the option?

KamilBabayev (approved):

h3ll0r:
> Thanks for the PR! Could you add a test or two to the PR that exercises the option?

Done! Just a small a small test to ensure that the get function works properly. At the moment I cannot deliver a server configuration to to simulate the error. If I can, I would deliver this test later.

tests/test_sftp.py
```diff
@@ -554,6 +554,32 @@ def progress_callback(x, y):
...
+
+        sftp.put(localname, sftp.FOLDER + "/dummy_file")
+
+        os.unlink(localname)
+        fd, localname = mkstemp()
+        os.close(fd)
+
+        sftp.get(sftp.FOLDER + "/dummy_file", localname, prefetch=False)
+
+        assert os.stat(localname).st_size == 4194304
```
bskinn, line 578: I'm not expert here, but looks good to me.

For readability, and to avoid the magic numbers/strings, maybe refactor:

1) The number of characters to e.g. `NUM_CHARS` at the test function scope, and then use it both with the `f.write(b'0' * NUM_CHARS)` above and also in this assertion for the file size; and,

2) The `sftp.FOLDER + "/dummy_file"` to e.g. `STFP_FILENAME` at test function scope?

h3ll0r: > I'm not expert here, but looks good to me.
> 
> For readability, and to avoid the magic numbers/strings, maybe refactor:
> 
> 1. The number of characters to e.g. `NUM_CHARS` at the test function scope, and then use it both with the `f.write(b'0' * NUM_CHARS)` above and also in this assertion for the file size; and,
> 2. The `sftp.FOLDER + "/dummy_file"` to e.g. `STFP_FILENAME` at test function scope?

me neither but I hope with this update I was able to adjust the readability according to your suggestions

bskinn: Yep! That looks good to me. I've flagged the PR for @bitprophet's attention; he'll be the one to set the plan for a merge &c.

Thanks for the contribution!

bitprophet:
Thank you! All merged.

PR #1849, opened 2021-05-03
[SFTP Client] Expose server version and extensions

Tomasito665:
The SFTP client supports both regular SSH_FXP_RENAME * through SFTPClient.rename and the extended "posix-rename@openssh.com" command through SFTPClient.posix_rename. However, the client does not provide any way of telling whether or not the server supports the "posix-rename@openssh.com" extension.

This pull request proposes to parse the extension pairs that the SFTP server returns in the SSH_FXP_VERSION * packet during protocol initialization and expose the result with a new method: `SFTPClient.get_server_extensions`. The result of calling this method is a dictionary, where each dictionary item represents an extension pair:

- key (Unicode string) -> extension_name
- value (Unicode string) -> extension_data

With this, checking if a server implements the POSIX rename extension becomes as easy as doing: "posix-rename@openssh.com" in cli.get_server_extensions().

For the sake of completeness, this pull request also adds an `SFTPClient.get_server_version` method, exposing the server version, which we already parsed but did not expose yet.

\* Terminology borrowed from the "SSH File Transfer Protocol" document:
https://filezilla-project.org/specs/draft-ietf-secsh-filexfer-02.txt


Tomasito665:
Ping! :-) 

Would anyone have time to have a look at this PR, please? 
@bitprophet maybe?

Thanks

Tomasito665:
Adding a friendly ping :-)

Would anyone have time to have a look at this PR, please?
My team at my previous company still needs this and was asking for it.

Thanks

PR #1857, opened 2021-05-25
Mux transport for client (rebased)

takluyver:
This is a straight rebase of #1341, which no longer merges cleanly. So far, I've not attempted to make any changes to it.

I wonder if it would be better, rather than adding another parameter to `Client.connect()` which causes it to take a completely different code path, to add a separate method, called something like `connect_multiplex()` or `connect_master()`, which just does this. Then it can be up to the caller whether to fall back to standard connection if using the multiplexing socket fails (whereas the current implementation silences errors and always tries to fall back).

(In principle, I'd also like to disentangle the code in Transport in a similar way, but it's less clear how to do that)

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -340,7 +340,8 @@ class SSHClient(ClosingContextManager):
         if controlpath:
             try:
                 t = self._transport = Transport(
-                    (hostname, port), controlpath=controlpath)
+                    (hostname, port), controlpath=controlpath
+                )
                 # If this connection succeeds, most (but not all) of the
                 # SSHClient connect steps can be bypassed
                 if self._log_channel is not None:
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -237,6 +237,7 @@ class SSHClient(ClosingContextManager):
         gss_trust_dns=True,
         passphrase=None,
         disabled_algorithms=None,
+        controlpath=None,
     ):
         """
         Connect to an SSH server and authenticate to it.  The server's host key
@@ -314,6 +315,8 @@ class SSHClient(ClosingContextManager):
         :param dict disabled_algorithms:
             an optional dict passed directly to `.Transport` and its keyword
             argument of the same name.
+        :param string controlpath: path to unix domain socket of running
+            ControlMaster multiplexing agent.
 
         :raises:
             `.BadHostKeyException` -- if the server's host key could not be
@@ -334,6 +337,19 @@ class SSHClient(ClosingContextManager):
         .. versionchanged:: 2.6
             Added the ``disabled_algorithms`` argument.
         """
+        if controlpath:
+            try:
+                t = self._transport = Transport(
+                    (hostname, port), controlpath=controlpath)
+                # If this connection succeeds, most (but not all) of the
+                # SSHClient connect steps can be bypassed
+                if self._log_channel is not None:
+                    t.set_log_channel(self._log_channel)
+                t.start_client(timeout=timeout)
+                return
+            except SSHException as e:
+                # Quietly fallback to traditional connection steps
+                pass
         if not sock:
             errors = {}
             # Try multiple possible address families (e.g. IPv4 vs IPv6)
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -348,7 +348,7 @@ class SSHClient(ClosingContextManager):
                     t.set_log_channel(self._log_channel)
                 t.start_client(timeout=timeout)
                 return
-            except SSHException as e:
+            except SSHException:
                 # Quietly fallback to traditional connection steps
                 pass
         if not sock:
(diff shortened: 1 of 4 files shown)


oferchen:
what is the delay with merging this pull?

thank you!

bskinn:
The failing CI, for one -- if I'm looking at the Travis run output correctly, there are some things that need to be fixed in the CI infrastructure/config first.

oferchen:
> The failing CI, for one -- if I'm looking at the Travis run output correctly, there are some things that need to be fixed in the CI infrastructure/config first.
Yeah I saw the failure log as well, Can you share who's responsible for the CI configuration?

oferchen:
> I believe all of that is still on @bitprophet's shoulders -- he's dividing his weekly OSS time among all of paramiko, fabric and invoke, so it can sometimes take a while for any given thing to get addressed.
> 
> It looks to me like there's more than one problem underlying this CI failure -- partly, Python 3.4 might need to be dropped; partly, there have been some ongoing issues with getting `setuptools_rust` and `cryptography` to play well again within CI. I know he's already aware of the `setuptools_rust` and `cryptography` problems, at least.

Hi @bitprophet ,
I don't mean to bother you but this feature was initially requested in 2016 and there are multiple issues open across paramiko, fabric and several other libraries. If I can assist with pushing this issue forward due to limited time to spend on it please let me know.

Thank you for your efforts,
Ofer

FYI if anyone want to test this pull request you can use the following,
[paramiko-controlmaster-16062021.tar.gz](https://github.com/paramiko/paramiko/files/6687321/paramiko-controlmaster-16062021.tar.gz)

bitprophet:
Either during or after the CI work, 3.4 and 3.5 will be dropped along with 2: http://bitprophet.org/blog/2021/02/25/byethon2/#how-exactly

I've put this ticket in a milestone so that ideally I'll revisit it once I'm ready to push forwards with feature work. Thanks!

ltalirz:
Hey guys, another highly interested party here :-)

As I notice that CI on PRs that came after this one is passing, may I suggest to simply try rebasing once more @takluyver ?

If there's anything others can do to help get this merged, please let us know!

ltalirz:
After a simply style fix, [tests are passing](https://github.com/paramiko/paramiko/pull/1931/checks?check_run_id=4306169156).

Note: I had to do the fix manually, since the black installed in my python 3.9 environment ended up making more changes and I was unable to reproduce the python 3.6 environment in which the CI runs.

Given that python 3.6 is EOL end of this year, I'll open another PR that moves the style checking etc. to python 3.7.

bskinn:
Perhaps updating to an even more recent version would be better?

Thoughts, @bitprophet?

bitprophet:
God, I didn't even realize 3.6 was that close to EOL. I'll be over here growing moss...

But no, the point in my mind of the "baseline CI" tasks are to use the oldest supported version since usually things that work on on version N will work identically well on N+1, N+2 etc. So bumping up to 3.7 seems like the correct choice right now.

With the downside that someone has to submit a version of that 2nd PR every year or so - tho I'm hoping to eventually leverage Circle's templatey junk so the pipeline definitions are more DRY (esp given I have one of those for every project).

----

Unless my US Thanksgiving goes very sour this year, I'm planning to work on Paramiko this Friday and some of the following week - hopefully I'll get to this ticket in the milestone. Thanks all for continuing to keep it going.

bitprophet:
A bit later: actually, my dayjob (which pays me to work on the OSS a day a week) has a /lot/ of users on Python 3.6, which undoubtedly means a large number of the greater ecosystem does too. So I'm curious what specifically the problem was for @ltalirz re:

> I was unable to reproduce the python 3.6 environment in which the CI runs.

since my general thinking is that I'm fine with dropping interpreter versions after they reach EOL, *but* only when the pain of supporting them outweighs some arbitrary userbase-% (I haven't looked at the download stats lately).

In my own personal and dayjob environments, 3.6 is still trivially accessible & _so far_ the libraries/tools I use haven't themselves become 3.6-incompatible. (with the caveat that a lot of said dependencies tend to get pinned & will shortly need re-examination…)

Also! CircleCI has tooling that makes it easy to run local containers matching what they run, which can be a big help with this kind of thing. I'm still getting my NixOS system stood up with containers & such but IIRC it was basically "install circleci's CLI tool, make sure Docker or Podman-aliased-to-Docker exist, and run `circleci local name-of-job` or similar.

ltalirz:
> So I'm curious what specifically the problem was for @ltalirz re:

I first installed the dev dependencies in a python 3.9 conda environment. That worked fine, but the version of `black` I got was probably slightly different from the one you have on CI under python 3.6, so it ended up making more changes than those requested on CI (which is a bit weird now that I see that the version is quite locked down... maybe some black dependency that differs?).

I then created a python 3.6 conda env, and here installing the dev dependencies didn't work because it was trying to change the version of some "system-level package" (don't remember which one) which it could not.

Anyhow, I guess you're right that I could have tried using a container - to me the update to python3.7 is up to you, I care much more about the SSH multiplexing support in this PR (or https://github.com/paramiko/paramiko/pull/1932).

Cheers!

bitprophet:
Rolling into #1931 and its chat.

Issue #822, opened 2016-09-30
Key-exchange timeout over session channels, but not sftp


mathrick:
We're running into the key-exchange timeout exception when talking to OpenSSH, but it only happens when using a "session" channel. SFTP transfers proceed just fine, and we see successful key exchange in the logs.

[This gist](https://gist.github.com/mathrick/1b2ead63553f2bfc8d3e129e1d4512e2) demonstrates the issue:
```
$ ssh hostname "base64 /dev/urandom | head -c 200000 > /tmp/file.txt"
$ python test_paramiko.py quicksilver -u mkatafiasz /tmp/file.txt --sftp
REKEY_BYTES: 102400
REKEY_PACKETS: 536870912
Using SFTPClient, which should succeed in rekeying
Read file 1 times
Read file 2 times
Read file 3 times
Read file 4 times
Read file 5 times
^C
KeyboardInterrupt

$ python test_paramiko.py quicksilver -u mkatafiasz /tmp/file.txt
REKEY_BYTES: 102400
REKEY_PACKETS: 536870912
Using SSHClient, which should fail in rekeying
Need rekey
Read file 1 times
Read file 2 times
Traceback (most recent call last):
No handlers could be found for logger "paramiko.transport"
  File "test_paramiko.py", line 45, in <module>
    for line in stdout:
  File "/usr/lib/python2.7/dist-packages/paramiko/file.py", line 106, in next
    line = self.readline()
  File "/usr/lib/python2.7/dist-packages/paramiko/file.py", line 214, in readline
    new_data = self._read(n)
  File "/usr/lib/python2.7/dist-packages/paramiko/channel.py", line 1253, in _read
    return self.channel.recv(size)
  File "/usr/lib/python2.7/dist-packages/paramiko/channel.py", line 628, in recv
    self.transport._send_user_message(m)
  File "/usr/lib/python2.7/dist-packages/paramiko/transport.py", line 1442, in _send_user_message
    raise SSHException('Key-exchange timed out waiting for key negotiation')
paramiko.SSHException: Key-exchange timed out waiting for key negotiation
```

The remote side is running OpenSSH (tried with 6.6 and 6.7), and setting `LogLevel DEBUG` yields the following output:

* SFTP (successfully rekeys)
```
Sep 27 17:48:42 myhost sshd[20158]: debug1: SSH2_MSG_KEXINIT received
Sep 27 17:48:42 myhost sshd[20158]: debug1: SSH2_MSG_KEXINIT sent
[...]

PR #1860, opened 2021-06-03
fix ssh-key-exchange thread locking

danielino:
fix issue #822 

adds a timeout for read_message. a default value of 0 means that timeout is not evaluated.

### DIFF
--- a/paramiko/packet.py
+++ b/paramiko/packet.py
@@ -276,7 +276,7 @@ class Packetizer(object):
             self.__timer_expired = False
             self.__handshake_complete = True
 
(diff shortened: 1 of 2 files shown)


ploxiln:
hmm ... between `packetizer.start_handshake(handshake_timeout)` and `packetizer.complete_handshake()`, the packetizer has a timer which it checks in every loop of `read_all()` by calling `self.handshake_timed_out()`. That seems redundant with this timeout you're adding. Can you figure out what that handshake timeout isn't working in your case?

PR #1868, opened 2021-06-17
feat(agent): support Windows OpenSSH

sengels:
Addresses #1509

### DIFF
--- a/paramiko/agent.py
+++ b/paramiko/agent.py
@@ -246,7 +246,12 @@ class AgentClientProxy(object):
             if win_pageant.can_talk_to_agent():
                 conn = win_pageant.PageantConnection()
             else:
-                return
+                import paramiko.win_openssh as win_openssh
+
+                if win_openssh.can_talk_to_agent():
+                    conn = win_openssh.OpenSSHAgentConnection()
+                else:
+                    return
         else:
             # no agent support
             return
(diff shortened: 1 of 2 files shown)


DaviidC:
Will this ever be merged? 
Is Windows OpenSSH `ssh-agent` support planned?

sengels:
I don't know, I've just been bitten by that again.

bitprophet:
Seems ok at a very brief glance, but it at least needs a changelog entry and ideally some basic testing, if plausible (I acknowledge that testing this kind of lower level OS stuff can get tricky, but maybe see what can be easily mocked?)

Can you also speak to #1837 (which looks like this was based on)?

I'm also curious if you've been able to test this on a system that supports both the newer native SSH _and_ WinPageant? I'm thinking of users who might be using the latter for a long time (but are on a modern enough Windows that the new SSH is around) and might end up bitten by a sudden switch in behavior.

sengels:
Yes, this commit is just a slight improvement over #1837 in that it doesn't use the windows specific api.

I can easily test this with both openssh and Pageant; I can try to come up with a simple test case, but I have no idea how to do mock testing in that regard?

wrt. changelog, can you let me know where I can put a suitable line?

sengels:
@bitprophet I fixed the lint issues, and added a versionchanged line.
I reworked the order in which the native openssh is used, so now this is backwards compatible in that putty's pageant is tried first and only if that cannot be found, the native version is used.
I see one test error in 3.6 but I have no idea why this should be the case. Maybe you have more insight?

bitprophet:
Thank you! and if that is the errors I've hit, it's some ephemeral problem on CircleCI that can, sadly, be ignored. I'll look into it.

bitprophet:
This'll go out in 2.10, hopefully today 🤞🏻

PR #1883, opened 2021-07-21
Not all hash algorithms are sha1

### LINKED ISSUE #1882: All server-side, non-group-exchange Diffie-Hellman algorithms are using SHA1 even when they shouldn't
All the non-group-exchange (i.e. `diffie-hellman-group{1,14,16}-*`) Diffie-Hellman algorithms derive from the KexGroup1 implementation in [paramiko/kex_group1.py](https://github.com/paramiko/paramiko/blob/master/paramiko/kex_group1.py). To support other digest algorithms, KexGroup1 has a `hash_algo` class variable which defaults to `sha1`. In the `_parse_kexdh_reply` method (used when we are the client side of the exchange), this is correctly used like so to generate the hash:

```
        self.transport._set_K_H(K, self.hash_algo(hm.asbytes()).digest())
```

Thus allowing the sha256 and sha512 variants to substitute those algorithms. Unfortunately in `_parse_kexdh_init` (used when we are the server side of the exchange) we have the sha1 algorithm hardcoded rather than using `self.hash_algo`:

```
        H = sha1(hm.asbytes()).digest()
```

This should almost certainly be:

```
        H = self.hash_algo(hm.asbytes()).digest()
```

jeamland:
Fixes #1882


bitprophet:
Cherry picked back to 2.5 on up!

Issue #1660, opened 2020-04-08
SFTP file transfer speed regression since 2.5.0 onwards


Jamie-:
I used Paramiko (as part of fabric) nightly to copy multiple large files over SSH to a machine on the same local network (1Gbps link). Taking a 7GB file as an example it used to take around 2 minutes to copy with Paramiko 2.4.1. I upgraded to 2.6.0, copying the same file now takes around 17 minutes. I did some digging and this is what I found.

My test setup was as follows:
* Linux workstation with SSD
* Docker container running Python 2 with Paramiko installed during build, mounting my home directory
* Copying files over SSH from docker container mounted dir to host's `/tmp`
* 3.1GB .tgz file (this is important as it is **compressed**, I've seen old GH issues where compression seemed to make a difference)

I went with docker for this as I could pin requirements knowing for certain old versions on my host wasn't affecting it. To change Paramiko version I'd update my requirements and rebuild the container. Not overly scientific but seemed repeatable enough. As a note, I'm using Python 2 because my nightly copy code is a monster code base all currently in Python 2 and progress migrating is very slow. I have not done any testing in 3, but given this is a regression I'm not so sure it matters.

Again, all numbers are copying from container mounted dir -> host over SSH to `/tmp`. To measure time I had `time.time()` before `sftp_client.put()` and again after then subtracted the difference.

`scp`: 10s
paramiko==2.4.1: 32s (+/- 1s with cryptography==1.7.1, 2.2.2, 2.8)
paramiko==2.4.3: 33s
paramiko==2.5.0: 280s
paramiko==2.6.0: 284s
paramiko==2.7.1: 284s

After testing 2.4.1 with 3 different cryptography versions I settled on 2.8 for the rest given how little difference it made.

As you can see between 2.4.3 and 2.5.0 there's a **very noticeable speed degradation**: ~30s to ~5m. And this is to all extents and purposes on the same machine.

I've searched lots of open bugs here, some dating back a good few years, and on StackOverflow and it seems most slowness people have seen is related to `default_window_size` and `default_max_packet_size`. However on the 2.4 branch I see this:
https://github.com/paramiko/paramiko/blob/2.4/paramiko/common.py#L226-L227

And on the 2.5 branch I see this:
https://github.com/paramiko/paramiko/blob/2.5/paramiko/common.py#L233-L234

The same on both. So I've concluded it's not that. Besides, those values (from my Googling) appear to be the same values used by OpenSSH.

[...]

PR #1884, opened 2021-07-21
Speed up SFTP transfer speed by reverting commit 7c1d29c

jzombi:
This reverts commit 7c1d29c ("Leverage b() in asbytes() to avoid duplication.")

Using exception handling for control flow in `common.asbytes()` (introduced in 7c1d29c) resulted in degraded performance. Changing back to the original code fixes the performance issue.

See #1660

### DIFF
--- a/paramiko/common.py
+++ b/paramiko/common.py
@@ -20,7 +20,7 @@
 Common constants and global variables.
 """
 import logging
-from paramiko.py3compat import byte_chr, PY2, long, b
+from paramiko.py3compat import byte_chr, PY2, long, bytes_types, text_type
 
 (
     MSG_DISCONNECT,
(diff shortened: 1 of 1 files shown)


iorijjw:
> This reverts commit [7c1d29c](https://github.com/paramiko/paramiko/commit/7c1d29c97c2e4861953706c8d0197ad399178b86) ("Leverage b() in asbytes() to avoid duplication.")
> 
> Using exception handling for control flow in `common.asbytes()` (introduced in [7c1d29c](https://github.com/paramiko/paramiko/commit/7c1d29c97c2e4861953706c8d0197ad399178b86)) resulted in degraded performance. Changing back to the original code fixes the performance issue.
> 
> See #1660

After revert 7c1d29c, no significant improvement was found after testing.

PR #1886, opened 2021-07-27
refactor: use absolute imports

kuwv:
This PR is intended to help improve the security of applications using paramiko by using explicit paths.

Relative paths can potentially be exploited my malicious libraries. This coupled with the potential use of sudo capabilities can allow for privilege escalation.

This is due to the module cache being world writable within CPython.

### DIFF
--- a/paramiko/agent.py
+++ b/paramiko/agent.py
@@ -367,7 +367,7 @@ class Agent(AgentSSH):
                 # probably a dangling env var: the ssh agent is gone
                 return
         elif sys.platform == "win32":
-            from . import win_pageant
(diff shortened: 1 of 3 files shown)


MrMino:
> Relative paths can potentially be exploited my malicious libraries.

Could you please elaborate on how? I'm not sure if Python import system was ever meant to be secure the way I think you want it to.

kuwv:
@MrMino I do have an example of using `importlib` that caused another a `config` from another module to be imported instead. But, it's buried in the history of another project. I'll close this until I can put something together.

PR #1892, opened 2021-08-11
support ssh config include directive

jsiverskog:
this is heavily based on the work by Fabien Di Tore <fabien@ditore.ch>.

this is based on #872 , but has been cleaned up, tested and docs added.

### DIFF
--- a/paramiko/config.py
+++ b/paramiko/config.py
@@ -23,6 +23,7 @@ Configuration file (aka ``ssh_config``) support.
 
 import fnmatch
 import getpass
+import glob
 import os
 import re
 import shlex
@@ -119,7 +120,7 @@ class SSHConfig(object):
         obj.parse(flo)
         return obj
 
-    def parse(self, file_obj):
+    def parse(self, file_obj, parsed_files=None):
         """
         Read an OpenSSH config from the given file object.
 
@@ -162,6 +163,27 @@ class SSHConfig(object):
                 # at the end (for compatibility with issue #415). After 3.x, it
                 # will simply not get stripped, leaving a nice explicit marker.
                 context["config"][key] = None
+            elif key == 'include':
+                # Support for Include directive in ssh_config
+                path = value
+                # According to SSH documentation if the path is relative,
+                # we look forward in ~/.ssh
+                if (not os.path.isabs(path)) and path[0] != '~':
+                    path = os.path.join("~", ".ssh", path)
+                # Expand the user home path
+                path = os.path.expanduser(path)
+                if parsed_files is None:
+                    parsed_files = []
+
+                # Parse every included file
+                for filename in glob.iglob(path):
+                    if os.path.isfile(filename):
+                        if filename in parsed_files:
+                            raise ValueError("Include loop detected in ssh "
+                                             "config file: %s" % filename)
+                        with open(filename) as fd:
+                            parsed_files.append(filename)
+                            self.parse(fd, parsed_files)
             # All other keywords get stored, directly or via append
             else:
                 if value.startswith('"') and value.endswith('"'):
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -990,3 +990,30 @@ class TestComplexMatching(object):
         # !canonical in a config that is canonicalized - does NOT match
         result = load_config("match-canonical-yes").lookup("www")
         assert result["user"] == "hidden"
+
+
+class TestSSHConfigInclude(object):
+    def setup(self):
+        config = """
+Include {basic}
+
+Host example.com
+    User exampleuser
+
+Include {basic2}
+""".format(basic=_config("basic"),
+           basic2=_config("basic2"))
+        self.config = SSHConfig.from_text(config)
+
+    def test_include(self):
+        result = self.config.lookup("www.paramiko.org")
+        assert result["hostname"] == "www.paramiko.org"
+        assert result["user"] == "rando"
+
+        result = self.config.lookup("example.com")
+        assert result["hostname"] == "example.com"
+        assert result["user"] == "exampleuser"
+
+        result = self.config.lookup("example.org")
+        assert result["hostname"] == "example.org"
+        assert result["user"] == "exampleuser2"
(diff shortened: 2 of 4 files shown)


danihodovic:
:clap:

PeterJCLaw:
Have tested this and can confirm it works.
Test was rebased on https://github.com/paramiko/paramiko/commit/9151b5a5ef6634142cc810193a59630c863549c3.
Python 3.8 on Ubuntu, test also involved wildcard matching and proxyjump.

@bitprophet or maybe @ploxiln it would be great if you had time to review this PR?

endremborza:
It would be lovely if this could be merged, a nice, tested, and documented solution for a problem quite a few of us hack around.

bskinn:
Labeled and flagged for bitprophet review.

Unfortunately, I can't provide a timeline for the review on this, as bitprophet is neck-deep in dropping Python 2 and updating to modern Sphinx for his various projects. We *will* get to the point of considering other feature adds like this; I just can't say when.

paramiko/config.py
```diff
@@ -162,6 +163,27 @@ def parse(self, file_obj):
                 # at the end (for compatibility with issue #415). After 3.x, it
                 # will simply not get stripped, leaving a nice explicit marker.
                 context["config"][key] = None
+            elif key == 'include':
+                # Support for Include directive in ssh_config
+                path = value
+                # According to SSH documentation if the path is relative,
+                # we look forward in ~/.ssh
+                if (not os.path.isabs(path)) and path[0] != '~':
```
anarcat, line 171: path.startswith('~') would be more pythonic here...

paramiko/config.py
```diff
@@ -162,6 +163,27 @@ def parse(self, file_obj):
...
+
+                # Parse every included file
+                for filename in glob.iglob(path):
+                    if os.path.isfile(filename):
+                        if filename in parsed_files:
+                            raise ValueError("Include loop detected in ssh "
+                                             "config file: %s" % filename)
+                        with open(filename) as fd:
+                            parsed_files.append(filename)
+                            self.parse(fd, parsed_files)
```
anarcat, line 186: i'm trying to wrap my head around the scope of `parsed_files` here and it hurts a little. it *seems* like it's limited to the caller function, so in theory you *could* still create a loop where file A includes file B includes file C includes file A, and *not* be detected... does that make sense?

i wonder if a class variable should be used here instead, in other words. i understand it would break parsing the same file multiple times, but is that even supported?

paramiko/config.py
```diff
@@ -162,6 +163,27 @@ def parse(self, file_obj):
...
+                # Support for Include directive in ssh_config
+                path = value
+                # According to SSH documentation if the path is relative,
+                # we look forward in ~/.ssh
+                if (not os.path.isabs(path)) and path[0] != '~':
+                    path = os.path.join("~", ".ssh", path)
+                # Expand the user home path
+                path = os.path.expanduser(path)
+                if parsed_files is None:
+                    parsed_files = []
```
anarcat, line 176: wouldn't a `set()` be more appropriate here?

tests/test_config.py
```diff
@@ -990,3 +990,30 @@ def test_negated_canonical(self, socket):
...
+        assert result["hostname"] == "www.paramiko.org"
+        assert result["user"] == "rando"
+
+        result = self.config.lookup("example.com")
+        assert result["hostname"] == "example.com"
+        assert result["user"] == "exampleuser"
+
+        result = self.config.lookup("example.org")
+        assert result["hostname"] == "example.org"
+        assert result["user"] == "exampleuser2"
```
anarcat, line 1019: maybe we should test the include loops here to clarify the problem we're trying to solve with the `parsed_files` detection above...

anarcat (changes requested):

paramiko/config.py
mrmeszaros: I think this catches a circular inclusion, but a test case would be nice and would make it clear.

scratchmex:
Sorry for the ping but I don't understand how has been +4 years to get this merged. This PR is ready to merge. @bskinn @bitprophet

bskinn:
No worries, @scratchmex, thanks for checking in and expressing interest in the PR. Definitely helps to know where there's still active interest.

The delay is due to several factors.

- bitprophet only has a certain allocation per week for open source, and has multiple projects to maintain with that allocation (paramiko, invoke, fabric, and others).
- He had to run a major modernization campaign (dropping Py2 and older Py3, and upgrading several other things) across all those projects, which ran through much of 2022.
- Since then, a lot of his paramiko focus has been on the key/auth overhaul of #387 & #23 & related, still progressing in stages.
- Prior to both of these, before he recruited triagers (👋 hi!), he was also shouldering the full load of ticket handling across all the projects.
- Life stuff. Nuff said.

We'll get this evaluated and merged when we can. Still no guarantees on that timing, unfortunately.

I'll take a close look next time I'm at-keyboard, to see if there's anything else needing doing before flagging it for bitprophet.

anarcat:
> Sorry for the ping but I don't understand how has been +4 years to get this merged. This PR is ready to merge. @bskinn @bitprophet

I'm not sure if this is a rhetorical question or not, but I'll assume you're genuinely wondering why this is not merged.

First, GitHub says, right there above the comment field, that "This branch has conflicts that must be resolved", so that needs to be fixed, normally by the submitter.

Then two people reviewed this code and one (myself) found a few improvements that could be done to the code for it to be more readable. Not a blocker, but typically this is the kind of stuff you fix before merging and maintainers (at least in my case, I'm not a maintainer here, but in my projects) tend to disregard PRs with pending comments like this that don't have feedback from the submitter.

So this needs some ACK or NACK from @jsiverskog at the very least. Otherwise, anyone else can also take this patch, improve on it, and submit it as a separate PR, and I'm sure the fine folks here will eventually take a look.

mrmeszaros:
@anarcat @jsiverskog I am missing this feature, so I can try to finish this.
would You be okay with that?

jsiverskog:
@mrmeszaros : of course. i don't use paramiko myself anymore.

mrmeszaros:
Okay, so after reading into the specs and experimenting with dumping custom configs with `ssh -G <TEST_HOST>` it seems to me that a proper implementation would be a lot bigger than a weekend job.

I am still working on it, will publish a PR in the upcoming week hopefully :)

PR #1893, opened 2021-08-14
add sphinx setuptools integration

kloczek:
This setup.cfg file entry allows build documentation by
`python setup.py build_sphinx`.

Signed-off-by: Tomasz Kłoczko <kloczek@github.com>


kloczek:
Using setup.py is now deprecated so this PR should be not merger.
Closing.

PR #1907, opened 2021-09-16
config - implement Match final

commonism:
This PR adds the glue required to parse & process Match final blocks.

### DIFF
--- a/tests/configs/match-final
+++ b/tests/configs/match-final
@@ -1,0 +1,11 @@
+Host jump
+    HostName jump.example.orig
+    Port 1003
+
+Host finally
+    HostName finally.example.org
+    Port 1001
+
+Match final host "*.example.org" !host jump.example.org
+    ProxyJump jump
+    Port 1002
--- a/tests/configs/match-final
+++ b/tests/configs/match-final
@@ -6,6 +6,9 @@ Host finally
     HostName finally.example.org
     Port 1001
 
+Host default-port
+    HostName default-port.example.org
+
 Match final host "*.example.org" !host jump.example.org
     ProxyJump jump
     Port 1002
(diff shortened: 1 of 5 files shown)


commonism:
Match final was published with OpenSSH 8.0 at 2019-04-17 (https://www.openssh.com/txt/release-8.0)
It's commonly used for ProxyJump configurations, I would appreciate if this could get merged.
I use this for a year already, without any problems.

commonism:
rebased & linted

bskinn:
Thanks for the bump, @commonism.

Have you had a chance to look into where a note about this new feature should be added to the docs?

Also, this will need an entry in `/sites/www/changelog.rst`.

commonism:
I added a note to the Match documentation with keywords supported as of today, adding "final".
Does not suffice?

Will look into adding a changelog entry.

tests/configs/match-final
```diff
@@ -0,0 +1,11 @@
...
+    HostName jump.example.orig
+    Port 1003
+
+Host finally
+    HostName finally.example.org
+    Port 1001
+
+Match final host "*.example.org" !host jump.example.org
+    ProxyJump jump
+    Port 1002
```
bskinn, line 11: Coming back to this, looking more closely at the changes.

Unless I'm missing something, there's no test case that finds this entry. Is that intentional? Should this entry never be reachable?

(I don't know how `Match final` is supposed to work, so this may be a silly question.)

If there is a case that would be able to reach this entry (looks like any arbitrary subdomain other than `jump` or `finally`?), it would be worth adding.

bskinn: Thank you _very much_ for your patience on this, btw.

commonism: It's basically "does not use the default port from the match final" for the hosts in `test_final`

Add a host without port so we can test positive as well?

commonism:
added in https://github.com/paramiko/paramiko/pull/1907/commits/31e60cc3e89768259713ad67a8f282f15e84af65

bitprophet:
Thanks for this! Just tidied up a bit & merged.

PR #1909, opened 2021-09-29
Use poll instead of select to support large file descriptor numbers

jasonbridges:
Processes that have a large number of file descriptors or even a few with large numbers will fail using the library in the current state due to file descriptors larger than 1024 which select can't handle. ref: https://man7.org/linux/man-pages/man2/select.2.html (search for FD_SETSIZE)

### DIFF
--- a/paramiko/agent.py
+++ b/paramiko/agent.py
@@ -28,7 +28,7 @@ import threading
 import time
 import tempfile
 import stat
-from select import select
+import select
 from paramiko.common import asbytes, io_sleep
 from paramiko.py3compat import byte_chr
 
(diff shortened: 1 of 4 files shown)


jasonbridges:
I wasn't able to find an issue with select.poll using Python3 on OS X. What OS'es are not supported by select.poll?

jun66j5:
`select.poll()` is not available on Python 2 for macOS and Python 2/3 for Windows. `select.seelct()` should be used when the `.poll()` is not available.

#### Python 2 for macOS
```
$ python
WARNING: Python 2.7 is not recommended.
This version is included in macOS for compatibility with legacy software.
Future versions of macOS will not include Python 2.7.
Instead, it is recommended that you transition to using 'python3' from within Terminal.

Python 2.7.16 (default, Jun  5 2020, 22:59:21)
[GCC 4.2.1 Compatible Apple LLVM 11.0.3 (clang-1103.0.29.20) (-macos10.15-objc- on darwin
>>> import select
>>> select.poll
AttributeError: 'module' object has no attribute 'poll'
```

#### Python 3 for macOS
```
$ python3
Python 3.9.7 (default, Sep  3 2021, 12:45:31)
[Clang 12.0.0 (clang-1200.0.32.29)] on darwin
>>> import select
>>> select.poll
<built-in function poll>
```

#### Python 2 for Windows
```
Python 2.7.10 (default, May 23 2015, 09:44:00) [MSC v.1500 64 bit (AMD64)] on win32
>>> import select
>>> select.poll
AttributeError: 'module' object has no attribute 'poll'
>>> select.select
<built-in function select>
```
#### Python 3 for Windows
```
Python 3.9.6 (tags/v3.9.6:db3ff76, Jun 28 2021, 15:26:21) [MSC v.1929 64 bit (AMD64)] on win32
>>> import select
>>> select.poll
AttributeError: module 'select' has no attribute 'poll'
>>> select.select
<built-in function select>
```

jasonbridges:
OK, I can add windows support. Should we consider supporting python2 still?

jun66j5:
I consider that we should support Python 2.7 because README.rst and setup.py in master branch say supporting Python 2.7.

If Python 2.7 is not needed to support, we could use [selectors module](https://docs.python.org/3/library/selectors.html) rather than `select.{poll,epoll,kqueue,select}`.

bskinn:
It's my understanding that Python 2 support will be entering deprecated status relatively soon. @bitprophet, can you comment on the Python 2 deprecation timeframe and value of implementing Python 2 support here?

bitprophet:
That's correct. I'm planning to pop out one more bugfix and/or feature release with the existing support 'window' (2.7, 3.4 etc) after which I'll enact https://bitprophet.org/blog/2021/02/25/byethon2/#how-exactly . I've already got quite a list of what wants including in said release so this PR would be a good candidate for the subsequent, Python 3.6+-only, release.

jasonbridges:
Thanks, I'll update accordingly.

PR #1914, opened 2021-10-18
Support non-utf8 path in SFTPClient

Inndy:
This PR add ability to customize text encoding to use in SFTP client which can solve issue #546

The only problem is that repeating `encoding='utf8'` everywhere is not elegant at all. Do some one have better idea?

I think it could be solved by giving default value `None` to `encoding` and assign default value it in `py3compat.{u,b}`

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -549,13 +549,15 @@ class SSHClient(ClosingContextManager):
         chan.invoke_shell()
         return chan
 
-    def open_sftp(self):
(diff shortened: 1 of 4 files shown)


Inndy:
We may use `sys.getdefaultencoding()` as default encoding value for better compatibility

bskinn:
@bitprophet, I flagged this for policy decision in case there's a centralized discussion of encoding issues somewhere else that I'm not aware of yet.

PR #1920, opened 2021-10-30
Add support for wildcards in known_hosts #1001

filipeg:
Adds support for wildcards only.


jun66j5:
The patterns in known_hosts file supports only `*` and `?` as wildcards. However, `fnmatch` accepts additionally `[seq]` and `[!seq]`.

PR #1925, opened 2021-11-08
SHA2 with ssh agent

Sandarr95:
I've been trying to get an ssh connection working with this project to my server. At first I thought just the rsa-sha2 algorithms were missing. Trying out https://github.com/paramiko/paramiko/pull/1643 still did not work as I was using an ssh-agent.

This PR will dynamically assign the SHA algorithm as specified in preferred keys. It won't effectively change anything until the `_preferred_keys` include those rsa-sha2 algorithms (defined in https://github.com/paramiko/paramiko/pull/1643).
While opening this PR I also noticed https://github.com/paramiko/paramiko/pull/1644 which seems to be a partial implementation that just changes the flag, might be good to consider it for ongoing refinement.

This is the first time I'm doing python library programming and contributing to a python project. I know very little about pip, running tests and probably a bunch of other stuff, so don't hesitate to give me feedback :).

### DIFF
--- a/paramiko/agent.py
+++ b/paramiko/agent.py
@@ -56,7 +56,7 @@ class AgentSSH(object):
         self._conn = None
         self._keys = ()
 
-    def get_keys(self):
+    def get_keys(self, sec_opts):
         """
         Return the list of keys available through the SSH agent, if any.  If
         no SSH agent was running (or it couldn't be contacted), an empty list
@@ -66,7 +66,8 @@ class AgentSSH(object):
             a tuple of `.AgentKey` objects representing keys available on the
             SSH agent
         """
-        return self._keys
+        rsa_prefs = tuple(filter(lambda sha: sha in SHA_MAP, sec_opts.key_types))
+        return tuple(map(lambda key: AgentKey(self, key, rsa_prefs), self._keys))
 
     def _connect(self, conn):
         self._conn = conn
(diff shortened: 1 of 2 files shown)


paramiko/agent.py
```diff
@@ -412,7 +423,7 @@ def sign_ssh_data(self, data):
         msg.add_byte(cSSH2_AGENTC_SIGN_REQUEST)
         msg.add_string(self.blob)
         msg.add_string(data)
-        msg.add_int(0)
+        msg.add_int(SHA_MAP[self.name])
```
Sandarr95, line 426: https://tools.ietf.org/id/draft-miller-ssh-agent-01.html#sigflagnum
choose sha strategy based on name; sha1 = 0, sha256 = 2, sha512 = 4, bit 1 is reserved

paramiko/agent.py
```diff
@@ -392,11 +401,13 @@ class AgentKey(PKey):
...
     """
 
-    def __init__(self, agent, blob):
+    def __init__(self, agent, blob, rsa_prefs):
         self.agent = agent
         self.blob = blob
         self.public_blob = None
         self.name = Message(blob).get_text()
+        if(self.name in rsa_prefs):
+            self.name = rsa_prefs[0]
```
Sandarr95, line 410: Not sure how I should negotiate the key. This is (I think) only taking the clients' preference into consideration, and will fail if the client prefers SHA2 and the server does not support SHA2.

https://datatracker.ietf.org/doc/html/draft-ietf-curdle-rsa-sha2-12#section-3.3

bitprophet:

PR #1928, opened 2021-11-12
Make changes to allow paramiko to mostly work under FIPS

vakwetu:
Right now, because of a couple of invocations of md5() to generate
fingerprints for log messages, paramiko does not work on systems where
FIPS mode is enabled.  This is because hashlib.md5() fails when FIPS
mode is enabled on the kernel.

An attribute has been added to hashlib in python 3.9 to allow
the use of md5() in non-security contexts.  This is because FIPS allows
the use of md5() when secret data is not involved - which is exactly the
case here.

This attribute "usedforsecurity" has been backported to python versions
prior to python 3.9 on some distributions.  We therefore add a check
to confirm whether the attribute is supported, and if so, annotate
accordingly to allow the operation to succeed.

See https://bugs.python.org/issue9216 for more details.

### DIFF
--- a/paramiko/pkey.py
+++ b/paramiko/pkey.py
@@ -23,7 +23,6 @@ Common API for all public keys.
 import base64
 from binascii import unhexlify
 import os
-from hashlib import md5
 import re
 import struct
 
@@ -184,7 +183,7 @@ class PKey(object):
             a 16-byte `string <str>` (binary) of the MD5 fingerprint, in SSH
             format.
         """
-        return md5(self.asbytes()).digest()
+        return util.md5(self.asbytes(), usedforsecurity=False).digest()
 
     def get_base64(self):
         """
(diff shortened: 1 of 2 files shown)


afaranha:
The Docs check failure was reported as an issue: [#1940](https://github.com/paramiko/paramiko/issues/1940)

gouthampacha (commented):
+1 Can this be merged please, so we don't have to resort to monkey patching? :) https://review.opendev.org/c/openstack/manila/+/819375/7/manila/utils.py#74

bskinn:
Cross-referencing https://github.com/paramiko/paramiko/pull/2189#issuecomment-2103706707, making the suggestion that updating the hash algorithm in these spots might be preferable to suppressing the errors about the obsolete MD5.

PR #1932, opened 2021-11-24
chore: bump python version version in CI to 3.7

ltalirz:
Python 3.6 end of life is December 31st, 2021.
Bumping python versions for various CI tasks (linting, docs, etc.) to
Python 3.7.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
   sdist-test-suite:
     executor:
       name: orb/default
-      version: "3.6"
+      version: "3.7"
     steps:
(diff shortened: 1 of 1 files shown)


bitprophet:
See https://github.com/paramiko/paramiko/pull/1857#issuecomment-980495453 btw. I'll probably close this for now but please respond over there re: what the problem was in case it's bigger than I expect.

ltalirz:
Hey Jeff, sure I'll reply there. 

Note that this PR only moves the "housekeeping" stuff over to python 3.7 (linting, docs, ...).
python 3.6 is still tested as before.

Issue #1758, opened 2020-10-02
Get fingerprint in format of keyscan


Sir-Will:
I'm trying to get the ssh fingerprint to verify it but `binascii.hexlify(key.get_fingerprint())` returns a bunch of numbers and letters and I need the `sha256` hash format like the command `ssh-keygen -lf <(ssh-keyscan host 2>/dev/null)` provides.

How do I convert the fingerprint to `SHA256:pYYzsM9jP1Gwn1K9xXjKL2t0HLrasCxBQdvg/mNkuLg`?

ZhouLihua:
Hi Will,  I have the same request, do you find any other solutions for this

Sir-Will:
I ended using paramiko only for the host file lookup and for verifying I run the ssh commands.

```python
keyscan = subprocess.run(["ssh-keyscan", "-H", repoHostName], stderr=subprocess.DEVNULL, stdout=subprocess.PIPE, universal_newlines=True).stdout
keygen = subprocess.run(["ssh-keygen", "-lf", "-"], input=str.encode(keyscan), stdout=subprocess.PIPE).stdout
key = keyscan.splitlines()[1]
fingerprints = keygen.decode()

fingerprint = re.findall(r"(SHA.+?)\s.+\s\(RSA\)", fingerprints)[0]

if fingerprint == repo['server']['fingerprintRsa']:
    # fingerprint matches
```

PR #1937, opened 2021-11-28
pkey: fingerprint in sha256/base64

ploxiln:
New method `fingerprint_sha256_b64()` gives the fingerprint format just like openssh has for a while (and avoids `md5()` which is problematic for security policy compliance stuff).

----------

inspired by https://github.com/paramiko/paramiko/pull/1103
addresses https://github.com/paramiko/paramiko/issues/1758


ploxiln:
wrong remote, sorry!

PR #1950, opened 2021-12-15
Added timeout for Channel operations

Tomer-Ezon:
When trying to work with Channel object some of the operations are unnecessarily blocking,
added the ability to pass a timeout parameter to the methods in Channel so that if one chooses to limit the running time of some method from some reason he is able to.


APCBoston:
I am not sure if this was the case when you opened this pr, but today you can call the `Channel.settimeout()` instance method to set a timeout for all channel IO after creating the object.

PR #1951, opened 2021-12-16
Parse tokens in ProxyJump

sweenu:
This way, ProxyJump parsing can be implemented by higher-level clients with the benefits of tokens being already parsed.

### DIFF
--- a/paramiko/config.py
+++ b/paramiko/config.py
@@ -63,6 +63,7 @@ class SSHConfig(object):
         "hostname": ["%h"],
         "identityfile": ["~", "%d", "%h", "%l", "%u", "%r"],
         "proxycommand": ["~", "%h", "%p", "%r"],
+        "proxyjump": ["~", "%h", "%p", "%r"],
         # Doesn't seem worth making this 'special' for now, it will fit well
         # enough (no actual match-exec config key to be confused with).
         "match-exec": ["%d", "%h", "%L", "%l", "%n", "%p", "%r", "%u"],


ploxiln:
The `ssh_config` man page on my machine says:

```
     ProxyCommand
             Specifies the command to use to connect to the server.  The command string extends to the
             end of the line, and is executed using the user's shell ‘exec’ directive to avoid a lin‐
             gering shell process.

             Arguments to ProxyCommand accept the tokens described in the TOKENS section.
             ... more omitted ...
```

```
     ProxyJump
             Specifies one or more jump proxies as either [user@]host[:port] or an ssh URI.  Multiple
             proxies may be separated by comma characters and will be visited sequentially.  Setting
             this option will cause ssh(1) to connect to the target host by first making a ssh(1) con‐
             nection to the specified ProxyJump host and then establishing a TCP forwarding to the ul‐
             timate target from there.

             Note that this option will compete with the ProxyCommand option - whichever is specified
             first will prevent later instances of the other from taking effect.

             Note also that the configuration for the destination host (either supplied via the com‐
             mand-line or the configuration file) is not generally applied to jump hosts.
             ~/.ssh/config should be used if specific configuration is required for jump hosts.
```

The ProxyJump description does not mention TOKENS or substitution. Does openssh client perform TOKEN substitution for ProxyJump? For most use cases it's not really needed, as the jump/bastion host connection details do not depend on the final target port/user

sweenu:
It does. At my company, the ssh config distributed to every employee is one that includes `ProxyJump`s with `%r` tokens. In our case, the final server do care about the username, that is why. ProxyJump is just a ProxyComand behind if I'm correct, therefore, it makes sense that what ProxyCommand supports is also supported by ProxyJump.

sweenu:
Is there something blocking this PR to get merged?

bskinn:
Looks like no specific blockers to me; only thing I can think of, is there any way to easily add a test?

Either way, I'll bump it to @bitprophet's attention.

bitprophet:
Thanks for this! Added test/changelog, removed tilde (it makes no sense here) and merged. Will go out next feature release, whenever that is (may be 3.0 with python 2 removal, in fact).

PR #1960, opened 2021-12-27
Quirks support for dealing with strange and/or buggy clients

torkve:
Some clients have unexpected behavior that can be somehow buggy, violate SSH protocol, but the server has to deal with such clients.

The example of such behavior is OpenSSH<7.8 in combination with gnome-keyring as ssh-agent: when rsa-sha2-{256,512} is used, gnome-keyring still returns old SHA1 based signature, and older OpenSSH client versions just silently accept it, pretending it to be the signature being requested. Unfortunately this combinations happens on Ubuntu 16.04 LTS which is still widespread.

This PR adds support for client-name based quirks to deal with such behavior, and the quirk to deal with one exact bug described above.

A similar quirk in OpenSSH itself just disables signature type validation, however I believe it would be better to just not send SSH2_MSG_EXT_INFO which is commonly used to inform client that rsa-sha2-{256,512} algorithms are supported.


bitprophet:
Thinking it'd be ideal to pull this in during/after 3.0 (when we drop Python 2; it should be the next non-bugfix level release for this project, so not super duper far away) and then use an enum (which will be in stdlib at that point) instead of string comparisons. I'll think about it. Thanks for this!

PR #1964, opened 2022-01-03
Fix a bug of comparing keys with none

huashengdun:
This bug exists after version 2.7.2.

### DIFF
--- a/paramiko/pkey.py
+++ b/paramiko/pkey.py
@@ -140,6 +140,8 @@ class PKey(object):
         return cmp(self.asbytes(), other.asbytes())  # noqa
 
     def __eq__(self, other):
+        if other is None:
+            return False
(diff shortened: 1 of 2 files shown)


huashengdun:
If this bug is not fixed, it will not throw the BadHostKeyException.

client.py
-----------
```python
419                 our_key = our_server_keys.get(server_key.get_name())
420                 if our_key != server_key:       
421                     if our_key is None:             
422                         our_key = list(our_server_keys.values())[0]
423                     raise BadHostKeyException(hostname, server_key, our_key)
```

bitprophet:
Thanks for this! Unfortunately I think the more general approach in #2024 will catch more user bugs, so I'm going to merge that one instead 🙏🏻

PR #1966, opened 2022-01-05
Socket timeout kwarg

### LINKED ISSUE #1934: Socket connection hangs for a very long time
The call to [`socket.connect()`](https://github.com/paramiko/paramiko/blob/a598ca14d6451343c69be97e0e1c65eb15c0e607/paramiko/transport.py#L409) in the initialization of a `Transport` instance hangs for a few minutes for an inaccessible address. 
I created a fork in which `__init__()` accepts a `socket_timeout` argument that can be passed to `socket.settimeout()` in order to prevent freezes. If it is a welcome contribution, I would be happy to come up with a PR.

ZviBaratz:
Added a `socket_timeout` keyword argument to fix #1934.


paramiko/transport.py
```diff
@@ -310,6 +311,7 @@ def __init__(
         gss_kex=False,
         gss_deleg_creds=True,
         disabled_algorithms=None,
+        socket_timeout=10,
```
jun66j5, line 307: I consider the default of `socket_timeout` should be `None`. It will change behavior if [socket.setdefaulttimeout](https://docs.python.org/3/library/socket.html#socket.setdefaulttimeout) is used.

ZviBaratz: Please see the updated code, I hope I understood correctly.

ZviBaratz:
Should I also add kwargs to specify other timeout values? It seems like they're all assigned in the `__init__()` and it could be very useful to enable the user to change them if required.

PR #1970, opened 2022-01-16
Print exceptions along with the initial log message

paalka:
By using the built-in exc_info functionality, the exception will be part of the initial log entry, preventing it from for instance creating separate events in Sentry.

Builds on the work in https://github.com/paramiko/paramiko/pull/1737 and https://github.com/paramiko/paramiko/pull/406


antonhedstrom:
Did this get stuck somehow?
We have the same experience as describes in #2059 and would very much appreciate this change.

PR #1971, opened 2022-01-16
Change default bufsize for SFTP files to match the max request size

ahjmorton:
I'd be very interested to hear if my reasoning for the change makes sense.

In the default case (where `bufsize` is -1) the default value of 8KB is used as the buffer for the SFTP file. This means if we accept a bunch of writes that are greater than 8KB but less than 32KB we'll submit a packet below the maximum size we're allowed, potentially affecting throughput. An example where I've observed this is streaming a large file through an SFTP file where the source data is also writing in 8KB chunks.

In terms of tradeoffs I can think of it does mean by default, on reads, we'll request at least 32KB at a time vs the 8KB we're currently doing. 
Also it means memory usage per SFTP_file instance will increase by 4x which may be sad if someone has a lot of files.

Interested to hear thoughts if the potential trade offs are acceptable.

### DIFF
--- a/paramiko/sftp_file.py
+++ b/paramiko/sftp_file.py
@@ -57,11 +57,14 @@ class SFTPFile(BufferedFile):
     # Some sftp servers will choke if you send read/write requests larger than
     # this size.
     MAX_REQUEST_SIZE = 32768
+    DEFAULT_BUFSIZE = MAX_REQUEST_SIZE
 
     def __init__(self, sftp, handle, mode="r", bufsize=-1):
         BufferedFile.__init__(self)
         self.sftp = sftp
         self.handle = handle
+        if bufsize < 0:
+            bufsize = self.DEFAULT_BUFSIZE
         BufferedFile._set_mode(self, mode, bufsize)
         self.pipelined = False
         self._prefetching = False


jun66j5:
I tried the patch with Python 3.9 and 2.7, however throughput doesn't appear improvement (using OpenSSH 8.2p1 on localhost, data file 128 MB).

#### Python 3.9

|     | with patch | without patch |
| --- | ---------: | ------------: |
| GET | 11.93 MB/s |    11.09 MB/s |
| GET | 11.02 MB/s |    81.16 MB/s |
| GET | 14.65 MB/s |    11.77 MB/s |
| PUT | 38.05 MB/s |    43.61 MB/s |
| PUT | 39.68 MB/s |    44.02 MB/s |
| PUT | 31.89 MB/s |    34.03 MB/s |

#### Python 2.7

|     | with patch | without patch |
| --- | ---------: | ------------: |
| GET | 12.23 MB/s |    10.15 MB/s |
| GET | 74.05 MB/s |    12.26 MB/s |
| GET | 10.71 MB/s |    69.46 MB/s |
| PUT |  6.52 MB/s |     6.08 MB/s |
| PUT |  6.39 MB/s |     6.59 MB/s |
| PUT |  6.60 MB/s |     6.57 MB/s |

Also, I think we could simply set 32KB to `_DEFAULT_BUFSIZE` to change the buffer size, rather than changing `bufsize` passed to `_set_mode()`.

```diff
diff --git a/paramiko/sftp_file.py b/paramiko/sftp_file.py
index 0104d8577..00c4b6af5 100644
--- a/paramiko/sftp_file.py
+++ b/paramiko/sftp_file.py
@@ -57,6 +57,7 @@ class SFTPFile(BufferedFile):
     # Some sftp servers will choke if you send read/write requests larger than
     # this size.
     MAX_REQUEST_SIZE = 32768
+    _DEFAULT_BUFSIZE = MAX_REQUEST_SIZE

     def __init__(self, sftp, handle, mode="r", bufsize=-1):
         BufferedFile.__init__(self)
```

Issue #1963, opened 2022-01-03
Regression in 2.9.x - authentication with signed ssh key fails


forsberg:
I'm using signed ssh keys to authenticate to a host. This has worked flawlessly up until Paramiko 2.8.1, but with 2.9.0 and 2.9.1 it fails.

Example program:

```python
import os
import sys
import paramiko
import logging

script_dir = os.path.dirname(os.path.abspath(__file__))

def try_connect(hostname):
    logging.basicConfig(level=logging.DEBUG, stream=sys.stderr)

    client = paramiko.SSHClient()
    client.set_missing_host_key_policy(paramiko.AutoAddPolicy())

    proxy_command = os.path.join(script_dir, "../../../ssh/proxy-command")

    sock = paramiko.ProxyCommand(f"{proxy_command} {hostname}")

    client.connect(hostname, 22, username="root", sock=sock, key_filename="/home/forsberg/.ssh/robot-test-cert.pub")

if __name__ == "__main__":
    try_connect(sys.argv[1])
```

1. The proxy command is just a netcat forwarding the TCP packets, and is probably not important anyway in this context.
2. `robot-test-cert.pub` is signed via ssh-keygen, and has the following metadata:

```
ssh-keygen -Lf ~/.ssh/robot-test-cert.pub 
/home/forsberg/.ssh/robot-test-cert.pub:
        Type: ssh-rsa-cert-v01@openssh.com user certificate
        Public key: RSA-CERT SHA256:JtAqrYvjJBLxkXDeSkXEuGyTR9uykjzaqlGIExjFi/g
        Signing CA: RSA SHA256:E5U81luuNJQIojmQVtDNfnKa4s89Q1gMTOIDqhk1jJE (using ssh-rsa)
        Key ID: "erik.forsberg"
        Serial: 1
        Valid: from 2022-01-02T09:36:22 to 2022-01-10T09:36:22
        Principals: 
                root
                ssh_gate
        Critical Options: (none)
        Extensions: 
                permit-X11-forwarding
                permit-agent-forwarding
                permit-port-forwarding
                permit-pty
                permit-user-rc
```

Running this program with paramiko 2.8.1 gives the following output:

```
python ssh_via_paramiko hostname.example.com
DEBUG:paramiko.transport:starting thread (client mode): 0xb8e398b0
DEBUG:paramiko.transport:Local version/idstring: SSH-2.0-paramiko_2.8.1
[...]

PR #1977, opened 2022-01-20
Fix publickey authentication with signed RSA key (#1963)

jun66j5:
When paramiko client uses signed RSA key for publickey authentication, the session blob is generated with `rsa-sha2-{256,512}` and `ssh-rsa` rather than `*-cert-v01@openssh.com` in `auth_handler.py`. As the result, the publickey authentication is failed.

`rsa-sha2-{256,512}-cert-v01@openssh.com` and `ssh-rsa-cert-v01@openssh.com` should be used for the session blob when signed RSA key is used.

### DIFF
--- a/paramiko/auth_handler.py
+++ b/paramiko/auth_handler.py
@@ -341,6 +341,8 @@ class AuthHandler(object):
                 DEBUG,
                 "NOTE: you may use the 'disabled_algorithms' SSHClient/Transport init kwarg to disable that or other algorithms if your server does not support them!",  # noqa
             )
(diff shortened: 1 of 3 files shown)


11chri:
I just checked your PR on Paramiko 2.9.2 and it fixes #1963

11chri (approved):
I just checked your PR on Paramiko 2.9.2 and it fixes https://github.com/paramiko/paramiko/issues/1963

bitprophet:
✨

PR #1985, opened 2022-02-10
Add six to `install_requires`

sondrelg:
Hi 👋 

[six](https://pypi.org/project/six/) is imported and used, in Paramiko, [here](https://github.com/paramiko/paramiko/blob/main/paramiko/pkey.py#L30). 

Unfortunately, since it's not listed as a dependency, importing paramiko seems to fail: 

```
Feb 10 10:53:57 info   File "/app/....py", line 6, in <module>
Feb 10 10:53:57 info     import paramiko
Feb 10 10:53:57 info   File "/usr/local/lib/python3.10/site-packages/paramiko/__init__.py", line 22, in <module>
Feb 10 10:53:57 info     from paramiko.transport import SecurityOptions, Transport
Feb 10 10:53:57 info   File "/usr/local/lib/python3.10/site-packages/paramiko/transport.py", line 91, in <module>
Feb 10 10:53:57 info     from paramiko.dsskey import DSSKey
Feb 10 10:53:57 info   File "/usr/local/lib/python3.10/site-packages/paramiko/dsskey.py", line 37, in <module>
Feb 10 10:53:57 info     from paramiko.pkey import PKey
Feb 10 10:53:57 info   File "/usr/local/lib/python3.10/site-packages/paramiko/pkey.py", line 30, in <module>
Feb 10 10:53:57 info     import six
Feb 10 10:53:57 info ModuleNotFoundError: No module named 'six'
```

This PR adds it to setup.py's `install_requires`. Currently haven't specified a version range.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -90,6 +90,6 @@ setup(
     # TODO 3.0: alternately, given how prevalent ed25519 is now, and how we use
     # invoke for (increasing) subproc stuff, consider making the default flavor
     # "full" and adding a "minimal" or similar that is just-crypto?
-    install_requires=["bcrypt>=3.1.3", "cryptography>=2.5", "pynacl>=1.0.1"],
+    install_requires=["bcrypt>=3.1.3", "cryptography>=2.5", "pynacl>=1.0.1", "six"],
     extras_require=extras_require,
 )


bskinn:
No, I haven't seen anything anywhere about issues with a `six` `ImportError`, though I don't touch the Python 2 sphere of the ecosystem very much. 

It's very odd, because any broader cause I can think of, I would think would be a _much_ noisier matter for the `paramiko` bugtracker than it has been.

sondrelg:
> What's odd is this is the first time I've seen this come up. Another contributor guesses maybe we got six via a transitive dependency which has since dropped it

This seems like the most likely explanation to me. We use poetry and auto-update dependencies within our specified major ranges daily, assuming tests pass, so unfortunately I'm not sure I'll be able to trace the change very easily.

I agree dropping six/v2 support altogether sounds reasonable here assuming that's days away, but it sound like it might be worth a patch bump otherwise? I doubt I'll be the last to see this issue 🙂 Just out of curiosity, is the release process very involved?

sondrelg:
I think it's a 3.10 issue. 

- Running `python -m venv .venv`, 
- activating it
- running `pip install paramiko`
- and importing paramiko

doesn't replicate the issue, since bcrypt is a dependency of paramiko and six is a dependency of bcrypt.

sondrelg:
We have `bcrypt = ">=3.1.3"` in our lockfile, which explains it.

bskinn:
Thanks for diagnosing, @sondrelg! Not sure how this affects your thought process, @bitprophet.

bitprophet:
The release process itself isn't that involved (though I admit I'm still way too conservative about it 😂 ) but I have an allergy to dependency updates - even innocuous-looking ones end up causing weird corner case issues in far-off areas of the ecosystem.

That said...this one should really be a noop for anyone with older bcrypt, and would fix this issue on newer bcrypt, so what the heck.

bitprophet:
Cherry-picked this back to 2.8 and merged up. Thanks!

PR #1987, opened 2022-02-16
Handle %C token in ControlPath

### LINKED ISSUE #1976: Support for %C token
An application using Paramiko was failing for me with the following error message:
`ssh_config percent_expand: unknown key %C`

The cause is this SSH config statement:
```
Host *
    ControlMaster auto
    ControlPath /tmp/ssh-cm-%C
    ControlPersist 30m
```

`%C` is commonly used for the ControlPath nowadays. It would be great if Paramiko could support that.

jbrand42:
Resolves #1976

### DIFF
--- a/paramiko/config.py
+++ b/paramiko/config.py
@@ -27,7 +27,7 @@ import os
 import re
 import shlex
 import socket
-import hashlib
+from hashlib import sha1
 from functools import partial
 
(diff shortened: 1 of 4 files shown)


bitprophet:
Merged and enhanced. Thanks!

PR #1995, opened 2022-03-03
Add optional parameters for encoding different from utf8 (default is still utf8)

hakaishi:
Continuation of original pull request #1716

Cloned the current main branch and added my function parameters. No much change. Fixed 2 failing tests too.

Optional parameters are "encoding" with default to utf8 and "errors" with default to "strict".

You could add a paramiko global or something to avoid the need to pass the encoding/errors parameters each time.

Best regards,
Hakaishi

### DIFF
--- a/paramiko/file.py
+++ b/paramiko/file.py
@@ -95,7 +95,7 @@ class BufferedFile(ClosingContextManager):
 
     if PY2:
 
-        def next(self):
+        def next(self, encoding="utf8", errors="strict"):
             """
             Returns the next line from the input, or raises
             ``StopIteration`` when EOF is hit.  Unlike Python file
@@ -105,14 +105,14 @@ class BufferedFile(ClosingContextManager):
 
             :returns: a line (`str`) read from the file.
             """
-            line = self.readline()
+            line = self.readline(encoding=encoding, errors=errors)
             if not line:
                 raise StopIteration
             return line
 
     else:
 
-        def __next__(self):
+        def __next__(self, encoding="utf8", errors="strict"):
             """
             Returns the next line from the input, or raises ``StopIteration``
             when EOF is hit.  Unlike python file objects, it's okay to mix
@@ -122,7 +122,7 @@ class BufferedFile(ClosingContextManager):
 
             :returns: a line (`str`) read from the file.
             """
-            line = self.readline()
+            line = self.readline(encoding=encoding, errors=errors)
             if not line:
                 raise StopIteration
             return line
@@ -228,7 +228,7 @@ class BufferedFile(ClosingContextManager):
         self._pos += len(result)
         return result
 
-    def readline(self, size=None):
+    def readline(self, size=None, encoding="utf8", errors="strict"):
         """
         Read one entire line from the file.  A trailing newline character is
         kept in the string (but may be absent when a file ends with an
@@ -294,7 +294,9 @@ class BufferedFile(ClosingContextManager):
             if (new_data is None) or (len(new_data) == 0):
                 self._rbuffer = bytes()
                 self._pos += len(line)
-                return line if self._flags & self.FLAG_BINARY else u(line)
+                return line if self._flags & self.FLAG_BINARY else u(line,
+                                                                    encoding,
+                                                                    errors)
             line += new_data
             self._realpos += len(new_data)
         # find the newline
@@ -306,7 +308,9 @@ class BufferedFile(ClosingContextManager):
         if pos == -1:
             # we couldn't find a newline in the truncated string, return it
             self._pos += len(line)
-            return line if self._flags & self.FLAG_BINARY else u(line)
+            return line if self._flags & self.FLAG_BINARY else u(line,
+                                                                 encoding,
+                                                                 errors)
         xpos = pos + 1
         if (
             line[pos] == cr_byte_value
@@ -331,9 +335,10 @@ class BufferedFile(ClosingContextManager):
         else:
             self._record_newline(lf)
         self._pos += len(line)
-        return line if self._flags & self.FLAG_BINARY else u(line)
+        return line if self._flags & self.FLAG_BINARY else u(line,
+                                                             encoding, errors)
 
-    def readlines(self, sizehint=None):
+    def readlines(self, sizehint=None, encoding="utf8", errors="strict"):
         """
         Read all remaining lines using `readline` and return them as a list.
         If the optional ``sizehint`` argument is present, instead of reading up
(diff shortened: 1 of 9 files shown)


paramiko/sftp_client.py
```diff
@@ -704,20 +846,26 @@ def putfo(self, fl, remotepath, file_size=0, callback=None, confirm=True):
         :param bool confirm:
             whether to do a stat() on the file afterwards to confirm the file
             size (since 1.7.7)
+        :param str encoding:
+            encoding for file paths/names and content (default: utf8)
```
jun66j5, line 850: The `encoding` and `errors` parameters are not used for the content because `putfo` uses binary mode.

paramiko/file.py
```diff
@@ -95,34 +95,42 @@ def flush(self):
...
-            line = self.readline()
+            line = self.readline(encoding=encoding, errors=errors)
             if not line:
                 raise StopIteration
             return line
 
     else:
 
-        def __next__(self):
+        def __next__(self, encoding="utf8", errors="strict"):
```
jun66j5, line 119: The `__next__` method doesn't have arguments. It is unable to use the arguments because Python iterator directly invokes it.

Instead, I suggest:
 * Add `encoding` and `errors` parameters to `__init__` method and store to the instance attributes
 * Use the attributes here

hakaishi: I see. I should have noticed that....

I'll try and see what to do best. Thanks.

paramiko/sftp_client.py
hakaishi: I guess you refer to the following part:
```
with self.file(
            remotepath, "wb", encoding=encoding, errors=errors
        ) as fr:
            fr
```
Right, it's in binary mode, but that's just the content, the file path might be non uft8 though. And just below that part the stat method is also used, which takes these parameters to get the actual remote path in the remote encoding.

Actually, this remotepath over there might need encoding too. I'm not sure... I'll try and test this week. (All you need is a VM or Docker with an OS using an encoding different from utf8...)
-> `remotepath.decode("utf-8", errors).encode(encoding, errors)`

hakaishi: Update:    
self.file is an object of open and open takes encoding and errors as parameter. Encoding and errors in open decodes the file name/path.

In other words: Everything should be correct here. You didn't dig deep enough, I guess.

paramiko/file.py
hakaishi: I fixed this and merged the newest commits from the main branch.

jun66j5: I think `readline` should use `self.encoding` and `self.errors` and remove added arguments, because passing two parameters to `__init__`.

hakaishi: Oops. I'll check again.

hakaishi: Better now?

paramiko/sftp_client.py
jun66j5: What I mean is that the comment for the parameter should be:
```
encoding for file paths/names (default: utf8)
```
I think the current comment leads misleading. In fact, the encoding parameter is not used for the content.

hakaishi: I fixed that now.

hakaishi:
By the way... How about having a encoding parameters at SFTPClient's init instead of every method (list etc)

hakaishi:
The new commit removes all encoding/errors parameters from SFTPClients methods and puts them instead into the constructor and the from_transport method.

hakaishi:
I will have to merge this manually over and over again if it won't get merged/pulled...   
Depressing...
Anything speaking against a merge?

hakaishi:
merged current main into this pull request.  
Any updates why this is not getting (some) priority? - Many people in many countries will be waiting for this (including me) as many servers don't run on UTF8.

tests/test_sftp.py
```diff
@@ -277,7 +278,7 @@ def test_listdir_in_locale(self, sftp):
         sftp.open(sftp.FOLDER + "/canard.txt", "w").close()
         try:
             folder_contents = sftp.listdir(sftp.FOLDER)
-            self.assertEqual(["canard.txt"], folder_contents)
+            TestCase().assertEqual(["canard.txt"], folder_contents)
```
bskinn, line 281: What is your reasoning behind this replacement? This seems like it will unnecessarily create a new, empty `TestCase` instance for each assert, instead of just using the one that already exists in the form of `self`.

tests/__init__.py
```diff
@@ -46,7 +46,7 @@ def _test_under_locale(testself, sftp):
                 try_locales.pop(0)
             else:
                 try:
-                    return test_method(testself)
+                    return test_method(testself, sftp)
```
bskinn, line 49: I'm relatively unfamiliar with the guts of the test suite, but what does adding the `sftp` here accomplish? I don't see anything obvious in any of the other changes where this is used.

tests/test_channelfile.py
```diff
@@ -6,23 +6,29 @@
 class ChannelFileBase(object):
     @patch("paramiko.channel.ChannelFile._set_mode")
     def test_defaults_to_unbuffered_reading(self, setmode):
-        self.klass(Channel(None))
+        self.klass(Channel(None), encoding="utf8", errors="strict")
```
bskinn, line 9: You've added `encoding="utf8"` cases to many tests, but I don't see any tests that probe compatibility with other encodings. Those would be good to add.

bskinn (commented):
I have a couple of questions, and a suggestion for additional tests.

Thanks for putting together a PR for this!

tests/test_sftp.py
hakaishi: Hmm? I don't remember changing this myself. I'll check when I come home tomorrow night.

Thank you.

tests/test_channelfile.py
hakaishi: Right. In the first place this is supposed to "not break the current tests".

Adding one or two different test cases where the system does not use utf8 would be very welcome. I'll try to put something together, but I'm not yet sure at how to do it.

We have to keep in mind that there are two different problems: Encoding of file content and encoding of file paths including system messages. (Welcome message on login and error messages etc)

bskinn: Ah, I see -- with this PR, you're going for a ~zero-effect change, making sure that adding the `encoding` and `errors` arguments didn't break anything in the default `utf8` use case? Thus, setting the stage for future development/testing when new encodings enter the mix?

hakaishi: Yes. That's right 🙂

PR #2003, opened 2022-03-13
util: store thread assigned id in thread-local storage

### LINKED ISSUE #2002: Memory leak with long-running processes
Paramiko stores thread-ids in a dictionary, which never gets cleaned up. In a long-running process, this leaks memory. One example is https://github.com/truenas/zettarepl which runs persistently in a TrueNAS system, and thus leaks memory.

The leaking structure is here: https://github.com/paramiko/paramiko/blob/main/paramiko/util.py#L228

rkojedzinszky:
Fixes #2002

### DIFF
--- a/paramiko/util.py
+++ b/paramiko/util.py
@@ -225,24 +225,20 @@ def mod_inverse(x, m):
     return u2
 
 
-_g_thread_ids = {}
+_g_thread_data = threading.local()
 _g_thread_counter = 0
 _g_thread_lock = threading.Lock()
 
 
 def get_thread_id():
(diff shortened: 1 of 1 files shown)


bitprophet:
Thanks! Looks like this only actually impacts that log_to_file function, which I suppose some users are referencing, but of course the practical effect is that it bloats memory for ALL users. Works for me under Python 2 and 3 (wasn't sure when threadlocals grew an 'id' attribute - not seeing it in a trivial test, but it's clearly working in the real...).

bitprophet:
Cherry-picked into 2.10 tree.

PR #2006, opened 2022-03-15
Fix Free Software Foundation address

pghmcfc:
They moved from Temple Place to Franklin Street in 2005.

### DIFF
--- a/demos/demo.py
+++ b/demos/demo.py
@@ -16,7 +16,7 @@
 #
 # You should have received a copy of the GNU Lesser General Public License
 # along with Paramiko; if not, write to the Free Software Foundation, Inc.,
-# 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA.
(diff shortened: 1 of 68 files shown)


bitprophet:
Big "thanks, I hate it" energy 😂  Prompted me to doublecheck and yea, LGPL does require us to have these headers (which I'm sure is inconsistently applied, but oh well). Otherwise I would just remove them instead. They better not move offices again...

bskinn:
Seems like a DOI-style approach to their physical location would be ideal here. Static object of record to be referenced in the license language, whose content can be updated with new addresses without pushing a maintenance task on all projects using the license.

Symlink to a symlink, basically. :smile:

bitprophet:
I really wonder if the intersect of the below sets of people is >0 in size:

- Has a license problem
- With Paramiko specifically
- ...and in particular, a weirdly unattached source file, no LICENSE file in sight
- Thinks the FSF needs contacting about it, and not us
- ...and doesn't want to email them, or call them, but instead wants to snail-mail or knock on a door

😂

Issue #2008, opened 2022-03-16
Windows OpenSSH error on version 2.10


danilofuchs:
Upgrading from 2.9.2 to 2.10.2 caused issues on Windows.

Command:
```python
ssh_client = paramiko.SSHClient()
ssh_client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
ssh_client.connect(self.ip, username=self.username, password=self.password, timeout=self.timeout_s)

data = 'cat some_file.txt'
ssh_client.exec_command(data, get_pty=True)
```

## 2.9.2:

Runs normally

## 2.10.2:

```
[Errno 22] Invalid argument: '\\\\.\\pipe\\openssh-ssh-agent'
```

Daviid-P:
I'm having the same problem.

```py
from pathlib import Path
PIPE_NAME = Path(r"\\.\pipe\openssh-ssh-agent") 
# WindowsPath(PIPE_NAME)
PIPE_NAME.exists() 
# True
_pipe = open(PIPE_NAME, "rb+", buffering=0)     
_pipe
# <_io.FileIO name='\\\\.\\pipe\\openssh-ssh-agent' mode='rb+' closefd=True>
```

Manually it doesn't seem to fail.

Using Python 3.10.2

jun66j5:
I found https://stackoverflow.com/a/67102903.

> os.path.exists delegates to os.stat, and on Windows, os.stat tries to [open](https://github.com/python/cpython/blob/v3.9.4/Modules/posixmodule.c#L1826) the file:
> ...
> Then it closes the file. At this point, the client can't reopen the pipe unless the server side first calls [DisconnectNamedPipe](https://docs.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-disconnectnamedpipe), which doesn't happen in this code. Also, when the server side tries to read from the pipe, it's reading from os.stat's connection attempt, which closed the pipe without writing any data.

That is reproduce like this:
```
>>> PIPE_NAME = r"\\.\pipe\openssh-ssh-agent"
>>> os.path.exists(PIPE_NAME) and open(PIPE_NAME, "rb+", buffering=0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OSError: [Errno 22] Invalid argument: '\\\\.\\pipe\\openssh-ssh-agent'
```
Or:
```
>>> [open(PIPE_NAME, "rb+", buffering=0) for i in range(1)]
[...]

PR #2010, opened 2022-03-18
Fix `Errno 22` when connecting pipe of openssh ssh agent

jun66j5:
Proposed changes for #2008.

* use `os.listdir()` to check existence of `\\.\pipe\openssh-ssh-agent`, because `os.path.exists()` uses `CreateFile` API and the pipe cannot be reopen unless the server calls `DisconnectNamedPipe()`
* retry when `open()` raises Errno 22 which maens that the server has not called `DisconnectNamedPipe()` yet.
* use `os.{open,read,write,close}()` because `self._pipe.read()` raises `IOError: [Errno 0] Error` on Python 2

### DIFF
--- a/paramiko/win_openssh.py
+++ b/paramiko/win_openssh.py
@@ -18,23 +18,39 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA.
 
 import os.path
+import time
 
 PIPE_NAME = r"\\.\pipe\openssh-ssh-agent"
 
 
 def can_talk_to_agent():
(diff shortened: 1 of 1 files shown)


aspenforest:
Why isn't this bugfix merged and released? It's been a month and paramiko on Windows is essentially useless without this bugfix.

YevgenyFarber:
When the patch will be released?

bitprophet:
Thanks for this, @jun66j5! I can't test it myself but I'll trust that it fixed things on your end and that of the other users who reported success. Cherry-picking this back to 2.10 and newer.

Issue #2004, opened 2022-03-13
Test suite will need to be able to handle SHA-1 signatures not being supported


pghmcfc:
Red Hat Enterprise Linux 9 will have SHA-1 signatures disabled by default:
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/VVLHQAWI3IQ7NRLKMUHJ27JV3V2JAFDP/
This change is already visible in CentOS 9 Stream, which is where development of RHEL-9 is happening.
Building paramiko on CentoOS 9 Stream results in many test suite failures due to lack of SHA-1 signature support, like this one:
```
_______________________ SSHClientTest.test_auth_timeout ________________________

self = <tests.test_client.SSHClientTest testMethod=test_auth_timeout>

    @slow
    def test_auth_timeout(self):
        """
        verify that the SSHClient has a configurable auth timeout
        """
        # Connect with a half second auth timeout
>       self.assertRaises(
            AuthenticationException,
            self._test_connection,
            password="unresponsive-server",
            auth_timeout=0.5,
        )

tests/test_client.py:501: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/test_client.py:205: in _test_connection
    self.tc.connect(**dict(self.connect_kwargs, **kwargs))
paramiko/client.py:406: in connect
    t.start_client(timeout=timeout)
paramiko/transport.py:698: in start_client
    raise e
paramiko/transport.py:2109: in run
    ptype, m = self.packetizer.read_message()
paramiko/packet.py:459: in read_message
    header = self.read_all(self.__block_size_in, check_rekey=True)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def read_all(self, n, check_rekey=False):
        """
        Read as close to N bytes as possible, blocking as long as necessary.
    
        :param int n: number of bytes to read
        :return: the data read, as a `str`
    
        :raises:
            ``EOFError`` -- if the socket was closed before all the bytes could
            be read
        """
        out = bytes()
        # handle over-reading from reading the banner line
        if len(self.__remainder) > 0:
            out = self.__remainder[:n]
            self.__remainder = self.__remainder[n:]
            n -= len(out)
[...]

PR #2011, opened 2022-03-20
Skip tests requiring sha1 signing if the backend doesn't support that

pghmcfc:
Red Hat Enterprise Linux 9 will have SHA-1 signatures disabled by default. It is likely that SHA-1 signatures will disappear elsewhere over time too. This change detects if sha1 signatures are not supported by the backend and skips tests that rely on that functionality. This is a workaround for #2004.

It would be good to reduce the reliance of the test suite on sha1 signatures except in the cases where that is explicitly being tested, and the markers added here give a decent starting point for seeing where to change things.


pghmcfc:
`tests/test_client.py::SSHClientTest::test_cleanup` looks to be a flaky test, at least on CircleCI. Everything else looks OK to me now.

bitprophet:
Yea there is some intermittent flakiness on Circle that I haven't dug into yet. Looks good on my end. Thanks!

Issue #2023, opened 2022-04-05
Comparing PKey to None fails


jsonar-cpapke:
We recently updated to paramiko version 2.10.3 and encountered a bug with PKey. If you create a key, and then check the equality vs `None`, it will throw an exception:
```
./site-packages/paramiko/pkey.py:143: in __eq__
    return self._fields == other._fields
E   AttributeError: 'NoneType' object has no attribute '_fields'
```

This issue can be duplicated by running this code:
```
RSAKey(filename=public_key_path) == None
```

This seems like a bug, as equals should instead return false when something is of the wrong type, instead of throwing an exception.

bitprophet:
Rolling into the PR.

PR #2024, opened 2022-04-05
don't throw exception when comparing PKey to non-PKey

jsonar-cpapke:
Fixes https://github.com/paramiko/paramiko/issues/2023

### DIFF
--- a/paramiko/pkey.py
+++ b/paramiko/pkey.py
@@ -140,7 +140,7 @@ class PKey(object):
         return cmp(self.asbytes(), other.asbytes())  # noqa
 
     def __eq__(self, other):
-        return self._fields == other._fields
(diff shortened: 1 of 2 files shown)


jsonar-cpapke:
I realize there is another pull request open for this same issue (see https://github.com/paramiko/paramiko/pull/1964), but I believe this fixes it better in general, instead of just for comparison against `None`.

jsonar-cpapke:
CI issues look to be unrelated to this PR, they should be fixed in https://github.com/paramiko/paramiko/pull/2025

bitprophet:
Cherry picked this back thru 2.9 for funsies. Thanks! Agree this fix is more generally applicable than #1964.

PR #1870, opened 2021-06-19
threading related deprecation fixes for python 3.10

Narendra-Neerukonda:
Fixes #1838

PR #2028, opened 2022-04-11
threading related deprecation fixes for python 3.10

### LINKED ISSUE #1838: Threading related deprecations in Python 3.10
notifyAll has been deprecated in favor of notifyAll
currentThread has been deprecated in favor of current_thread
setDaemon has been deprecated in favor of setting daemon attribute

Reference : https://bugs.python.org/issue43723

```
➜  paramiko git:(master) rg -t py -w 'currentThread|notifyAll|activeCount|getName|setName|isDaemon|setDaemon'
demos/rforward.py
83:        thr.setDaemon(True)

paramiko/buffered_pipe.py
104:            self._cv.notifyAll()
206:            self._cv.notifyAll()

paramiko/util.py
235:    tid = id(threading.currentThread())

paramiko/sftp_file.py
530:        t.setDaemon(True)

paramiko/transport.py
420:        self.setDaemon(True)

paramiko/channel.py
1069:            self.out_buffer_cv.notifyAll()
1233:        self.out_buffer_cv.notifyAll()

tests/loop.py
84:            self.__cv.notifyAll()

tests/test_transport.py
747:                self.setDaemon(True)
771:                self.setDaemon(True)
```

graingert:
fixes #1838 

remake of https://github.com/paramiko/paramiko/pull/1870 with py2.6 support

### DIFF
--- a/demos/rforward.py
+++ b/demos/rforward.py
@@ -35,6 +35,7 @@ import threading
 from optparse import OptionParser
 
 import paramiko
+from paramiko.py3compat import set_daemon
 
 SSH_PORT = 22
 DEFAULT_PORT = 4000
@@ -80,7 +81,7 @@ def reverse_forward_tunnel(server_port, remote_host, remote_port, transport):
         thr = threading.Thread(
             target=handler, args=(chan, remote_host, remote_port)
         )
-        thr.setDaemon(True)
+        set_daemon(thr, True)
         thr.start()
 
 
--- a/paramiko/buffered_pipe.py
+++ b/paramiko/buffered_pipe.py
@@ -25,7 +25,7 @@ read operations are blocking and can have a timeout set.
 import array
 import threading
 import time
-from paramiko.py3compat import PY2, b
+from paramiko.py3compat import PY2, b, notify_all
 
 
 class PipeTimeout(IOError):
@@ -101,7 +101,7 @@ class BufferedPipe(object):
             if self._event is not None:
                 self._event.set()
             self._buffer_frombytes(b(data))
-            self._cv.notifyAll()
+            notify_all(self._cv)
         finally:
             self._lock.release()
 
@@ -203,7 +203,7 @@ class BufferedPipe(object):
         self._lock.acquire()
         try:
             self._closed = True
-            self._cv.notifyAll()
+            notify_all(self._cv)
             if self._event is not None:
                 self._event.set()
         finally:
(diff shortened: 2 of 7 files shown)


demos/rforward.py
```diff
@@ -80,7 +81,7 @@ def reverse_forward_tunnel(server_port, remote_host, remote_port, transport):
         thr = threading.Thread(
             target=handler, args=(chan, remote_host, remote_port)
         )
-        thr.setDaemon(True)
+        set_daemon(thr, True)
```
jun66j5, line 84: We could use simply `thr.daemon = True` with Python 2.7 and 3.

* https://docs.python.org/2/library/threading.html#threading.Thread.daemon
* https://docs.python.org/3/library/threading.html#threading.Thread.daemon

paramiko/buffered_pipe.py
```diff
@@ -101,7 +101,7 @@ def feed(self, data):
             if self._event is not None:
                 self._event.set()
             self._buffer_frombytes(b(data))
-            self._cv.notifyAll()
+            notify_all(self._cv)
```
jun66j5, line 104: We could use simply `self._cv.notify_all()` with Python 2.7 and 3.

* https://docs.python.org/2/library/threading.html#threading.Condition.notify_all
* https://docs.python.org/3/library/threading.html#threading.Condition.notify_all

demos/rforward.py
graingert: Woops I got the description wrong, this is for py2.6 support

jun66j5: The `daemon` attribute is available since Python 2.6 and works.

graingert:
These methods have been supported since 2.6

demos/rforward.py
jun66j5: I've confirmed with Python 2.6 and just confirmed now.

```
$ python2.6
Python 2.6.9 (default, Jul 29 2017, 12:47:14)
[GCC 5.4.0 20160609] on linux4
>>> import threading
>>> import time
>>> t = threading.Thread(target=lambda: time.sleep(1000))
>>> t.daemon = True
>>> t.start()
>>> t.is_alive()
True
>>> #=> it is able to exit without no wait
```

jun66j5: > Wait only 3.3 needs the old method names? It seems the new way is supported from 2.6+? [#1870 (comment)](https://github.com/paramiko/paramiko/pull/1870#issuecomment-1063258752)

I recommend checking documentation. The `daemon` and `notify_all` are available on Python 3.0.

* https://docs.python.org/3.0/library/threading.html#threading.Thread.daemon
* https://docs.python.org/3.0/library/threading.html#threading.Condition.notify_all

PR #2031, opened 2022-04-12
Fix spelling mistake of distributed

bdrung:
Spelling mistake was found by lintian.


bitprophet:
Thank you! I bet that's been there long enough to obtain a US driver's license.

Issue #2027, opened 2022-04-08
use unittest.mock instead of mock


pgajdos:
Could you please consider to utilize built-in unittest.mock instead of mock, as a fallback at least?

bskinn:
Can you explain in more detail the problems you're encountering from the use of `mock` instead of `unittest.mock`?

pghmcfc:
Downstream distributions are trying to get rid of `mock`, e.g.:

https://fedoraproject.org/wiki/Changes/DeprecatePythonMock

Here is the patch I use for the Fedora build of `paramiko`:
https://src.fedoraproject.org/rpms/python-paramiko/blob/rawhide/f/0003-remove-mock-dep.patch

pgajdos:
Sure, we can keep patches temporarily. Thanks for your attention!

bskinn:
For reference, on what timeframe would you hope to be able to drop the patches?

An intermediate step with the guarded import & fallback to `unittest.mock` might be feasible sooner.

pghmcfc:
I'll keep the patch in Fedora as long as it's needed to build without `mock`. I want to be able to build without `mock` because it hasn't been packaged for the upcoming Red Hat Enterprise Linux 9 and I prefer to be able to build the same package on all of my target distribution releases where possible. So there's no big rush for this from my perspective - it's not hard to maintain this patch downstream if I have to.

The previous link won't work any longer as I renumbered the patches when I did the update for 2.10.4, so it's now here:
https://src.fedoraproject.org/rpms/python-paramiko/blob/rawhide/f/0004-remove-mock-dep.patch

The patch already does the guarded import and fallback, except with `unittest.mock` as the preferred option and falling back to `mock` if necessary, I'd be happy to provide the patch as a pull request if that was helpful.

bskinn:
Ok, probably will just aim it for 3.0, then.

Although, a further question: is Fedora packaging older release lines of paramiko? That is, e.g., 2.8 and 2.9 and 2.10? Or, just `latest`? If the former, cherry picking a guarded-import version back to those lines would be important.

A PR with guarded imports would be fantastic...should simplify all of the needed workflow getting this incorporated. Thanks!

pghmcfc:
[...]

PR #2037, opened 2022-04-26
Prefer built-in unittest.mock over external package

pghmcfc:
Prefer and use built-in unittest.mock in Python 3.3+ instead of unnecessarily requiring the external mock package. This helps distributions that are phasing out Python 2 to remove redundant packages.

This is based on #1666 by @mgorny and also addresses #2027.


bskinn:
This PR preferred over #1666 due to the addition of the Python < 3.3 conditional on the `mock` dependency in `dev-requirements.txt`.

bitprophet:
Thanks for this! Though I think it will become moot when we drop Python <3.6 support in the next release: I'll just change everything to unittest.mock, straight up.

PR #2046, opened 2022-05-07
Fixes #2045 Timeouts should be decided with a monotonic clock

### LINKED ISSUE #2045: Timeouts should be decided with a monotonic clock
The clock `time.time()` can be disrupted by changes to the system clock. This can cause unexpected timeouts or prevent a timeout happening.

Instead a monotonic clock such as `time.monotonic()` should be used.

ibygrave:
I've tried to retain Python 2.7 compatibility by using `timeit.default_timer` as the monotonic clock. If Python 2.7 support was not needed this change would be simpler.

### DIFF
--- a/paramiko/auth_handler.py
+++ b/paramiko/auth_handler.py
@@ -59,6 +59,7 @@ from paramiko.common import (
     MSG_USERAUTH_GSSAPI_MIC,
     MSG_NAMES,
     cMSG_USERAUTH_BANNER,
+    timer,
 )
 from paramiko.message import Message
 from paramiko.py3compat import b, u
@@ -234,7 +235,7 @@ class AuthHandler(object):
     def wait_for_response(self, event):
         max_ts = None
         if self.transport.auth_timeout is not None:
-            max_ts = time.time() + self.transport.auth_timeout
+            max_ts = timer() + self.transport.auth_timeout
         while True:
             event.wait(0.1)
             if not self.transport.is_active():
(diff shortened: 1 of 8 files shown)


bskinn:
Python 2 support is being removed in the relatively near term, with a 3.0 release.  Given that this isn't an urgent fix, I'd say go ahead and rewrite in the simpler, Python 3-only fashion, and we'll just aim it for 3.0.

Keep the current git history intact, though, so that if @bitprophet should want to use the Python 2-compatible version he can just cherry-pick from 94d840f or thereabouts.

ibygrave:
Thanks.

For the Python 3 only version would you prefer I keep the `paramiko.common.timer` abstraction, or use `time.monotonic` directly?

bskinn:
That's a call for @bitprophet -- I'll bump the question to his attention.

bitprophet:
Re: Python 2, nah, that's going out the window RSN.

Re: whether it's nice hiding `time.monotonic` behind `paramiko.common.timer`, that's still a good question. My gut says that the way it's phrased right now - exposing the use of `monotonic` - is better for two reasons:

- Just less code/indirection. Always good.
- Makes it slightly more obvious to readers/contributors that "you should be using monotonic for timeout related code". I wasn't even aware that monotonic existed until reading this, though it makes a lot of sense (avoids timeouts from incorrectly firing during time changes like DST, NTP/PTP updates, etc).
    - So having this sprinkled about might clue somebody else in that "oh I should reach for this, not `time.time`, when I care about elapsed durations".

I'll put this in a milestone so there's a chance I'll get to it when crafting 3.0. Thanks!

bitprophet:
Linking to #2045 (doing it in the subject doesn't work, sadly)

PR #2053, opened 2022-05-13
Imlement ctime timestammp attribute

pvannierop:
This PR will implement the `str_ctime` attribute (timestamp last changed).

This is a provisional PR that serves to discuss the implications of this. Be aware that this is untested code. I do not fully understand the implementation, but hope to discuss problems/changes to the implementation where needed.

### DIFF
--- a/paramiko/sftp_attr.py
+++ b/paramiko/sftp_attr.py
@@ -35,6 +35,7 @@ class SFTPAttributes(object):
         - ``st_mode``
         - ``st_atime``
         - ``st_mtime``
+        - ``st_ctime``
 
     Because SFTP allows flags to have other arbitrary named attributes, these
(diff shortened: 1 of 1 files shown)


jun66j5:
This PR is incorrect.

> Be aware that this is untested code.

You should post after testing.

> I do not fully understand the implementation, but hope to discuss problems/changes to the implementation where needed.

OpenSSH sftp server doesn't send ctime field of file.

See https://github.com/openssh/openssh-portable/blob/V_9_0_P1/sftp-common.c#L98-L101

jun66j5:
See also 5. File Attributes in https://www.openssh.com/txt/draft-ietf-secsh-filexfer-02.txt

pvannierop:
@jun66j5 Thank you for clarifying this. I was not aware that openSSH is the limiting factor here. Closing this PR.

PR #2056, opened 2022-05-17
Convert hard-coded block size to constant

maresb:
Closes https://github.com/paramiko/paramiko/issues/1784

### DIFF
--- a/paramiko/sftp.py
+++ b/paramiko/sftp.py
@@ -82,6 +82,9 @@ SFTP_FLAG_CREATE = 0x8
 SFTP_FLAG_TRUNC = 0x10
 SFTP_FLAG_EXCL = 0x20
 
+# Block size used by getfo and putfo in paramiko.sftp_client
+SFTP_FILE_OBJECT_BLOCK_SIZE = 32768
+
 _VERSION = 3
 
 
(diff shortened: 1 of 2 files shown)


bskinn:
Are there any tests that would make sense to add to probe this? I can't imagine it makes sense to actually engage with AWS for it. Maybe a mock, to make sure that `reader.read` is called with the correct block size when it's changed?

----

Separately, it looks like your intended mechanism of usage would be something like this?

```
from paramiko import sftp
sftp.SFTP_FILE_OBJECT_BLOCK_SIZE = {some big number}
```

I think that @bitprophet's input is needed on whether that mechanism is preferred, versus e.g. a passed-parameter paradigm.

Either way, once that's decided the feature will need to be documented

bskinn:
> I am on the road atm, so optimistically the earliest I could write a test is in ~1½ weeks.

No worries, bitprophet's OSS work is on a weekly cycle anyways, and the time is split between paramiko, invoke and fabric, so this'll proceed over a multi-week, possibly multi-month, timeline.

bitprophet:
+1 on configurability, but yea, I think if we're gonna merge something for this I'd rather at least a little time be spent exploring how to do it "right" re: configuration at the Client, Transport, SFTP or SFTPClient level somewhere.

I don't anticipate that would be a whole lot more code, likely a constructor arg somewhere setting an instance attribute + a few passings-on of that instance attribute until it arrives at the spot being changed here.

Issue #1779, opened 2020-11-19
use SFTPClient's get method to download big file from sftp server will block


cfh0081:
Use SFTPClient's get method to download big file from sftp server will block. The error stack information as below:
[sftp_download_error_stack_20201118.txt](https://github.com/paramiko/paramiko/files/5563801/sftp_download_error_stack_20201118.txt)
I found the problem is here:
```
python3.6\Lib\site-packages\paramiko\sftp_file.py

    def _prefetch_thread(self, chunks):
        # do these read requests in a temporary thread because there may be
        # a lot of them, so it may block.
        for offset, length in chunks:
            num = self.sftp._async_request(
                self, CMD_READ, self.handle, long(offset), int(length)
            )
            with self._prefetch_lock:
                self._prefetch_extents[num] = (offset, length)
```
too many requests at same time. And my solution as below:
```
python3.6\Lib\site-packages\paramiko\sftp_file.py

    def _prefetch_thread(self, chunks):
        # do these read requests in a temporary thread because there may be
        # a lot of them, so it may block.
        max_request_num = 512
        to_wait = False
        for offset, length in chunks:
            num = self.sftp._async_request(
                self, CMD_READ, self.handle, long(offset), int(length)
            )
            with self._prefetch_lock:
                self._prefetch_extents[num] = (offset, length)
                if len(self._prefetch_extents) >= max_request_num:
                    to_wait = True
            
            if to_wait:
                time.sleep(1)
                to_wait = False
```
**I hope the official version can fix the problem with an appropriate concurrency number limit and this number is configurable. Or use a better solution to solve the problem.** Thank you!

bitprophet:
Rolling into #2058 - thanks!

Issue #151, opened 2013-03-22
EOFError when getting a big file (900MB) via SFTP.


etienned:
I'm trying to use duplicity to backup my server to a CrushFTP server (windows) in SFTP, but it always drop connection when getting some big files.

So I tried to get a problematic file with paramiko directly and got the same error `Server connection dropped` meaning an `EOFError` was raise (bytes received == 0). I put a print statement in `_read_all` method to show how many bytes were received. On one thread it's getting rapidly to 32777 then stay there. Another thread goes up to 7049 then back to 0 and then never more then 200. After a long long time (many minutes) I receive the `EOFError`.

If I try to get the same file with `lftp` it works without any problems.

Using paramiko 1.10.0, python 2.6.5 on Ubuntu Server LTS 10.0.4 amd64.

kaorihinata:
Not that I'm aware of. I assume the current maintainer mostly uses it for his deployment solution (fabric), so issues with large files on platforms he doesn't (officially?) support anyway don't seem to attract much interest. It's also a pain to reproduce this bug without an appropriate file and server combination, and nobody has come forward to try to troubleshoot it, so I've just avoided using paramiko since it can't be relied on for my purposes.

rsheshadri:
Any suggestions on other modules I can use instead of paramiko?

rsheshadri:
for large files ie

bitprophet:
@kaorihinata is mostly right, though I do my best to look at things from a "pure" paramiko standpoint (i.e. I won't ignore an issue just because it doesn't impact Fabric, even if Fabric-related issues do get more love). The problem here is much more the difficulty in reproducing & the nonstandard platform :(

Always open to merging patches that users say "this fixes my problem X" and which can be proven to not break eg POSIX platforms, but this ticket's not at that stage yet unless I'm missing something.

kaorihinata:
To be honest, I'd probably blame the vendor for their loose interpretation of the standard and dubious definition of "production ready" when it comes to code. It may be the case that OpenSSH works with these servers due to workarounds for broken servers. If I come across the issue again, I will try to determine the cause.

rsheshadri:
@kaorihinata so what do you use currently to transfer files > 1GB? Fabric if so can you point me to a link where I can find sample code to implement a similar get operation using it

kaorihinata:
[...]

Issue #1587, opened 2020-01-06
Issues downloading large files


kkr78:
There are several files on sftp and the file size varies from 100MB to 2GB.  The following code worked on 95% of the files. In most cases, it's an issue downloading large files (>500MB). In some cases, its trouble downloading the same file which succeed in the past. I tried to adjust the window size and packet size but I couldn't make it work for all files. I m looking for a consistent code that works for all files. I followed instructions from here https://stackoverflow.com/questions/45891553/paramiko-hangs-on-get-after-ownloading-20-mb-of-file/48170689#48170689. Am I missing anything here? Any inputs greatly appreciated.

Code

client = paramiko.SSHClient()
client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
client.load_system_host_keys()
client.connect(connection['host'], username=connection['login'], password=connection['password'])
tr = client.get_transport()
tr.default_max_packet_size= 200 * 1024 * 1024
tr.default_window_size= paramiko.common.MAX_WINDOW_SIZE // 3
tr.packetizer.REKEY_BYTES = pow(2, 40)
tr.packetizer.REKEY_PACKETS = pow(2, 40)
sftp = client.open_sftp()

remote_sftp_file, local_sftp_file = '/users/xxx/1.txt', './test.txt' 

sftp.get(remote_sftp_file,local_sftp_file)

Error

[2020-01-03 17:26:53,651] {logging_mixin.py:112} INFO - [2020-01-03 17:26:53,651] {transport.py:1819} ERROR - Socket exception: Connection reset by peer (104)
[2020-01-03 17:26:56,233] {logging_mixin.py:112} WARNING - Exception in thread Thread-3:
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/airflow/.local/lib/python3.6/site-packages/paramiko/sftp_file.py", line 538, in _prefetch_thread
    self, CMD_READ, self.handle, long(offset), int(length)
  File "/usr/local/airflow/.local/lib/python3.6/site-packages/paramiko/sftp_client.py", line 837, in _async_request
    self._send_packet(t, msg)
  File "/usr/local/airflow/.local/lib/python3.6/site-packages/paramiko/sftp.py", line 198, in _send_packet
    self._write_all(out)
[...]

PR #2058, opened 2022-05-17
Allow limiting the number of outstanding prefetch requests

kschoelhorn:
Currently prefetch will send all requests for all chunks for the file in one shot. With the default chunk size of 32k, this can result in many thousand outstanding requests for large files. Some servers like Serv-U 15.2.3.742 seem to be dropping requests after a certain number, which results in the file download hanging indefinitely (or until the server closes the connection).

Fix this issue by letting the user specify a limit for the number of outstanding requests. This is similar to openssh's sftp, which limits the number of outstanding requests to 64 by default.

Note that this is just a quick initial implementation to fix the problem I had while using paramiko with Serv-U. This could also be a setting of the SFTPClient, that limits any `_async_request`s, not just those from prefetching, though it seems the only other user is `listdir_iter`, which already has the `read_aheads` parameter, which also limits the number of requests.

See also #1779 where this was previously proposed and #151, which mixes different issues, but also contains descriptions of get calls hanging, that are fixed by disabling prefetching and are probably the same issue.

### DIFF
--- a/paramiko/sftp_client.py
+++ b/paramiko/sftp_client.py
@@ -758,7 +758,14 @@ class SFTPClient(BaseSFTP, ClosingContextManager):
         with open(localpath, "rb") as fl:
             return self.putfo(fl, remotepath, file_size, callback, confirm)
 
-    def getfo(self, remotepath, fl, callback=None, prefetch=True):
+    def getfo(
+        self,
+        remotepath,
+        fl,
+        callback=None,
+        prefetch=True,
+        max_concurrent_prefetch_requests=None,
+    ):
         """
         Copy a remote file (``remotepath``) from the SFTP server and write to
         an open file or file-like object, ``fl``.  Any exception raised by
@@ -773,6 +780,10 @@ class SFTPClient(BaseSFTP, ClosingContextManager):
             the bytes transferred so far and the total bytes to be transferred
         :param bool prefetch:
             controls whether prefetching is performed (default: True)
+        :param int max_concurrent_prefetch_requests:
+            The maximum number of concurrent read requests to prefetch.
+            When this is ``None`` (the default), do not limit the number of
+            concurrent prefetch requests.
         :return: the `number <int>` of bytes written to the opened file object
 
         .. versionadded:: 1.10
@@ -782,12 +793,19 @@ class SFTPClient(BaseSFTP, ClosingContextManager):
         file_size = self.stat(remotepath).st_size
         with self.open(remotepath, "rb") as fr:
             if prefetch:
-                fr.prefetch(file_size)
+                fr.prefetch(file_size, max_concurrent_prefetch_requests)
             return self._transfer_with_callback(
                 reader=fr, writer=fl, file_size=file_size, callback=callback
             )
 
-    def get(self, remotepath, localpath, callback=None, prefetch=True):
+    def get(
+        self,
+        remotepath,
+        localpath,
+        callback=None,
+        prefetch=True,
+        max_concurrent_prefetch_requests=None,
+    ):
         """
         Copy a remote file (``remotepath``) from the SFTP server to the local
         host as ``localpath``.  Any exception raised by operations will be
@@ -800,6 +818,10 @@ class SFTPClient(BaseSFTP, ClosingContextManager):
             the bytes transferred so far and the total bytes to be transferred
         :param bool prefetch:
             controls whether prefetching is performed (default: True)
+        :param int max_concurrent_prefetch_requests:
+            The maximum number of concurrent read requests to prefetch.
+            When this is ``None`` (the default), do not limit the number of
+            concurrent prefetch requests.
 
         .. versionadded:: 1.4
         .. versionchanged:: 1.7.4
@@ -808,7 +830,13 @@ class SFTPClient(BaseSFTP, ClosingContextManager):
             Added the ``prefetch`` keyword argument.
         """
         with open(localpath, "wb") as fl:
-            size = self.getfo(remotepath, fl, callback, prefetch)
+            size = self.getfo(
+                remotepath,
+                fl,
+                callback,
+                prefetch,
+                max_concurrent_prefetch_requests,
+            )
         s = os.stat(localpath)
         if s.st_size != size:
             raise IOError(
--- a/paramiko/sftp_client.py
+++ b/paramiko/sftp_client.py
@@ -783,7 +783,10 @@ class SFTPClient(BaseSFTP, ClosingContextManager):
         :param int max_concurrent_prefetch_requests:
             The maximum number of concurrent read requests to prefetch.
             When this is ``None`` (the default), do not limit the number of
-            concurrent prefetch requests.
+            concurrent prefetch requests. Note: OpenSSH's sftp internally
+            imposes a limit of 64 concurrent requests, while Paramiko imposes
+            no limit by default; consider setting a limit if a file can be
+            successfully received with sftp but hangs with Paramiko.
         :return: the `number <int>` of bytes written to the opened file object
 
         .. versionadded:: 1.10
@@ -821,7 +824,10 @@ class SFTPClient(BaseSFTP, ClosingContextManager):
         :param int max_concurrent_prefetch_requests:
             The maximum number of concurrent read requests to prefetch.
             When this is ``None`` (the default), do not limit the number of
-            concurrent prefetch requests.
+            concurrent prefetch requests. Note: OpenSSH's sftp internally
+            imposes a limit of 64 concurrent requests, while Paramiko imposes
+            no limit by default; consider setting a limit if a file can be
+            successfully received with sftp but hangs with Paramiko.
 
         .. versionadded:: 1.4
         .. versionchanged:: 1.7.4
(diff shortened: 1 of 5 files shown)


bskinn:
Looks like this is specific to SFTP?

paramiko/sftp_file.py
```diff
@@ -522,18 +530,26 @@ def _get_size(self):
...
-    def _prefetch_thread(self, chunks):
+    def _prefetch_thread(self, chunks, max_outstanding_requests):
         # do these read requests in a temporary thread because there may be
         # a lot of them, so it may block.
         for offset, length in chunks:
+            # Limit the number of outstanding requests in a busy-loop
+            if max_outstanding_requests is not None:
+                while True:
+                    with self._prefetch_lock:
+                        if len(self._prefetch_extents) < max_outstanding_requests:
```
bskinn, line 549: Checking my understanding of the implementation here -- does each call to `self.sftp._async_request()` cause `self._prefetch_extents` to increment by one, and then once each chunk is received, `._prefetch_extents` is decremented?

And, thus, this new code causes this queuing within `self._prefetch_thread()` to idle whenever we hit our max-requests cap?

paramiko/sftp_file.py
```diff
@@ -473,9 +477,9 @@ def prefetch(self, file_size=None):
             chunks.append((n, chunk))
             n += chunk
         if len(chunks) > 0:
-            self._start_prefetch(chunks)
+            self._start_prefetch(chunks, max_outstanding_requests)
 
-    def readv(self, chunks):
+    def readv(self, chunks, max_outstanding_requests = None):
```
bskinn, line 482: It looks like you haven't added any use of `max_outstanding_requests` in `readv()` here -- seems like either it shouldn't be added, or there's a piece of implementation missing.

bskinn (changes requested):
Might be a piece of implementation missing

bskinn:
Will need bitprophet's input on that parameter name, I think -- paramiko tends toward more concise names.

bskinn:
@kschoelhorn Are there any tests that could be added for this, that would be less onerous than mocking a server and monitoring the length of the request queue?

bskinn:
Per https://github.com/paramiko/paramiko/issues/1587#issuecomment-657647904 and the [SO post](https://stackoverflow.com/a/58339631) linked therein, this feature might mitigate or solve #1587?

The workaround posted in [this Gist](https://gist.github.com/vznncv/cb454c21d901438cc228916fbe6f070f), linked from https://github.com/paramiko/paramiko/issues/1587#issuecomment-1137274920, suggests it likely would help.

The implementation in this PR adds a parameter to the public API, versus the private class/member implementation in the Gist, so this PR is likely preferred.

paramiko/sftp_file.py
kschoelhorn: Good catch! I meant to forward the parameter to `_start_prefetch`.

paramiko/sftp_file.py
kschoelhorn: The new async request number returned by `_async_request` is added to `_prefetch_extents` by the code below (hidden by the diff view on github by default:

```
            with self._prefetch_lock:
                self._prefetch_extents[num] = (offset, length)
```

The request number is then removed again in `_async_response`.

So yes, the background thread just idles here if the limit has been reached.

kschoelhorn:
> @kschoelhorn Are there any tests that could be added for this, that would be less onerous than mocking a server and monitoring the length of the request queue?

I'll think about some tests. I think we should be able to mock `SFTPClient` and then simulate a large file transfer.

bwinston-sdp:
Is this still a possible PR pending tests? I've run into a similar situation as mentioned in the issues/Gist linked, and this works beautifully. I'm happy to rebase/write some tests if @kschoelhorn is MIA and/or wants to pass it off.

bskinn:
> Is this still a possible PR pending tests?

Yes, it definitely is -- it's not at the top of the priority list for paramiko at the moment, though, so it hasn't seen any specific action.

Up to you and @kschoelhorn to negotiate the test work. Whoever would like to contribute them is welcome!

bwinston-sdp:
Not sure if it was the "right way" to do it, but I created a PR to @kschoelhorn's branch, rebased on main, with lint fixes and an added test. LMK if I should do it another way (create a new PR directly to paramiko with changes, reference/credit this one?).

kschoelhorn:
@bwinston-sdp: Thank you very much for the test, I'll check and merge to this PR later!

bwinston-sdp:
Thanks @kschoelhorn!! I figured the test might be flaky because of the background thread, but wasn't sure exactly how to ensure stability (I've not done much multi-threaded coding).

One other thing I thought of -- I know the `max_outstanding_requests` gets passed through `readv` as well, not sure if we need another test for that.

kschoelhorn:
> One other thing I thought of -- I know the `max_outstanding_requests` gets passed through `readv` as well, not sure if we need another test for that.

Both use the same prefetch code internally, so I think one test is enough.

paramiko/sftp_client.py
```diff
@@ -773,6 +780,10 @@ def getfo(self, remotepath, fl, callback=None, prefetch=True):
             the bytes transferred so far and the total bytes to be transferred
         :param bool prefetch:
             controls whether prefetching is performed (default: True)
+        :param int max_outstanding_prefetch_requests:
+            Limit the number of outstanding requests instead of sending all
```
bskinn, line 784: Is this parameter defining the maximum _outstanding_ requests, or the max _concurrent_ requests? It seems more the latter to me.

Separately, this parameter is an `int` -- its description should explicitly say how that value is used. It's _implicit_ that the value is the max number of concurrent requests, but it should be made explicit.

bskinn (changes requested):

bskinn:
Also--please add a CHANGELOG entry to `/sites/www/changelog.rst` -- thanks!

bwinston-sdp:
I updated these and pushed to the PR on @kschoelhorn's branch.

bskinn:
:+1: - Linking directly to https://github.com/kschoelhorn/paramiko/pull/1, for reference.

paramiko/sftp_client.py
kschoelhorn: I meant outstanding in the sense of "not yet done": the requests were sent, but we have not yet received an answer. But concurrent is also fine! :slightly_smiling_face:

bskinn: @kschoelhorn Yeah, that was what prompted my suggestion: to my mind, 'outstanding' could be ambiguous, meaning either:

1. The number of requests currently in-flight (the intended meaning)
2. The total number of requests yet to be completed, including both in-flight and not-yet-sent requests (not the intended meaning)

Meaning (2) doesn't make a whole lot of sense, practically, but if that's the sense of the word that someone reaches first, then the name could be confusing. Thus, my recommendation for 'concurrent'.

kschoelhorn:
Thanks again to @bwinston-sdp! I merged your changes, rebased to latest master and fixed a few lints.

@bskinn: I guess you can remove the "Needs changelog/docs" flag now. 🙂

bskinn:
> I guess you can remove the "Needs changelog/docs" flag now. 🙂

Done!

paramiko/sftp_file.py
```diff
@@ -485,6 +492,13 @@ def readv(self, chunks):
         :param chunks:
             a list of ``(offset, length)`` tuples indicating which sections of
             the file to read
+        :param int max_concurrent_prefetch_requests:
+            The maximum number of concurrent read requests to prefetch.
```
bitprophet, line 496: I'd replace this param description everywhere but `SFTPClient.get`, with something like:

```
The maximum number of concurrent read requests to prefetch. See the API docs for `.SFTPClient.get` for details.
```

(IOW I don't think it's useful to have the same paragraph in all 4+ places :))

TallGibbs:
Sorry if this seems really silly, but all this testing and (proposed) change seems like exactly what I need to fix my current issue in some SFTP code.  Did this change get implemented?  Is there documentation for how to specify the "max_outstanding_prefetch_requests" variable?  I am trying to transfer 1GB file through SFTP and the host is closing the connection with the "EOFError"

bwinston-sdp:
@TallGibbs this feature is included in 3.3, with an additional parameter in the `get()` function: https://docs.paramiko.org/en/latest/api/sftp.html?highlight=prefetch#paramiko.sftp_client.SFTPClient.get

TallGibbs:
> @TallGibbs this feature is included in 3.3, with an additional parameter in the `get()` function: https://docs.paramiko.org/en/latest/api/sftp.html?highlight=prefetch#paramiko.sftp_client.SFTPClient.get

Oh awesome!  Looks like I need to update then....I'm currently using 2.8.1

Thank you for verifying

TallGibbs:
@bwinston-sdp I just updated to 3.4.0 and modified my code to include get(file, local_path, max_concurrent_prefetch_requests=32) and it worked like a charm!  Super simple implementation (WAY easier than all the solutions on S.O.).  This was a file just under 1 GB, for anyone else encountering these issues.

Thanks to everyone involved for solving this SFTP problem

Issue #2063, opened 2022-06-01
Treat EOFError in early connection case differently


Anthchirp:
<details>
  <summary>When connecting to a busy OpenSSH server, or a server with a low `MaxStartups` setting, the server can reject incoming unauthenticated connections.
</summary>
  
```
     MaxStartups
             Specifies the maximum number of concurrent unauthenticated
             connections to the SSH daemon.  Additional connections will
             be dropped until authentication succeeds or the
             LoginGraceTime expires for a connection.  The default is
             10:30:100.

             Alternatively, random early drop can be enabled by
             specifying the three colon separated values start:rate:full
             (e.g. "10:30:60").  sshd(8) will refuse connection attempts
             with a probability of rate/100 (30%) if there are currently
             start (10) unauthenticated connections.  The probability
             increases linearly and all connection attempts are refused
             if the number of unauthenticated connections reaches full
             (60).
```
</details>

On the client this results in an `EOFError` during reading, which is then caught and replaced with an `SSHException` saying "**Error reading SSH protocol banner**":
https://github.com/paramiko/paramiko/blob/0f888a5a9d4bcfdc1d98296ad165d7da2a7e63db/paramiko/packet.py#L603-L609
https://github.com/paramiko/paramiko/blob/0f888a5a9d4bcfdc1d98296ad165d7da2a7e63db/paramiko/transport.py#L2270-L2277

While this is technically correct, I don't find it very helpful for debugging or instructive to the user, and it's not easy to distinguish the error programmatically either. Could we treat `EOFError`s differently and return an `SSHException` with a different message, eg. "**Connection rejected by server**" if nothing has been read so far?

I'm happy to make a PR to this effect.

bskinn:
Seems to me like a reasonable feature to add! I've flagged for bitprophet's input.

jun66j5:
I don't think your suggestion is reasonable.

The `_check_banner()` method is used on both client mode and server mode. Especially, the `"Connection rejected by server"` message  is not reasonable for server mode.

Anthchirp:
I think your suggestion would be a user experience improvement 👍 
Do you want me to prepare a PR?

Out of interest - why would this need `#noqa`?

bskinn:
You can go ahead and work on a PR, @Anthchirp, sure -- up to you whether or not you want to wait for @bitprophet to opine.

jun66j5:
@Anthchirp

Yes, could you please create pullreq if possible?

[...]

PR #2066, opened 2022-06-06
Improve exception's message when connection closed while reading protocol banner

jun66j5:
Proposed changes for #2063.


bskinn:
Thanks, @jun66j5! This probably calls for a CHANGELOG entry, can you add?

jun66j5:
See #2100

Issue #1078, opened 2017-09-28
Sporadic exception when exiting with sys.exit


Noctevent:
Around 20% of the time, my script which uses paramiko for a simple ssh command:

```
ssh = SSHClient()
ssh.load_system_host_keys()
try:
    ssh.connect(machine, username=user, password=password)
    stdin, stdout, stderr = ssh.exec_command(commande)

    return_code = stdout.channel.recv_exit_status()

    out = stdout.read().decode('utf-8')
    err = stderr.read().decode('utf-8')

    if out.strip():
        print(out.strip())
    if err.strip():
        print(err.strip())

    # stderr.close()
    # stdout.close()
    # stdin.close()

    # print("CR=" + str(code_retour))
    print("CR=0")
    rc = 0

except SSHException as e:
    print(e)
    rc = -2
finally:
    ssh.close()

sys.exit(cr)
```
Spews out those exceptions:
```
Exception ignored in: <object repr() failed>
Traceback (most recent call last):
  File "/home/user/.miniconda2/envs/py3/lib/python3.5/site-packages/paramiko/file.py", line 62, in __del__
  File "/home/user/.miniconda2/envs/py3/lib/python3.5/site-packages/paramiko/file.py", line 80, in close
  File "/home/user/.miniconda2/envs/py3/lib/python3.5/site-packages/paramiko/file.py", line 89, in flush
TypeError: 'NoneType' object is not callable
Exception ignored in: <object repr() failed>
Traceback (most recent call last):
  File "/home/user/.miniconda2/envs/py3/lib/python3.5/site-packages/paramiko/file.py", line 62, in __del__
  File "/home/user/.miniconda2/envs/py3/lib/python3.5/site-packages/paramiko/file.py", line 80, in close
  File "/home/user/.miniconda2/envs/py3/lib/python3.5/site-packages/paramiko/file.py", line 89, in flush
TypeError: 'NoneType' object is not callable
Exception ignored in: <object repr() failed>
Traceback (most recent call last):
  File "/home/user/.miniconda2/envs/py3/lib/python3.5/site-packages/paramiko/file.py", line 62, in __del__
  File "/home/user/.miniconda2/envs/py3/lib/python3.5/site-packages/paramiko/file.py", line 80, in close
  File "/home/user/.miniconda2/envs/py3/lib/python3.5/site-packages/paramiko/file.py", line 89, in flush
TypeError: 'NoneType' object is not callable
```
[...]

PR #2068, opened 2022-06-16
Sporadic exception when exiting with sys.exit

marbad1994:
Sometimes when exiting there is a TypeError which
prints: Exception ignored in: <function BufferedFile.__del__ at ...>
TypeError: 'NoneType' object is not callable
Add a try/except statement to silence the output

Fix: #1078

### DIFF
--- a/paramiko/file.py
+++ b/paramiko/file.py
@@ -58,7 +58,10 @@ class BufferedFile (ClosingContextManager):
         self._size = 0
 
     def __del__(self):
-        self.close()
+        try:
+            self.close()
+        except TypeError:
+            pass
         
     def __iter__(self):
         """


bskinn:
Ah, sure -- if this really is just a nuisance error, this is one way to silence it.

Looking at the traceback in [one of the comments to the linked issue](https://github.com/paramiko/paramiko/issues/1078#issuecomment-596257424), I wonder if it might be better to fix this deeper in the code... looks like it might be a bug in `message.py`. Decision for bitprophet.

marbad1994:
> Ah, sure -- if this really is just a nuisance error, this is one way to silence it.
> 
> Looking at the traceback in [one of the comments to the linked issue](https://github.com/paramiko/paramiko/issues/1078#issuecomment-596257424), I wonder if it might be better to fix this deeper in the code... looks like it might be a bug in `message.py`. Decision for bitprophet.

Well I don't think it's that simple. I have deep dived into this and can't seem to find the source of the TypeError anywhere. Regarding the post you are referring to, I have similar error output only difference is that it points to the filter class for the logging instead of Message.

But as you said, we'll need a decision whether it's worth to look further into or if we just should silence the error, this way.

jun66j5:
I don't think erasing `TypeError` in `__del__` method is good idea.

The root cause is that `BytesIO` reference is garbage collected from `paramiko/file.py` on Python interpreter termination but the `BytesIO()` is invoked `__del__` method.

Ideally, stdin, stdout, stderr objects from `SSHClient.exec_command` should be closed before closing instance of `SSHClient` by paramiko user.

Work around for the issue (untested because unable to reproduce on my environment):

```diff
--- a/paramiko/file.py
+++ b/paramiko/file.py
@@ -90,7 +90,10 @@ class BufferedFile(ClosingContextManager):
         buffering is not turned on.
         """
         self._write_all(self._wbuffer.getvalue())
-        self._wbuffer = BytesIO()
+        # Prevent "TypeError: 'NoneType' object is not callable" from
+        # invoking BytesIO() on Python interpreter termination
+        self._wbuffer.seek(0)
+        self._wbuffer.truncate(0)
         return

     if PY2:
```

marbad1994:
> ```diff
> self._wbuffer.seek(0)
> ```

That might a good point. I tested your solution and unfortunately it didn't solve the problem.
I still get Exception ignored in: <function BufferedFile.__del__ at 0x7fc292d73320> and TypeError: 'NoneType' object is not callable.

jun66j5:
Thanks for the testing. Hmmm, I try to reproduce it again.

PR #2084, opened 2022-07-13
Fix SFTPClient.getfo() docstring

cvubrugier:
In SFTPClient.getfo(), the "remotepath" argument should be a string
whereas "fl" should be a file-like object used for destination.

Also document the type of the "fl" argument in SFTPClient.putfo().


paramiko/sftp_client.py
```diff
@@ -692,7 +692,7 @@ def putfo(self, fl, remotepath, file_size=0, callback=None, confirm=True):
 
         The SFTP operations use pipelining for speed.
 
-        :param fl: opened file or file-like object to copy
+        :param fl object: opened file or file-like object to copy
```
jun66j5, line 695:
```suggestion
        :param object fl: opened file or file-like object to copy
```

cvubrugier: Thank you @jun66j5 for your code review.
You are correct: the type declaration should precede the argument name in the docstring.
This should be fixed now.

PR #1456, opened 2019-06-12
Host Based Authentication

calccrypto:
Updated version of https://github.com/bws/xdd/blob/master/contrib/paramiko-hostbased-auth.diff
Permission has been obtained from author.

PR #2093, opened 2022-07-26
Host based authentication

asomers:
Updated version of https://github.com/paramiko/paramiko/pull/1456 with
Python 3 support, ecdsa and ed25519 support, and a few other fixes.


asomers:
cc @calccrypto @bskinn

bskinn:
Thanks for creating this updated PR, @asomers, really appreciate it.

bitprophet's next priority for paramiko is dropping Python 2 and older versions of Python 3.x --- I believe he plans to keep 3.6+ compat. This PR almost certainly won't get looked into until after that point.

I mention this for two reasons: one, to let you know that it's going to be some time before it gets any attention, and I apologize for that. Two, since Python 2 is exiting of the picture, it might change your implementation?

Also, this seems potentially related to #387 ... AFAIK that's the *next* priority after the drop of Python 2. So, it might get swept up into that general overhaul. Will have to see, though.

asomers:
How might it change the implementation, @bskinn ?  I didn't design anything here.  I just copied the old PR and fixed its bugs.

bskinn:
I don't know -- I haven't looked at the changes in detail yet. In https://github.com/paramiko/paramiko/pull/1456#issuecomment-1170505944 you mentioned some Python 2/3 issues, so I thought that the impending drop of Py2 might be relevant. Sounds like it's not, which is great. :slightly_smiling_face:

PR #2098, opened 2022-08-05
Fix TypeError in output on Windows

opain-replika:
TypeError: write() argument must be str, not byte


bskinn:
Per guidance from bitprophet, I'm closing all issues/PRs related to demos as `wontfix`, as he plans to remove all of the demos from the paramiko codebase in a future release.

Please consider [Fabric]'s various uses of paramiko to be the most current available examples/demos.

If there's a common paramiko use case that could use demonstration code and that's not already served by Fabric's API, please open a feature-request issue to suggest it, or a PR to add it.

[fabric]: https://github.com/fabric/fabric

PR #2100, opened 2022-08-18
Improve exception message when connection closed while reading protocol banner

### LINKED ISSUE #2063: Treat EOFError in early connection case differently

Anthchirp:
Now with added changelog entry.

Fixes #2063, closes #2063.

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -2271,9 +2271,13 @@ class Transport(threading.Thread, ClosingContextManager):
                 buf = self.packetizer.readline(timeout)
             except ProxyCommandFailure:
                 raise
+            except EOFError:
+                raise SSHException(
+                    "Connection closed by remote host while reading SSH protocol banner"  # noqa
+                )
             except Exception as e:
                 raise SSHException(
-                    "Error reading SSH protocol banner" + str(e)
+                    "Error reading SSH protocol banner: " + str(e)
                 )
             if buf[:4] == "SSH-":
                 break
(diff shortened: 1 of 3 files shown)


Anthchirp:
Could we get this reviewed/merged?

PR #2103, opened 2022-08-22
add `Ed25519Key.generate()` and drop requirement of nacl

boyxuper:
note: cryptography>=2.6 has ed25519 support, so maybe we can boost the version requirement to 2.6 for merge test?

### DIFF
--- a/paramiko/ed25519key.py
+++ b/paramiko/ed25519key.py
@@ -15,16 +15,24 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA.
 
 import bcrypt
-
+from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.asymmetric import ed25519
 from cryptography.hazmat.primitives.ciphers import Cipher
-
-import nacl.signing
 
 from paramiko.message import Message
 from paramiko.pkey import PKey, OPENSSH_AUTH_MAGIC, _unpad_openssh
-from paramiko.util import b
 from paramiko.ssh_exception import SSHException, PasswordRequiredException
+from paramiko.util import b
+
+
+def public_bytes_raw(verify_key):
+    try:
+        # for cryptography>=40.0, new API enabled
+        return verify_key.public_bytes_raw()
+    except AttributeError:
+        # for backward support
+        return verify_key._raw_public_bytes()
 
 
 class Ed25519Key(PKey):
@@ -42,7 +50,13 @@ class Ed25519Key(PKey):
     name = "ssh-ed25519"
 
     def __init__(
-        self, msg=None, data=None, filename=None, password=None, file_obj=None
+        self,
+        msg=None,
+        data=None,
+        filename=None,
+        password=None,
+        file_obj=None,
+        vals=None,
     ):
         self.public_blob = None
         verifying_key = signing_key = None
@@ -54,7 +68,11 @@ class Ed25519Key(PKey):
                 key_type=self.name,
                 cert_type="ssh-ed25519-cert-v01@openssh.com",
             )
-            verifying_key = nacl.signing.VerifyKey(msg.get_binary())
+            verifying_key = ed25519.Ed25519PublicKey.from_public_bytes(
+                msg.get_binary()
+            )
+        elif vals:
+            signing_key, verifying_key = vals
         elif filename is not None:
             with open(filename, "r") as f:
                 pkformat, data = self._read_private_key("OPENSSH", f)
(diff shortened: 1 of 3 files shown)


bskinn:
The next unit of work for paramiko is planned to be the drop of Python 2 and older 3.x, after which will be the major key/auth overhaul indicated by #387.

Please correct me if I'm wrong, but it looks like this PR would be under the umbrella of the auth overhaul? If so, we'll flag this as something to consider as part of that effort.

Either way, thank you for the contribution! We'll get it worked in as soon as possible...no firm idea of what that timeline will be, though.

utkonos:
I figured I'd add the following over here too in case someone is searching for a way to go ahead and generate ed25519 keys and use the result with `paramiko` right now and not wait.

The following code snippet works for ed25519 key generation using `cryptography` and then using `io.StringIO` as an interface to get the generated key into `paramiko`. From there, you're off to the races.

```python
import io

from cryptography.hazmat.primitives import serialization
from cryptography.hazmat.primitives import asymmetric
import paramiko

c_ed25519key = asymmetric.ed25519.Ed25519PrivateKey.generate()
privpem = c_ed25519key.private_bytes(encoding=serialization.Encoding.PEM,
                                     format=serialization.PrivateFormat.OpenSSH,
                                     encryption_algorithm=serialization.NoEncryption())
print(privpem.decode())

priv_obj = io.StringIO(privpem.decode())
p_ed25519key = paramiko.ed25519key.Ed25519Key.from_private_key(priv_obj)
print(p_ed25519key.get_base64())

pub = c_ed25519key.public_key()
openssh_pub = pub.public_bytes(encoding=serialization.Encoding.OpenSSH,
                               format=serialization.PublicFormat.OpenSSH)
print(openssh_pub)
```

This and other info is over in
https://github.com/paramiko/paramiko/issues/1136

bskinn:
Thanks very much for the research and implementation here, @utkonos! I expect they'll be helpful once the key/auth overhaul (#387) gets underway.

utkonos:
I don't think much needs to be done to change `Ed25519Key` to add a `generate()` method or allow key material to be a direct input parameter to that class. The former is probably 3-4 lines of code maximum, and the latter is just as few. Right now the class object can accept key material in `PEM` format and once in there the key material is stored exactly as it is stored in the object that `cryptography` emits. I'm sure there are bureaucratic reasons to wait, but I don't see any technical reasons to wait.

No worries tho. Proceed as you see fit.

bskinn:
Nod, it does seem like a fairly small, straightforward change. But, I'm not deeply familiar with the codebase---I'm just helping triage---so I hesitate to say anything authoritative.

In any event, it'll be bitprophet's call how he wants to tackle it. If the key handling system is going to be substantially different from the current version, I could see it making more sense only taking the time to implement and release it the new way.

utkonos:
Totally understand. I can see that the PR that we're taking under touches quite a few things. I would propose a much more narrow PR that sinply adds a `generate()` method and nothing else. Right now, one has to manually get the key material into the class through a complex pathway, but once the key material is in there, everything else is the same. `paramiko` is and has been setup to use this key type for a while, there should be no other code changes required. The removal of `nacl` and any other code changes can wait for an overhaul.

But again, no worries. There is always monkey patching. Which is what I'm looking at next.

utkonos:
I am incorrect about the above. All the `nacl` code does need to be removed for this all to work correctly. The contents of `pkey.signing_key` is a `nacl` object. Nothing will work unless `nacl` is removed completely and `cryptography` is coded in place of it. However, I was able to do a quick monkeypatch that just adds the bare minimum code using `nacl` to get the job done.

image

bskinn:
Thanks again for all your work here, @utkonos -- it'll be a big help.

bitprophet:
Curious how this and #1806 overlap 🤔 I may try folding this into the 3.0 release after all (it'd let us remove a dependency outright!), but per Brian's notes, this may instead wait for a later feature-oriented release.

boyxuper:
@bitprophet just rebased on v4.0. can we review this after years? :D

paramiko/ed25519key.py
```diff
@@ -42,7 +50,13 @@ class Ed25519Key(PKey):
...
 
     def __init__(
-        self, msg=None, data=None, filename=None, password=None, file_obj=None
+        self,
+        msg=None,
+        data=None,
+        filename=None,
+        password=None,
+        file_obj=None,
+        vals=None,
```
toabctl, line 59: do we really need `vals` here and change the constructor signature?

paramiko/ed25519key.py
```diff
@@ -15,16 +15,24 @@
...
-import nacl.signing
-
 from paramiko.message import Message
 from paramiko.pkey import PKey, OPENSSH_AUTH_MAGIC, _unpad_openssh
-from paramiko.util import b
 from paramiko.ssh_exception import SSHException, PasswordRequiredException
+from paramiko.util import b
+
+
+def public_bytes_raw(verify_key):
```
toabctl, line 29: this is not required I think. it simply calls
```
self.public_bytes(
            _serialization.Encoding.Raw, _serialization.PublicFormat.Raw
        )
```

so we can either do the same or create our own helper method instead of calling a private method.

toabctl:
@boyxuper could you rebase? @bitprophet is this something that would be accepted?

PR #2106, opened 2022-08-24
Support forwarding pageant

Tomeamis:
Currently, when paramiko is running from Windows, and Pageant is used as ssh agent, it fails with `AuthenticationException("Unable to connect to SSH agent")`. This PR enables successfully forwarding Pageant.

I only altered the `_communicate` method for `AgentRemoteProxy` seeing as I have no idea what `AgentLocalProxy` is for. I could also make it so that it is only used with Pageant, but I tested it on Linux (WSL, but still) and it seemed fine.

### DIFF
--- a/paramiko/agent.py
+++ b/paramiko/agent.py
@@ -118,6 +118,12 @@ class AgentProxyThread(threading.Thread):
         self._agent = agent
         self._exit = False
 
+    def _agent_is_pageant(self):
+        if sys.platform != "win32":
+            return False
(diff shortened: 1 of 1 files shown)


Tomeamis:
Also, I'm not the only one who hit this, see e.g. fabric/fabric#1621

ALERTua:
anyone?

Tomeamis:
So, is anyone going to look at this, or...?

Issue #2105, opened 2022-08-23
_check_banner timeout for subsequent lines


matt-leolabs:
Hello,

I am doing ssh over ssm, and while I can set the banner_timeout for the first line, the subsequent lines still fail unless I increase the timeout there. The error I get below has been discussed many times, but I am able to fix it only if I overwrite the timeout in the `else` clause of the `_check_banner` method here: https://github.com/paramiko/paramiko/blob/main/paramiko/transport.py#L2269

Why doesn't the else clause also respect the banner timeout?

```Exception: Error reading SSH protocol banner
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/paramiko/transport.py", line 2139, in _check_banner
    buf = self.packetizer.readline(timeout)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/paramiko/packet.py", line 367, in readline
    buf += self._read_timeout(timeout)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/paramiko/packet.py", line 577, in _read_timeout
    raise socket.timeout()
socket.timeout

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/paramiko/transport.py", line 1966, in run
    self._check_banner()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/paramiko/transport.py", line 2144, in _check_banner
    "Error reading SSH protocol banner" + str(e)
paramiko.ssh_exception.SSHException: Error reading SSH protocol banner

Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/paramiko/transport.py", line 2139, in _check_banner
    buf = self.packetizer.readline(timeout)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/paramiko/packet.py", line 367, in readline
    buf += self._read_timeout(timeout)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/paramiko/packet.py", line 577, in _read_timeout
    raise socket.timeout()
socket.timeout

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #2107, opened 2022-08-24
use banner timeout for subsequent lines

matt-leolabs:
To address https://github.com/paramiko/paramiko/issues/2105

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -237,6 +237,7 @@ class SSHClient(ClosingContextManager):
         gss_trust_dns=True,
         passphrase=None,
         disabled_algorithms=None,
+        use_banner_timeout_for_other_lines=None
     ):
         """
         Connect to an SSH server and authenticate to it.  The server's host key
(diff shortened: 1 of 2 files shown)


bskinn:
Seems like a pretty straightforward approach -- it'll be bitprophet's call on whether he likes the implementation or not.

Are there any easy tests you could add for this?

You'll need to add documentation to the docstrings for the new parameter to `connect()` and the new member of `Transport`.

Thanks for the PR!

matt-leolabs:
Thank you for the response and tags! I will add documentation and look into adding some tests shortly.

matt-leolabs:
@bskinn I've added the requested docs.

In terms of adding tests, I think the most "straightforward" way would be to monkey-patch the `readline` method of `self.packetizer` in `transport.py`, but that doesn't seem very elegant. Curious to hear your thoughts on this as well.

bskinn:
@matt-leolabs -- thanks for the docs adds, AFAICT they look good!

Unfortunately, I'm not actually all that familiar with the paramiko internals...I'm just doing my best to help bitprophet triage. So, I can't really advise you. If a sensible test would be too tricky to put in place, then I'd say just don't worry about it for now. bitprophet will make that call once he takes a look.

Which, unfortunately, will probably not be very soon -- he's neck-deep in Python 2 removal and Sphinx modernizing in his various projects, and won't be moving on to feature requests until after that's all taken care of.

We *will* get there...just can't say when.

Issue #2054, opened 2022-05-13
Make Transport cipher/algorithm choice lists fully user-configurable


bitprophet:
This is an age old problem: users need ability to shuffle the order of, or flat out remove, certain options for kex, hostkey checks, pubkey auth offering, and so forth. This isn't currently possible without editing your Paramiko or doing monkeypatching on your Transport class or instance. A weak bone was thrown to this with `disabled_algorithms` recently but it's just a mediocre band-aid by all accounts.

Transport needs updating so that these lists are easy to override/modify/etc:

* Minimum: straight up overwriting w/o monkeypatching, which allows most of the below, if not super user friendly (outside of the cases where one truly wants a very small list of supported algos)
* Nicer: easy/ier API for removing problematic algorithms, eg "drop Blowfish so Cryptography shuts up (n.b. this is actually going to be moot soon but a good example of the issue)" or "drop SHA1-based stuff to force use of SHA2". `disabled_algorithms` allows this but is a bit clunky due to use of dicts. Maybe we want stuff like `Transport.remove_algorithms()` or w/e.
* Maybe: allow shuffling-without-removal, eg shift undesirable algorithms lower down the list. I don't see when one would want this over straight removal or overwriting though.
* Extra credit: build in some easy way to bridge SSHConfig settings for these, though that's a higher level problem yet unsolved. Probably best to start with one of the other bullet points + Fabric consuming the new API instead, since that already works.

Related:

* #387 insofar as that's another moderate sized auth/connection related shakeup badly overdue
* #1527 and #1652 are reverse-duplicates of this (I thought we had a ticket for this ages ago but ... can't find it?)

bskinn:
Believe this also covers #2049?

last-partizan:
I've came here after reading #1961, and instead of creating new issues i'm adding comment here.

>  If anyone can identify legit issues with the new feature that are not fixable by using disabled_algorithms (or pinning to 2.8.x until one's servers are upgraded - absolutely an acceptable solution!)

We having many junos devices in our network, some of them new, some of them old. Not all can be upgraded. And we're using [py-junos-eznc](https://github.com/Juniper/py-junos-eznc) with ssh transport to comminicate with those devices. It uses netconf which uses paramiko.

After updating to latest `paramiko` we can't connect to some junos devices. For now we're pinned paramiko to 2.8, but please, consider using some fallback algorithm to allow seamlessly connecting to old hardware.

```
[...]

PR #2125, opened 2022-10-07
Add Transport factory to SSHClient.connect

chiefnoah:
Adds a transport_factory argument to `SSHClient.connect` that allows you to dynamically generate a Transport instance without, and therefore modify inner connection parameters before a connection gets established.

This should address some of the issues in #2054 with minial changes to the API and no changes to Transport while allowing for arbitrary control over Transports API.

An example of how to use this:

```python
def make_transport(sock, *args, **kwargs) -> Transport:
    t = Transport(sock, *args, **kwargs)
    o = t.get_security_options()
    o.ciphers = ('aes256-cbc',)  # only allow aes256-cbc
    return t

client = SSHClient()
client.connect("example.com", transport_factory=make_transport)

```

Given this, it's probably a good idea to deprecate the args to `SSHClient.connect` that just get passed to `Transport`, as they make this a bit more awkward to deal with (ie. we wouldn't need `*args` and `**kwargs`).

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -237,6 +237,7 @@ class SSHClient(ClosingContextManager):
         gss_trust_dns=True,
         passphrase=None,
         disabled_algorithms=None,
+        transport_factory=Transport,
     ):
         """
         Connect to an SSH server and authenticate to it.  The server's host key
@@ -314,6 +315,10 @@ class SSHClient(ClosingContextManager):
         :param dict disabled_algorithms:
             an optional dict passed directly to `.Transport` and its keyword
             argument of the same name.
+        :param transport_factory: an optional callable that takes in a new `socket`
+            `gss_kex`, `gss_deleg_creds`, `disabled_algorithms` and generates a
+            `.Transport` instance to be used by this client. Defaults to
+            `.Transport.__init__`.
 
         :raises:
             `.BadHostKeyException` -- if the server's host key could not be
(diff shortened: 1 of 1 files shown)


bskinn:
Thanks for the PR, @chiefnoah! I've flagged it for bitprophet to review in terms of the approach and if/how it fits his vision for things.

Unfortunately, I can't provide a timeline for a review on this, as bitprophet is neck-deep in dropping Python 2 and updating to modern Sphinx for his various projects. We *will* get there; I just can't say when.

chiefnoah:
The approach is definitely a divergence from the normal Paramiko API conventions, however it buys you quite a bit of functionality without a full rewrite or breaking change. `SSHClient` already leaks a fair amount of `Transport`'s API in `connect`, but just inverting the dependency isn't really easy without duplicating or moving all of the `socket` logic somewhere-else:tm:. Unfortunately, having pretty explicit control over the `Transport` connection paramters is necessary when dealing with ~~crappy~~ opinionated router SSH implementations, so I'm sure I'm not the only one who's found themselves facing a partial rewrite of `SSHClient.connect` just to get access to that.

bitprophet:
To be fair, "normal Paramiko API conventions" are not what I would personally have set up if I'd written it myself, nor something one wants to emulate w/o good reason 😂  plus I use factories in a few spots in Invoke/Fabric land.

In this case I agree it's a good, cheap way to hand more control to callers in lieu of more useful API rewrites (which will still come, someday). Thanks!

bitprophet:
2.12 out now!

PR #1250, opened 2018-07-05
made invoke_shell to utilize environment variable(s)

fahadnaeemkhan:
invoke_shell is currently ignoring the environment dict that is passed by caller.

PR #2129, opened 2022-10-20
add missing update env vars call on client.invoke_shell

chenb67:


jun66j5:
The same has been requested: #1250, #1820

Issue #512, opened 2015-04-12
Unexpected SSHException: Error reading SSH protocol banner(10038, 'An operation was attempted on something that is not a socket)


toniatop:
I am trying to set up a ssh connection through paramiko, using a proxy. 

Considering the following code:

    import paramiko
    ssh = paramiko.SSHClient()
    ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
    sock = paramiko.ProxyCommand('ssh user@proxy_host nc host 22')
    ssh.connect(host, username='my_username',sock=sock) 
    stdin,stdout,stderr = ssh.exec_command("ls .") 
    ssh.close()
Under cygwin it works just fine. 
On the same machine, under Windows 8.1 it gives:

    No handlers could be found for logger "paramiko.transport"
    Exception in thread Thread-1:
    Traceback (most recent call last):
      File "C:\Users\user\Anaconda\lib\threading.py", line 810, in __bootstrap_inner
        self.run()
      File "build\bdist.win-amd64\egg\paramiko\transport.py", line 1673, in run
        self.sock.close()
      File "build\bdist.win-amd64\egg\paramiko\proxy.py", line 110, in close
        os.kill(self.process.pid, signal.SIGTERM)
    WindowsError: [Error 5] Access is denied
    
    Traceback (most recent call last):
      File ".\test_python_ssh.py", line 14, in <module>
        ssh.connect(host, username='my_username',sock=sock) 
      File "build\bdist.win-amd64\egg\paramiko\client.py", line 324, in connect
      File "build\bdist.win-amd64\egg\paramiko\transport.py", line 406, in start_client
    paramiko.ssh_exception.SSHException: Error reading SSH protocol banner(10038, 'An operation was attempted on something t
    hat is not a socket')

Any ideas how to solve the situation?

jrvilda:
I have a similar problem.

Environment:
```
>ver
Microsoft Windows [Versión 6.1.7601]
>python
ActivePython 2.6.7.20 (ActiveState Software Inc.) based on
Python 2.6.7 (r267:88850, Jun 27 2011, 13:56:33) [MSC v.1500 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> print paramiko.__version__
1.15.2
```

Execution failed :

```
>>> import paramiko
>>> ssh = paramiko.SSHClient()
>>> ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
[...]

PR #2130, opened 2022-10-25
Use WaitForSingleObject() for ProxyCommand instead of select() on Windows

jun66j5:
Proposed fix for #512.

`select()` with non-socket objects doesn't work on Windows, but `WaitForSingleObject()` is able to wait for stdout of ProxyCommand.

### DIFF
--- a/paramiko/proxy.py
+++ b/paramiko/proxy.py
@@ -19,7 +19,6 @@
 
 import os
 import shlex
-import signal
 from select import select
 import socket
 import time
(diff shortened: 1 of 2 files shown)


trinacriax:
Hi @jun66j5, why did you close the PR? AFAIU it has not been merged.

jun66j5:
paramiko's author didn't seem interested in the PR, so I closed it. If you think it should be merged, you could submit new PR using changes in the PR.

bskinn:
It's not that he was specifically not interested, @jun66j5 -- the project has been specifically on a bugfix/feature hold (except for critical problems) for quite some time, throughout the process of dropping Python 2 & old 3.x and of the current key/auth rework.

(Combined with the fact that paramiko is only one of several projects he maintains.)

Especially given @trinacriax's interest, it seems to me worth keeping this open. It'll eventually get looked at; I just don't know how long it will be before that happens.

jun66j5:
My interests in the PR and fixing it on Windows went away.

trinacriax:
Hi Jun, 
I had the same issue on Windows and this fix worked to me.
I understand you are not interested anymore in this PR merge, however, closing the PR is not worth for the project and for Windows users having this issue.
Indeed, I kindly ask authors, including @bitprophet and @bskinn, to review this PR towards either the merge or the reject with comments. 
Jun, thanks again for the fix.

bskinn:
@trinacriax I'm only a triager, I can do some basic review but all the evaluation is ultimately up to bitprophet.

I suspect jun66j5 doesn't want to reopen this PR because then they will have to be engaged throughout the review process, even though they don't care about it any more.

If you have specific need for it, @trinacriax, I would do as jun66j5 suggested: take the patch from this PR and create a new PR with the same changes; we can iterate forward from there.

trinacriax:
Hi @bskinn, thanks for the information.
Honestly, I hope @jun66j5 will reopen the PR, following the process, and taking the deserved credit for the fix proposed.

TechArtistG:
> @trinacriax I'm only a triager, I can do some basic review but all the evaluation is ultimately up to bitprophet.
> 
> I suspect jun66j5 doesn't want to reopen this PR because then they will have to be engaged throughout the review process, even though they don't care about it any more.
> 
> If you have specific need for it, @trinacriax, I would do as jun66j5 suggested: take the patch from this PR and create a new PR with the same changes; we can iterate forward from there.

I know it's a bit since this post but +1 to resurrect this PR.

samuel-deal-tisseo:
I'm also a bit late, but I'm too really interrested for this PR.
I monkey patched my paramiko (3.5.1, since it's the version current pyinfra uses) using this PR, and it fixes my issues.

This PR is better than the PR #1924 which hangs at the end of the connection

PR #2134, opened 2022-10-31
Fix exception handling when another exception already occurred.

nightwatch92:
SSH script will exit with the following error "During handling of the above exception, another exception occurred", when we try to ssh from host A into a host B that has record of A in hosts.deny.

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -2271,6 +2271,8 @@ class Transport(threading.Thread, ClosingContextManager):
                 buf = self.packetizer.readline(timeout)
             except ProxyCommandFailure:
                 raise
+            except ConnectionResetError:
+                raise
             except Exception as e:
                 raise SSHException(
                     "Error reading SSH protocol banner" + str(e)


bskinn:
So we know more clearly what this is meant to fix, can you explain more about what the problem was?

If I read correctly what you wrote, it sounds like in this situation (where B has A in `hosts.deny`), the error message coming from paramiko when trying to connect from A is unclear: that is, the exception that's currently raised in this situation provides unhelpful/incomplete/misleading information ('error reading protocol banner', instead of 'connection reset'), making it harder for the user to understand what went wrong?

And, adding `ConnectionResetError` as a special case to be raised as-is, *does* pass this helpful information up to the user?

Thanks!

nightwatch92:
I think the information is misleading, but the bigger problem for me is that the try block fails and I can't catch the exception.
Let say we have the following code and try to execute it from A which is as mentioned is in B's("hosts.deny.server.com") hosts.deny.

```python
import paramiko

def ssh(host):
    conn = paramiko.client.SSHClient()
    conn.set_missing_host_key_policy(paramiko.client.AutoAddPolicy)
    try:
        conn.connect(hostname=host, username='root', port=22, timeout=10, banner_timeout=20, key_filename="/path-to-my-key")
    except Exception as e:
        print(f"could not ssh to {host} {e}")

    return conn
ssh("hosts.deny.server.com")
```
After executing the code, we are going to have the following error: 

```shell
Exception (client): Error reading SSH protocol banner[Errno 104] Connection reset by peer
  File "/home/user/tests/.env/lib/python3.10/site-packages/paramiko/transport.py", line 2271, in _check_banner
    buf = self.packetizer.readline(timeout)
  File "/home/user/tests/.env/lib/python3.10/site-packages/paramiko/packet.py", line 380, in readline
    buf += self._read_timeout(timeout)
  File "/home/user/tests/.env/lib/python3.10/site-packages/paramiko/packet.py", line 607, in _read_timeout
    x = self.__socket.recv(128)
ConnectionResetError: [Errno 104] Connection reset by peer

During handling of the above exception, another exception occurred:

  File "/home/user/tests/.env/lib/python3.10/site-packages/paramiko/transport.py", line 2094, in run
    self._check_banner()
  File "/home/user/tests/.env/lib/python3.10/site-packages/paramiko/transport.py", line 2279, in _check_banner
    raise SSHException(
paramiko.ssh_exception.SSHException: Error reading SSH protocol banner[Errno 104] Connection reset by peer

could not ssh to hosts.deny.server.com Error reading SSH protocol banner[Errno 104] Connection reset by peer

```

The main problem is that we cannot catch the exception properly.

PR #2137, opened 2022-11-09
add support to also read inlcuded config files from ssh

mortbauer:


jun66j5:
The same improvement #1892 has been requested, with unit tests.

PR #2150, opened 2022-11-21
Use unittest.mock for Python => 3.3

jelly:
Since Python 3.3 unittest.mock has been included in Python itself and it's no longer required to use an external module.


bskinn:
[This is already coming][1] as part of bitprophet's work to drop Python 2 and older 3.x. Thanks for the work to create the PR , though!


[1]: https://github.com/paramiko/paramiko/pull/2037#issuecomment-1146351932

PR #2151, opened 2022-11-25
add collections module compatibility for python3.10

chymdyugah:


jlisicki96-vf:
Any news on that? It has been a month.

This PR already takes backwards compatibility into account.

bskinn:
I've raised these to bitprophet's attention, and he's got them bookmarked. He's working on dropping Python 2 and older 3.x for all of his packages; once it's paramiko's turn for that surgery, he'll be wrapping this PR (and others) into that work.

Thanks for checking in!

alex:
`is_callable` doesn't seem to actually be used, as far as I can tell. Can it just be deleted?

bskinn:
> `is_callable` doesn't seem to actually be used, as far as I can tell. Can it just be deleted?

Indeed. Maybe it can!

bitprophet:
Yea, weird! I think maybe the original py3 compat contributor was overzealous there, I don't even see evidence it ever _was_ used.

Also, I think we should just nuke the entire py3compat file now that we're dropping 2 entirely. I'm working on this now so I think we can just close this out. Thanks!

PR #2158, opened 2022-12-15
Fix FSF address in LICENSE file

mtelka:
... to match https://www.gnu.org/licenses/old-licenses/lgpl-2.1.txt.


bskinn:
Hm, thanks for this.

Is this just something that was missed in https://github.com/paramiko/paramiko/pull/2006?

mtelka:
> Is this just something that was missed in #2006?
Yes: https://web.archive.org/web/20070816041004/https://www.gnu.org/licenses/old-licenses/lgpl-2.1.html

PR #2110, opened 2022-08-26
Speed up SFTP transfer speed

wyn1995:
Speed up SFTP transfer speed by remove unnecessary `__repr__` function calls.

My test code:
```python
from time import time
from io import BytesIO

from paramiko.client import AutoAddPolicy, SSHClient

def main() -> None:
    # create fileobj to transfer
    file_obj = BytesIO()
    file_obj.seek(4 * 1024 * 1024 * 1024 - 1)
    file_obj.write(b"1")
    file_obj.seek(0)

    ssh_client = SSHClient()
    ssh_client.set_missing_host_key_policy(AutoAddPolicy())
    ssh_client.connect(hostname="<hostname>", username="<username>", password="<password>", compress=True)
    
    sftp_client = ssh_client.open_sftp()

    start_time = time()
    sftp_client.putfo(file_obj, "/dev/shm/chunk.bin")

    print(f"{time() - start_time}s")


if __name__ == "__main__":
    main()
```

before: 
cost 111.59725689888s

after:
cost 42.47094106674194s

PR #2165, opened 2023-01-10
Improve performance by reducing expensive bytes conversion

alex:
In two core parts of the codebase, complex type-switch based code is used to convert a value to bytes. However, in all cases (except for one, fixed in this PR), the caller is specifying a Message instance. We can make this code much simpler by directly calling the correct method on Message.

In #2110 this is measured to massively speed up large SFTP transfers.


alex:
Failing test appears to be the same as on the 3.0 branch. Unfortunately its all sftp tests that are failing, so a bit hard to be confident that this doesn't break anything.

bitprophet:
It's in! Thanks Alex!

PR #2166, opened 2023-01-10
Drop Python 2!

bitprophet:
This will become version 3.0. Very minor API changes intended; mostly stuff that should be low impact.

Remaining:

- [x] fix SFTP tests, which are broken for unknown reasons. Hopefully simple/dumb localized mistake somewhere.
- [x] rest of my global py2 drop checklist (internal), basically metadata/packaging tweaks, massaging lots of legacy code styles, and making sure any remaining deprecation warnings are handled.


bitprophet:
Tests (besides the release test, which is never happy on new-major-release branches, it's fine) green now! 🎉 

Back to the rest of the regular dropping-py2 bullshit instead of the "removing py3compat.py" bullshit.

bitprophet:
The failing check is just the release test, which is only failing because it doesn't handle major revisions well yet (sigh). Otherwise I think this is good to go.

setup.py
```diff
@@ -75,28 +76,19 @@
         "Topic :: Internet",
         "Topic :: Security :: Cryptography",
         "Programming Language :: Python",
-        "Programming Language :: Python :: 2",
-        "Programming Language :: Python :: 2.7",
         "Programming Language :: Python :: 3",
```
hugovk, line 79: Optionally:
```suggestion
        "Programming Language :: Python :: 3 :: Only",
```

hugovk (commented):
Good stuff! I only skimmed through, you've got the important `python_requires` there. 👍 

---

There's a "(2.7, 3.4+)" to update on the README.

---

This could be simplified in `demos/forward.py`:

```python
try:
    import SocketServer
except ImportError:
    import socketserver as SocketServer
```

* https://docs.python.org/2.7/library/socketserver.html
* https://docs.python.org/3.6/library/socketserver.html

---

https://github.com/asottile/pyupgrade is a tool you could also run to upgrade syntax for 3.6+:

```sh
python -m pip install pyupgrade
pyupgrade **/*.py --py36-plus
```

Although it could wait for a follow-up.

bitprophet:
Thanks for the check @hugovk !

- readme: great spot! I need to update the regexen I use to search for mentions of old python. Found another 1-2 minor bits just now too.
- demos: I want to just delete that folder entirely in the next big release (people should be using Fabric for trivial use cases instead of copy/paste/modify'ing the demos), so I'll opt to ignore for now - thanks tho.
- I think I have pyupgrade bookmarked, but yea, agree it can wait - this took long enough that I'm in no rush to delay it further :D

bskinn:
For a lighter touch than pyupgrade, you could pass the codebase through [flynt](https://pypi.org/project/flynt/)? I haven't used it myself, but the Python Bytes guys seem to like it a lot.

sites/www/changelog.rst
```diff
@@ -2,6 +2,84 @@
...
+    This change is backwards incompatible. Such references should be
+    search-and-replaced with their modern Python 3.6+ equivalents; in some
+    cases, still-useful methods or values have been moved to `paramiko.util`
+    (most) or `paramiko.common` (``byte_*``).
+
+- :support:`-` Drop support for Python versions less than 3.6, including Python
+  2. So long and thanks for all the fish!
+
+  .. warning::
+    This change is backwards compatible. However, our packaging metadata has
```
bskinn, line 74: Assuming this is a typo?

```suggestion
    This change is backwards incompatible. However, our packaging metadata has
```

paramiko/channel.py
```diff
@@ -851,10 +849,10 @@ def sendall_stderr(self, s):
         """
         Send data to the channel's "stderr" stream, without allowing partial
         results.  Unlike `send_stderr`, this method continues to send data
-        from the given string until all data has been sent or an error occurs.
-        Nothing is returned.
+        from the given bytestring until all data has been sent or an error
+        occurs. Nothing is returned.
 
-        :param str s: data to send to the client as "stderr" output.
+        :param bytes s: data to send to the client as "stderr" output.
```
bskinn, line 855: This type was changed to `bytes` here -- was the analogous type on line 831 in the docstring for `sendall()` also supposed to be changed to `bytes`? (I can't tag a comment on a line that isn't diffed, annoyingly.)

paramiko/file.py
```diff
@@ -93,39 +94,20 @@ def flush(self):
...
-            :raises: ``StopIteration`` -- when the end of the file is reached.
+        :raises: ``StopIteration`` -- when the end of the file is reached.
 
-            :returns: a line (`str`) read from the file.
-            """
-            line = self.readline()
-            if not line:
-                raise StopIteration
-            return line
+        :returns: a line (`str`) read from the file.
```
bskinn, line 105: `str` or `bytes`?

tests/test_sftp.py
```diff
@@ -28,14 +28,15 @@
...
 from tempfile import mkstemp
 
 import pytest
 
-from paramiko.py3compat import PY2, b, u, StringIO
 from paramiko.common import o777, o600, o666, o644
-from tests import requireNonAsciiLocale
 from paramiko.sftp_attr import SFTPAttributes
+from paramiko.util import b, u
+from tests import requireNonAsciiLocale
```
bskinn, line 39: Nit - Why the initial lowercase 'r' on this otherwise camelcase?

tests/test_file.py
```diff
@@ -198,7 +188,7 @@ def test_write_bad_type(self):
             self.assertRaises(TypeError, f.write, object())
 
     def test_write_unicode_as_binary(self):
-        text = u"\xa7 why is writing text to a binary file allowed?\n"
+        text = "\xa7 why is writing text to a binary file allowed?\n"
```
bskinn, line 191: What... this *works?*

bskinn (commented):
I scanned about 2/3 of the codebase in the GH PR diff viewer, and did a handful of full-codebase searches for some possible badnesses. Couple things to look at.

sites/www/changelog.rst
bitprophet: I think at one point it _wasn't_ a typo (eg I was saying "it's backwards compatible unless you're on old python") but then the rest of the paragraph changed, so yea, now it is. Thx!

paramiko/channel.py
bitprophet: Yea, that one was missed for some reason, good spot!

paramiko/file.py
bitprophet: Either-or actually (see `SFTPFile.readline`), I'll enhance this to make that clearer.

tests/test_sftp.py
bitprophet: Whatever random contributor added that did it that way, apparently. This isn't worth changing now, I'd rather look harder at the underlying helper when we do a second cleanup pass, and either nix it entirely or rename it.

tests/test_file.py
bitprophet: Which? the thing being tested or simply using `\x` inside a Unicode literal?

If the latter, it's totes [a thing](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals), though I'm unclear on why we're doing it in this particular test - maybe to ensure the unicode string is itself a little "funny looking" / not 100% printable.

If the former, yea, no idea off the top of my head, but see line 381 of `paramiko.file`, looks like it's some sort of compatibility shim behavior.

bskinn: The ability to pass a `str` to `f.write()` on something opened as binary (`rb+`).

Unless there's something special about `LoopbackFile` that's different from conventional `open()`?

bskinn: I guess that's maybe why you said "see line 381 of ..", that probably explains it. I didn't go look yet.

PR #2169, opened 2023-01-14
Check that the requested user matches the user from krb5 token

sarfarazahmad89:
So I was trying to write a custom SSH server off of paramiko, which is when we realized that paramiko just simply trusts the __username__ whenever gssapi auth is used. The idea behind this PR is to compare the provided username with the initiator-name in the client provided token. Though I am not a 100% sure if this is a general enough fix. It just splices on '@' and takes the first element in the array.  I am well aware of the disclaimer given in the Server documentation, https://docs.paramiko.org/en/stable/api/server.html#paramiko.server.ServerInterface.check_auth_gssapi_with_mic
 but I was wondering if paramiko should (or could even , _in the general sense_) incorporate this check?

### DIFF
--- a/paramiko/ssh_gss.py
+++ b/paramiko/ssh_gss.py
@@ -123,6 +123,7 @@ class _SSH_GSSAuth(object):
         self._gss_deleg_creds = gss_deleg_creds
         self._gss_host = None
         self._username = None
+        self._username_from_krb5_token = None
         self._session_id = None
         self._service = "ssh-connection"
         """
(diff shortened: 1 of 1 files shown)


sarfarazahmad89:
@bskinn  anything on this one? I realize I didn't do a very good with the write up there. Paramiko at the moment don't really care if the `initiator_name` in the kerberos ticket matches the user requested with the -l flag. So if I write a SSH server in paramiko that does gssapi/krb5 authentication then with an active ticket as say `userA`, I can do ssh -l userB -o PreferredAuthentication=gssapi-with-mic somessshserver -l userB. 

The patch does just that (it forces username with -l to match the initiator_name in the kerberos ticket). I can extend it to the older gssapi package too, if you believe this is the right thing to do.

bskinn:
Sorry for the slow action on this, it's been a busy month here.

On first glance, this seems like a really important check to make. I doubt it should be _mandatory_ -- an escape hatch would be smart, for exceptional circumstances: `--dangerously-ignore-krb-user-mismatch` or similar.

Ultimately bitprophet is the one who will vet this for inclusion. I'll mark it for his attention.

This may make sense to incorporate as part of the broader key/auth re-work of #387. Will see.

Thanks for the PR and the nudge!

PR #2173, opened 2023-01-23
Split known_hosts file entries on a single tab or space

alexdotc:
The parser currently splits on only a single space but should probably be more permissive.

The OpenSSH manpage (in "SSH_KNOWN_HOSTS FILE FORMAT") states: "...The fields are separated by spaces" which seems like 1 or more for each separator but maybe I'm misreading the phrasing/plurality. Either way, the implementation code itself will accept multiple spaces and tab characters when reading a known_hosts file in practice as well.


alexdotc:
Actually on further inspection it seems like it will take either a single space or a single tab but nothing else. I'll update this shortly to account for that instead

alexdotc:
(Updated)

bskinn:
Sorry for the delay!

This looks pretty clean, thanks for the PR!

Could you check in the test suite to see if there's a good place to add a test case for this?

Also, this might be worth a mention in the [docs](https://docs.paramiko.org/en/stable/api/config.html).

Finally, please add a brief [CHANGELOG entry](https://github.com/paramiko/paramiko/blob/main/sites/www/changelog.rst)

alexdotc:
Thanks, all set on all 3 counts

bskinn:
Flagged for @bitprophet review -- thanks!

Issue #1715, opened 2020-07-07
ProxyCommand leaves open pipes


aayla-secura:
`ProxyCommand.close` does not close `stdin`, `stdout` and `stderr` which leaves 3 open pipes after each connection. Eventually this results in `OSError: Too many files open`. I suggest the following fix (taking into account the suggested fix for issue #1713):

```
diff -U0 -r a/proxy.py b/proxy.py
--- a/proxy.py	2020-07-08 09:13:04.000000000 +1200
+++ b/proxy.py	2020-07-08 09:58:24.000000000 +1200
@@ -122 +122,5 @@
-        os.kill(self.process.pid, signal.SIGTERM)
+        self.process.stdin.close()
+        self.process.stdout.close()
+        self.process.stderr.close()
+        self.process.terminate()
+        self.process.wait()
```

NOTE: With this fix there is now an exception thrown due to a race condition between `Transport.run` and `Transport.stop_thread`:

```
Exception: ProxyCommand("ssh -W 1.2.3.4:22 proxy") returned nonzero exit status: Bad file descriptor
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/paramiko/proxy.py", line 107, in recv
    r, w, x = select([self.process.stdout], [], [], select_timeout)
OSError: [Errno 9] Bad file descriptor

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/paramiko/transport.py", line 2055, in run
    ptype, m = self.packetizer.read_message()
  File "/usr/local/lib/python3.7/site-packages/paramiko/packet.py", line 459, in read_message
    header = self.read_all(self.__block_size_in, check_rekey=True)
  File "/usr/local/lib/python3.7/site-packages/paramiko/packet.py", line 301, in read_all
    x = self.__socket.recv(n)
  File "/usr/local/lib/python3.7/site-packages/paramiko/proxy.py", line 119, in recv
    raise ProxyCommandFailure(" ".join(self.cmd), e.strerror)
paramiko.ssh_exception.ProxyCommandFailure: ProxyCommand("ssh -W 1.2.3.4:22 proxy") returned nonzero exit status: Bad file descriptor
```

This is because the descriptors will be closed while the other thread is still in the main loop in `Transport.stop_thread`, before the next time it checks `self.active`. Perhaps you could just add another `except` block for `OSError`?

===
[...]

Issue #1713, opened 2020-07-04
ProxyCommand leaves a zombie process


aayla-secura:
`ProxyCommand.close` does not wait for the pid, so the process is left in a zombie state. I suggest the following fix:

```
--- a/proxy.py	2020-07-05 08:26:14.000000000 +1200
+++ b/proxy.py	2020-07-05 08:26:11.000000000 +1200
@@ -112 +112,5 @@
-        os.kill(self.process.pid, signal.SIGTERM)
+        try:
+            os.kill(self.process.pid, signal.SIGTERM)
+            os.waitpid(self.process.pid, 0)
+        except ProcessLookupError:
+            pass
```

Catching `ProcessLookupError` is necessary when `close` is called multiple times.

Here's a test for it:

```
import logging
from time import sleep
import sys
import os
from contextlib import contextmanager
from psutil import Process, NoSuchProcess
from paramiko import SSHClient, ProxyCommand
from paramiko.client import MissingHostKeyPolicy

logger = logging.getLogger('paramiko test')
handler = logging.StreamHandler(sys.stdout)
handler.setFormatter(logging.Formatter(
    fmt=('[%(asctime)s] %(name)s '
         '(%(threadName)s): %(message)s'),
    datefmt='%d/%b/%Y %H:%M:%S'))
logger.addHandler(handler)
logger.setLevel(logging.INFO)


SSH_HOST = '1.2.3.4'                       # CHANGE THIS
SSH_USER = 'user'                           # CHANGE THIS
PROXY = 'ssh -W 1.2.3.4:22 proxy'  # CHANGE THIS


class IgnoreMissingHostKeyPolicy(MissingHostKeyPolicy):
    def missing_host_key(self, client, hostname, key):
        pass

def _proxy_status(proxy, pref):
    try:
        proc = Process(proxy.process.pid)
    except NoSuchProcess:
        logger.info('{}: proxy process killed, good'.format(pref))
    else:
        logger.info('{}: proxy process {} is {}'.format(
            pref, proc.pid, proc.status()))

def _exec(conn, proxy, cmd='whoami'):
    stdin, stdout, stderr = conn.exec_command(cmd)
    logger.info('{} -> {}'.format(
        cmd, stdout.read().decode('utf-8').rstrip('\r\n')))
    sleep(0.5)
    _proxy_status(proxy, 'after exec')

@contextmanager
def connect():
    proxy = ProxyCommand(PROXY)
    _proxy_status(proxy, 'after create')
[...]

PR #2174, opened 2023-01-24
ProxyCommand leaves open pipes (Issue #1715)

thomas-rotszyld:
Followup issue #1715 and issue #1713 
ProxyCommand.close does not close stdin, stdout and stderr which leaves 3 open pipes after each connection. Eventually this results in OSError: Too many files open. I suggest the following fix (taking into account the suggested fix for issue https://github.com/paramiko/paramiko/issues/1713)

Suggested Fixes tested and working (over several thousand of devices).

**Before Fix:**
filedescriptor out of range in select() error message after 200-300 SSH connections through a proxy SSH.

After Running Python script (get running config on device)
Filedescriptor Exception raised after 200/300 Devices, Scripts exited with error but Open Pipes remain opened): 

lsof | grep "python" | grep "pipe" | awk '{print $2}' | wc -l shows a very high number of open pipe connexion:

```
root@b3894971d232:/opt# lsof | grep "python" | grep "pipe" | awk '{print $2}' | wc -l
66921
```

Relaunching the same script raises immediately the Fildescriptor Exception (Open Pipes consumes too many Open Files reaching ulimit for bad reason)

**After Fix:**
Same Script runs without Filedescription Exception.
After exiting the script, the lsof comman does normal open pipes number.

```
root@b3894971d232:/opt# lsof | grep "python" | grep "pipe" | awk '{print $2}' | wc -l 
5429
```

### DIFF
--- a/paramiko/proxy.py
+++ b/paramiko/proxy.py
@@ -106,9 +106,7 @@ class ProxyCommand(ClosingContextManager):
 
                 r, w, x = select([self.process.stdout], [], [], select_timeout)
                 if r and r[0] == self.process.stdout:
-                    buffer += os.read(
-                        self.process.stdout.fileno(), size - len(buffer)
-                    )
+                    buffer += os.read(self.process.stdout.fileno(), size - len(buffer))
             return buffer
         except socket.timeout:
             if buffer:
@@ -119,7 +117,11 @@ class ProxyCommand(ClosingContextManager):
             raise ProxyCommandFailure(" ".join(self.cmd), e.strerror)
 
     def close(self):
-        os.kill(self.process.pid, signal.SIGTERM)
+        self.process.stdin.close()
+        self.process.stdout.close()
+        self.process.stderr.close()
+        self.process.terminate()
+        self.process.wait()
 
     @property
     def closed(self):


tests/test_proxy.py
```diff
@@ -94,13 +93,6 @@ def test_recv_raises_ProxyCommandFailure_on_non_timeout_error(
         assert info.value.command == "hi"
         assert info.value.error == "whoops"
 
-    @patch("paramiko.proxy.subprocess.Popen")
-    @patch("paramiko.proxy.os.kill")
-    def test_close_kills_subprocess(self, os_kill, Popen):
```
bskinn, line 99: Is there another test that you could replace this test with, instead of only deleting it?

paramiko/proxy.py
```diff
@@ -119,7 +118,11 @@ def recv(self, size):
             raise ProxyCommandFailure(" ".join(self.cmd), e.strerror)
 
     def close(self):
-        os.kill(self.process.pid, signal.SIGTERM)
+        self.process.stdin.close()
+        self.process.stdout.close()
+        self.process.stderr.close()
+        self.process.terminate()
+        self.process.wait()
```
bskinn, line 125: Is this `.wait()` a good idea? Does `os.kill()` wait for the process to close? I can't tell from the [docs](https://docs.python.org/3/library/os.html#os.kill).

I'm concerned this could lead to hangs if for some reason a process fails to terminate cleanly.

bskinn:
Thanks for the PR, @thomas-rotszyld

milan2655:
i have a similar setup and without the suggested fix I was getting "OSError: Too many files open".

wit the fix everything is working perfectly but I'm getting following error:

ERROR:paramiko.transport:Unknown exception: I/O operation on closed file
ERROR:paramiko.transport:Traceback (most recent call last):
ERROR:paramiko.transport:  File ".venv/lib/python3.10/site-packages/paramiko/transport.py", line 2109, in run
ERROR:paramiko.transport:    ptype, m = self.packetizer.read_message()
ERROR:paramiko.transport:  File ".venv/lib/python3.10/site-packages/paramiko/packet.py", line 453, in read_message
ERROR:paramiko.transport:    header = self.read_all(self.__block_size_in, check_rekey=True)
ERROR:paramiko.transport:  File ".venv/lib/python3.10/site-packages/paramiko/packet.py", line 301, in read_all
ERROR:paramiko.transport:    x = self.__socket.recv(n)
ERROR:paramiko.transport:  File ".venv/lib/python3.10/site-packages/paramiko/proxy.py", line 111, in recv
ERROR:paramiko.transport:    buffer += os.read(self.process.stdout.fileno(), size - len(buffer))
ERROR:paramiko.transport:ValueError: I/O operation on closed file
ERROR:paramiko.transport:

FloLaco:
@milan2655 which python library do you use on top of paramiko ? (netmiko, nornir, etc...) or only paramiko ?

milan2655:
i use nornir with netmiko_send_command task!

Issue #1134, opened 2017-12-17
No way to add a host key from a string


alexreg:
It seems there is currently no way to add a host key from a string, but rather only a file. See <http://docs.paramiko.org/en/2.4/api/hostkeys.html>.

I would suggest either exposing the `HostKeys.entries` attribute so that `HostKeyEntry`s can be added directly, or adding some proxy methods on the `HostKeys` class for adding/removing/updating `HostKeyEntry`s.

Furthermore, it would be nice if the `HostKey` class could load from a file-like object.

alexreg:
One more thing: it would be nice if the `HostKeyEntry` constructor could take a `key_type` parameter... or is this unnecessary?

ploxiln:
You can add a private key object:
http://docs.paramiko.org/en/2.4/api/hostkeys.html#paramiko.hostkeys.HostKeys.add

You can make a private key object like `paramiko.RSAKey(data=b"key byte string")`

alexreg:
Ah right. Slightly non-intuitive, but fair enough. I had just figured out the `paramiko.RSAKey(data=b"key byte string")` bit myself, but that's not so obvious, so perhaps could go in the docs. I used a base64 decode too, to convert from the standard OpenSSH string format.

benjimin:
I think it makes sense to allow a key as a string, without requiring the caller to know which supported crypto algorithm it uses.

For example, the application code may be loading the key (directly into memory) from some secure secret store and not wish to persist it onto a filesystem (as even a temp file increases exposure to credential leak), and if the key is updated in future to use a different supported crypto algorithm (say if the credential migrates from RSA to Ed25519 during a routine key rotation) the same application code should still work without modification.

Ideally we should be able to write:

```python
paramiko.SSHClient().connect(hostname="example.com", username="jane", key=my_private_key_string)
```

Instead of needing to write:

```python
for alg in [paramiko.RSAKey, paramiko.ECDSAKey]:
  try:
    key = alg.from_private_key(io.StringIO(my_private_key_string))
  except:
    continue
  else:
    break
paramiko.SSHClient().connect(hostname="example.com", username="jane", pkey=key)
```

PR #2181, opened 2023-02-06
Allow addition of HostKeyEntrys to HostKeys

daemcnally:
To address #1134.

This also allows users to add their own file parsing behaviour. For example if the user wishes to ignore lines that raise `InvalidHostKeyError`s .

### DIFF
--- a/paramiko/hostkeys.py
+++ b/paramiko/hostkeys.py
@@ -72,6 +72,21 @@ class HostKeys(MutableMapping):
                 e.key = key
                 return
         self._entries.append(HostKeyEntry([hostname], key))
+        
+    def add_host_key_entry(self, hostkeyentry):
+        """
+        Add a host key entry to the table from a host key entry.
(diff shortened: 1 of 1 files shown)


bskinn:
Interesting, so this is refactoring existing behavior to expose it for direct usage?

bskinn:
Also, as this is key-related, flagging for possible relevance to the overall key/auth overhaul of #387

daemcnally:
> Interesting, so this is refactoring existing behavior to expose it for direct usage?

Yes that's right, the purpose I wanted to use it for was reading a host key file but ignoring any keys that were invalid rather than raising an exception. This meant that I needed to read the file line by line, and it was annoying having a method to generate a host key entry but not to be able to add it in to the map.

destroyer5656:
I have been having an issue where I regenerate host keys every boot during a test and paramiko faults because it can't handle having a different hostkey with the same hostname. Could I use this function to ignore that check?

PR #2189, opened 2023-02-22
Enable FIPS compatibility for python 3.9+

davidesba:
Starting from python 3.9, a keyword-only argument "usedforsecurity" is introduced in hashlib hashing algorithms.
In a FIPS environment MD5 algorithm can not be used in a security context.
This change sets this not security context for MD5 fingerprint.


jun66j5:
The same pull #1928 has been requested.

bskinn:
Indeed, @jun66j5, but #1928 takes a different approach. Both might be valuable to consider.

jvrsantacruz:
Changes in this PR do not change functionality and it should be harmless to merge. 

Could we have the fix merged before evaluating the other related PR (#1928) that might bring many changes?

bskinn:
> Changes in this PR do not change functionality and it should be harmless to merge.

Perhaps so, but that's not the only consideration.

Are there any non-FIPS circumstances where it might be desirable for the `usedforsecurity` argument to be passed as `True`?

If so, then at least some discussion seems worthwhile as to whether `usedforsecurity` should also be exposed in the user-facing API.

jvrsantacruz:
> Are there any non-FIPS circumstances where it might be desirable for the usedforsecurity argument to be passed as True?

@bskinn luckily no, the usage of md5 in the code is to [calculate the fingerprint of the public key](https://github.com/paramiko/paramiko/blob/c598539e92baaf4c8cf5f01a117a1f0619065e9c/paramiko/pkey.py#L161) which is not a security issue as it is not hashing passwords or secrets but the public key and always that, so adding `usedforsecurity=False` is just a way of reflecting the current usage and it should be never `True` in any case.

bskinn:
> @bskinn luckily no, the usage of md5 in the code is to [calculate the fingerprint of the public key](https://github.com/paramiko/paramiko/blob/c598539e92baaf4c8cf5f01a117a1f0619065e9c/paramiko/pkey.py#L161) which is not a security issue as it is not hashing passwords or secrets but the public key and always that, so adding `usedforsecurity=False` is just a way of reflecting the current usage and it should be never `True` in any case.

Ah, ok -- makes sense.

It would be good to document this explicitly, either with a code comment or a remark in the docstring.

If there's a straightforward test that could be added, that'd be ideal too.

This also calls for a CHANGELOG entry, I think.

jvrsantacruz:
Hi! How are you? 

Can we help somehow to make this advance and merge it? 

We're deploying a fork of paramiko and it's not the best situation.

Thank you!

bskinn:
Flagging for consideration as part of the key/auth work on #387

scovetta:
I'd rather see the hash algorithm changed to some SHA-256 or SHA-512, since IMHO, hashing public keys *is* a security context, presumably because the caller would be making some decision based on the hash. So marking `usedforsecurity=False` would be misleading.

Issue #2090, opened 2022-07-22
disabled_algorithms getting ignored?


gusostow:
RSA authentication with paramiko as a client broke in versions >=2.9. I believe this is because paramiko is defaulting to sha2 algorithms instead of ssh-rsa as mentioned in the [changelog](https://www.paramiko.org/changelog.html). However, the recommended fix of using `disabled_algorithms` seems to be getting ignored by `Transport`:

Version: `2.11.0`

```
import io
import paramiko

creds = ...   # creds excluded

transport = paramiko.Transport(
    (creds["host"], 22),
    disabled_algorithms={'keys': ['rsa-sha2-256', 'rsa-sha2-512']}
)

private_key = paramiko.rsakey.RSAKey.from_private_key(
    io.StringIO(creds["private_key"])
)
transport.connect(username=creds["username"], password=None, pkey=private_key)
```

Produces
```
13:27:02.432 [DEBUG] starting thread (client mode): 0x81edec50
13:27:02.432 [DEBUG] Local version/idstring: SSH-2.0-paramiko_2.11.0
13:27:02.432 [DEBUG] Remote version/idstring: SSH-2.0-SSHD
13:27:02.432 [INFO ] Connected (version 2.0, client SSHD)
13:27:02.440 [DEBUG] === Key exchange possibilities ===
13:27:02.440 [DEBUG] kex algos: diffie-hellman-group1-sha1, diffie-hellman-group14-sha1, diffie-hellman-group-exchange-sha1, diffie-hellman-group-exchange-sha256
13:27:02.441 [DEBUG] server key: ssh-rsa
13:27:02.441 [DEBUG] client encrypt: aes128-ctr, aes256-cbc, aes192-cbc, aes128-cbc, aes256-ctr, aes192-ctr, blowfish-cbc, 3des-cbc
13:27:02.441 [DEBUG] server encrypt: aes128-ctr, aes256-cbc, aes192-cbc, aes128-cbc, aes256-ctr, aes192-ctr, blowfish-cbc, 3des-cbc
13:27:02.441 [DEBUG] client mac: hmac-sha1, hmac-md5, hmac-sha1-96, hmac-md5-96, hmac-sha256, hmac-sha256@ssh.com, hmac-sha2-256, hmac-sha2-512
[...]

PR #2200, opened 2023-03-08
changelog: fix disabled_algorithms suggestion for legacy servers

perrinjerome:
Fixes https://github.com/paramiko/paramiko/issues/2090

### DIFF
--- a/sites/www/changelog.rst
+++ b/sites/www/changelog.rst
@@ -246,7 +246,7 @@ Changelog
     required if your target systems do not support either RSA2 or the
     ``server-sig-algs`` protocol extension.
 
-    Specifically, you need to specify ``disabled_algorithms={'keys':
+    Specifically, you need to specify ``disabled_algorithms={'pubkeys':
     ['rsa-sha2-256', 'rsa-sha2-512']}`` in either `SSHClient
     <paramiko.client.SSHClient.__init__>` or `Transport
     <paramiko.transport.Transport.__init__>`. See below for details on why.


sites/www/changelog.rst
```diff
@@ -246,7 +246,7 @@ Changelog
     required if your target systems do not support either RSA2 or the
     ``server-sig-algs`` protocol extension.
 
-    Specifically, you need to specify ``disabled_algorithms={'keys':
+    Specifically, you need to specify ``disabled_algorithms={'pubkeys':
```
bskinn, line 249: Is `pubkeys` the correct key to use in `disabled_algorithms` in *all* cases? Or is the behavior more complex than that? (Honest question, I'm not deeply familiar.)

perrinjerome: It's always hard to tell about *all* cases so the short answer is "I am not sure", but many things make me believe this was a typo.

This changelog entry has a more detailed explanation below and here it is mentioned to use `pubkeys`:

https://github.com/paramiko/paramiko/blob/4bf5d30e0dd6eb7ebbecc03445c3b77f61ad761d/sites/www/changelog.rst?plain=1#L275-L276

Using `pubkeys` was the solution for  https://github.com/paramiko/paramiko/issues/2090#issuecomment-1192829965 and also for me when trying to connect to a "XFB.Gateway" server.

Also, this is documented as a workardound for servers not supporting the [`server-sig-algs`]( https://www.rfc-editor.org/rfc/rfc8308#section-3.1) extension and this is an extension for public key algorithms.

`keys` was added in https://github.com/paramiko/paramiko/commit/5f222495b5a62f3a1c465292bcace15888f40515 maybe @bitprophet can confirm ? Thanks !

bitprophet:
Thanks for posting! But nah, the original changelog should have been correct, the issue it was correcting for was _host keys_ related which, for legacy reasons, is what `keys` maps to here. `pubkeys` is specifically about user authentication, which as you found, is listed elsewhere in the changelog.

It's a little buried / requires the user to briefly check the code to see which attribute they're trying to mess with (and that's also where the confusion about keys vs pubkeys arises), but `disabled_algorithms`' format is doc'd here: https://docs.paramiko.org/en/stable/api/transport.html#paramiko.transport.Transport.__init__

This will all get _much_ better soon once I have a newer, cleaner auth system available for opt-in (being tracked under #387 for now).

Kache:
Then perhaps the docs should have a short list of the different keys users would need to address the ["slightly backwards incompatible" change](https://www.paramiko.org/changelog.html#2.9.0)?

Perhaps like:

 * `keys` - host keys
 * `pubkeys` - user auth
 * Others, plus link to source where `Keys should match the last word in the class’ builtin algorithm tuple attributes` is defined

PR #2201, opened 2023-03-09
Update client.py

DavidVaknin456:
I'm trying to connect to a remote server that does not support ipv6, this server return network error num 51(instead of ipv6 error),
 I remove the if that raises an error (if the error is not related to ipv6) in order to support more errors and continue to try connect with ipv4
 it work for me 
 thanks

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -367,10 +367,6 @@ class SSHClient(ClosingContextManager):
                     # to close sockets explicitly
                     if sock:
                         sock.close()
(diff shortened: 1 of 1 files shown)


bskinn:
Flagging for evaluation.

I doubt this is an acceptable solution, but there may well be some enhancement to the error handling here that will allow you to handle this situation gracefully.

Can you provide the code you're using, and a complete traceback of the error you get? Thanks!

Issue #2204, opened 2023-03-11
Issues with the parsing of known_hosts file


keachi:
I'll reopen part of #69 as the issue still exists.

If a known_hosts file has more than one whitespace or tab between hostname and key data paramiko fails to read that line.

ssh known_hosts file example:
```
www.example.com    ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDPcwVpzi/QGvgDJFFDirVKpnBdRZzpxS0v58gQNUxKvNNcrTDouCZWaMH5EUEFZ4XMAK0tfsRWOKPf59oT0JFm53bzKdvpu3FN6aKoIXL6n2pB85O/+ScBefpfMS4H/0GWVmhWYVeAUq3gfxVrEM7xDEZxnxl5ZQtPBG/XCVA9v8E4xuUvKYlPtz6w99QocNveE9khg55oPqROCGlDumgaAoH0x/NKncn/YIrx9I8MXUZcEDglXhGs0iQb7B4cGNuBoSmE5e1Jit2R38xxgh3AArZEc9tWlDj0s8eEmfMDt9nzigFwfsDEHZ/9ecUE5UqHRElxBQm5MIbjcQNEDbodG68LuKROCJh8mhmNSUTA17Yi3eYUwgDzmERn4zYDLFmepmAKRMyoqYKgMa35/V4TN5ByyFacaRz2pArESF8ScGTt8c2acqADPs1xqsPpQT3watBrV0gRL4xFc9Eq4qgnvLA6w8Q4etFZN8KhmPg9Ay2ikVfPLcpmN3h9hq8130s=
```

https://github.com/paramiko/paramiko/blob/ee2829903f0108b6ea6da59cf29bff9c3cff3a74/paramiko/hostkeys.py#L328

ptjm:
I just had this problem. To fix it, I think it should be sufficient to replace this:
```        fields = re.split(" |\t", line) ```
with this
```        fields = line.split()```
in the module cited above. Does that seem reasonable?

0cjs:
This would be closed by #2205. (The GitHub auto-close system did not recognise that "solves #2204" in that PR's head comment will close this; I think you need to say "closes #2204" there to make it work.)

PR #2205, opened 2023-03-11
Parse host keys with more than one space or tab

keachi:
This solves #2204.
known_hosts with more than one whitespace or tab are supported by OpenSSH. This should be the case of Paramiko as well.

### DIFF
--- a/paramiko/hostkeys.py
+++ b/paramiko/hostkeys.py
@@ -329,8 +329,9 @@ class HostKeyEntry:
         """
         Parses the given line of text to find the names for the host,
         the type of key, and the key data. The line is expected to be in the
-        format used by the OpenSSH known_hosts file. Fields are separated by a
-        single space or tab.
+        format used by the OpenSSH known_hosts file. Fields are separated by
+        spaces or tabs between host and key and a single space or tab between
+        the type of key and the key data.
 
         Lines are expected to not have leading or trailing whitespace.
         We don't bother to check for comments or empty lines.  All of
@@ -339,7 +340,7 @@ class HostKeyEntry:
         :param str line: a line from an OpenSSH known_hosts file
         """
         log = get_logger("paramiko.hostkeys")
-        fields = re.split(" |\t", line)
+        fields = re.split(r"[ \t]+", line)
         if len(fields) < 3:
             # Bad number of fields
             msg = "Not enough fields found in known_hosts in line {} ({!r})"
--- a/tests/test_hostkeys.py
+++ b/tests/test_hostkeys.py
@@ -36,6 +36,14 @@ broken.example.com ssh-rsa AAAA
 happy.example.com ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAIEA8bP1ZA7DCZDB9J0s50l31M\
 BGQ3GQ/Fc7SX6gkpXkwcZryoi4kNFhHu5LvHcZPdxXV1D+uTMfGS1eyd2Yz/DoNWXNAl8TI0cAsW\
 5ymME3bQ4J/k1IKxCtz/bAlAqFgKoc+EolMziDYqWIATtW0rYTJvzGAzTmMj80/QpsFH+Pc2M=
+mixed.example.com\t ssh-rsa\t AAAAB3NzaC1yc2EAAAADAQABAAABgQDwlJemtKlEJvcOMcj\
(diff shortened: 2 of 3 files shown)


paramiko/hostkeys.py
```diff
@@ -330,7 +330,8 @@ def from_line(cls, line, lineno=None):
         Parses the given line of text to find the names for the host,
         the type of key, and the key data. The line is expected to be in the
         format used by the OpenSSH known_hosts file. Fields are separated by a
-        single space or tab.
+        spaces or tabs between host and key and a single space or tab between
```
bskinn, line 333: Typo: Need to remove the ` a` at the end of the preceding line.

tests/test_hostkeys.py
```diff
@@ -33,7 +33,7 @@
 9q/kAYG6sP9f5zuUYP8i7FOFp/6ncCEbbtg/lB+A3iidyxoSWl+9jtoyyDOOVX4UIDV9G11Ml8om3\
 D+jrpI9cycZHqilK0HmxDeCuxbwyMuaCygU9gS2qoRvNLWZk70OpIKSSpBo0Wl3/XUmz9uhc=
 broken.example.com ssh-rsa AAAA
-happy.example.com ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAIEA8bP1ZA7DCZDB9J0s50l31M\
+happy.example.com  ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAIEA8bP1ZA7DCZDB9J0s50l31M\
```
bskinn, line 36: Instead of revising the existing test-case host entries, please create new ones to test these additional whitespace use-cases.

Per my other comment, if a mix of spaces and tabs is acceptable syntax, please add a test case for that, too, after revising the regex.

paramiko/hostkeys.py
```diff
@@ -339,7 +340,9 @@ def from_line(cls, line, lineno=None):
         :param str line: a line from an OpenSSH known_hosts file
         """
         log = get_logger("paramiko.hostkeys")
-        fields = re.split(" |\t", line)
+        fields = re.split(" +|\t+", line, 1)
```
bskinn, line 343: Can you explain why you've implemented this in this manner?

Specific questions:

- This first split (with `maxsplit=1`) will always produce a list of length two, so why bother with the `if len(fields) == 2`? Just split twice.

- Why are you splitting on a run of spaces or a run of tabs in the first split, but only a single space or tab in the second split? Does that match the OpenSSH behavior? Either way, is there any reason not to _also_ allow a run of spaces or tabs for the second whitespace separator?

- What about cases where spaces and tabs are intermixed in the whitespace? Is that valid for OpenSSH? If not, does it matter? Would it be good for paramiko to tolerate those cases also? If so, then this isn't the right regex to use.

My inclination would be to replace these three lines with something like the following (passing `maxsplit` as a keyword arg so that it's clear what that value means):

```
fields = re.split(r"[ \t]+", line, maxsplit=2)
```

But, there might be reasons why this is a bad idea. Quite open to being told why we shouldn't do this. 😀

bskinn (changes requested):
Thanks for the PR! 

Couple of requested changes, and one request for comment/explanation.

paramiko/hostkeys.py
keachi: Fixed

paramiko/hostkeys.py
keachi: The idea was to split with more than one whitespace / tab only the hostname and the key. But I actually tested it now with OpenSSH 9.0p1 and it's possible to split the keytype and base64-encoded key with more than one whitespace character.
Mixing spaces and tabs is supported too.

tests/test_hostkeys.py
keachi: I added a new key to test this behaviour.

bskinn:
Great, thanks for taking care of these, @keachi!

One thing left - this is worthy of a CHANGELOG entry -- please add one to `/sites/www/changelog.rst`, if you would.

keachi:
Fully agree. I added the changelog entry.

bskinn:
Flagged for bitprophet review. Thanks, @keachi!

0cjs:
This bug just wasted two hours of my day. What is preventing this quite simple THREE YEAR OLD fix from being merged?

PR #2211, opened 2023-03-17
Use assertEqual instead of assertEquals for Python 3.12 compatibility.

tirkarthi:
https://docs.python.org/3.12/whatsnew/3.12.html#removed . This method was removed in Python 3.12

### DIFF
--- a/tests/test_gssapi.py
+++ b/tests/test_gssapi.py
@@ -54,7 +54,7 @@ class GSSAPITest(KerberosTestCase):
 
         oid = encoder.encode(ObjectIdentifier(self.krb5_mech))
         mech, __ = decoder.decode(oid)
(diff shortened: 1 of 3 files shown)


bskinn:
Thanks for the PR! I think I remember seeing this.

Looks like you've done a pretty thorough job, but to be sure: Have you checked the entire test suite for this deprecated usage?

Is the new method available in Python 3.6+? (There are jobs in the CI that test on 3.6, so I'm guessing this is all good, but want to specifically ask.)

This is worth a CHANGELOG entry -- please add one, if you would.

bskinn:
It looks like @mgorny found two more instances that needed replacement in #2275. Closing this PR in favor of that one. Thank you for this PR, though, @tirkarthi!

PR #2216, opened 2023-03-26
Adding flexibility for weird SSH servers

Etrenak:
The purpose of this PR is to add flexibility through options for the client. I actually have a Nokia network device that needs these two new options to be compatible with Paramiko.

The `start_with_auth_none` option adds tolerance to SSH servers that consistently respond with an `SSH_MSG_USERAUTH_FAILURE` message on the first `SSH_MSG_USERAUTH_REQUEST` attempt.

The `resend_service_requests` option forces Paramiko not to resend service requests after it has already received an accept message from the server for that service. The RFCs are not precise on this point, and some SSH servers close the connection after receiving a second message for the same service.

These two options, if enabled, are very similar to the way the OpenSSH client works, which always sends a first authentication attempt `'none'` and does not send back a request to open a service if it has already been accepted.

### DIFF
--- a/paramiko/auth_handler.py
+++ b/paramiko/auth_handler.py
@@ -114,7 +114,7 @@ class AuthHandler:
             self.auth_event = event
             self.auth_method = "none"
             self.username = username
-            self._request_auth()
+            self._request_service_auth()
         finally:
             self.transport.lock.release()
 
@@ -125,7 +125,7 @@ class AuthHandler:
             self.auth_method = "publickey"
             self.username = username
             self.private_key = key
-            self._request_auth()
+            self._request_service_auth()
         finally:
             self.transport.lock.release()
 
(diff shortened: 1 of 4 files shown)


bskinn:
This definitely will be affected by the ongoing work of #387.

matthijskooijman:
> The resend_service_requests option forces Paramiko not to resend service requests after it has already received an accept message from the server for that service. The RFCs are not precise on this point, and some SSH servers close the connection after receiving a second message for the same service.

IIUC, this is also helpful for compatibility between a paramiko client and an AsyncSSH server. Since AsyncSSH 2.14.1, to fix an [authentication injection issue](https://github.com/ronf/asyncssh/security/advisories/GHSA-c35q-ffpf-5qpm), asyncssh [rejects duplicate `userauth-request` service requests](https://github.com/ronf/asyncssh/blob/c3dc8693600395c6a7ed44b60d2265ddfbb27079/asyncssh/connection.py#L2138-L2139). I've seen this cause connection failures in practice when multiple keys are available locally with paramiko 3.4.0 and AsyncSSH 2.14.1. As noted, OpenSSH client only sends one `userauth-request` and works without issues.

PR #2217, opened 2023-03-28
Support setting server_hostkey_name in connect()

MasterofJOKers:
This allows advanced users to define something like `HostKeyAlias` from SSH's config file by allowing to pass `server_hostkey_name` to `SSHClient.connect`.


bskinn:
Flagging for attention related to the key/auth overhaul of #387

tests/test_client.py
```diff
@@ -678,6 +678,36 @@ def test_host_key_negotiation_3(self):
...
+            **self.connect_kwargs,
+        )
+
+        # show that we can connect if we set the explicit name for hostkey
+        # lookup
+        _test(server_hostkey_name=hostname, **self.connect_kwargs)
+        self.event.wait(1.0)
+        self.assertTrue(self.event.is_set())
+        self.assertTrue(self.ts.is_active())
+        self.assertEqual(True, self.ts.is_authenticated())
```
bskinn, line 709: Is there a reason you didn't keep using `assertTrue()` here?

tests/test_client.py
```diff
@@ -678,6 +678,36 @@ def test_host_key_negotiation_3(self):
...
+            kfile = "ecdsa-256.key"
+            host_key = paramiko.ECDSAKey.from_private_key_file(_support(kfile))
+            known_hosts = self.tc.get_host_keys()
+            known_hosts.add(hostname, host_key.get_name(), host_key)
+
+            self.tc.connect(password="pygmalion", **connect_args)
+
+        # make sure there is no hostname for the default
+        self.assertRaises(
+            paramiko.ssh_exception.SSHException,
```
bskinn, line 698: Is there any (good) way to narrow the scope of this catch?

`SSHException` is a pretty large umbrella at the moment.

paramiko/client.py
```diff
@@ -335,6 +336,10 @@ def connect(
                 ``password``, ``key_filename`` and ``allow_agent``) and will
                 trigger an exception if given alongside them.
 
+        :param server_hostkey_name:
```
bskinn, line 339: Since this is a simple type, it seems to me we should put that type in the `option` here:

```suggestion
        :param str server_hostkey_name:
```

bskinn (commented):
Thanks for the PR!

One recommended minor change and two questions for you.

tests/test_client.py
MasterofJOKers: Good question. Since I hadn't done any development on paramiko, yet, I copied some other test's code and made it work for me. [That code](https://github.com/paramiko/paramiko/blob/main/tests/test_client.py#L659-L663) already does the `assertEqual()`.

I'll change it to `assertTrue()`.

tests/test_client.py
MasterofJOKers: I don't think so, but I don't know the code-base well enough to be sure.
This is the exception coming out of `_test(**self.connect_kwargs)`:
```
paramiko.ssh_exception.SSHException: Server '[127.0.0.1]:60469' not found in known_hosts
```

bskinn: That's pretty much what I figured. This should be fine, especially in a test.

Issue #2221, opened 2023-04-06
KEX - curve25519-sha256@libssh.org doesn't match curve25519-sha256


probabble:
curve25519-sha256@libssh.org is present in the available KEX in my paramiko client, and my server has curve25519-sha256 as its (only) set Kex option. When paramiko tries to negotiate a connection, it fails because those two strings are different. 

I believe that these are the same under the hood and paramiko could use the same implementation with a different key ("curve25519-sha256")

bskinn:
Flagging to include in the key/auth work of #387. Thanks for the suggestion!

probabble:
since i need this sooner than later, here's a PR 
https://github.com/paramiko/paramiko/pull/2222

probabble:
evidence that the two are indeed the same: https://www.rfc-editor.org/rfc/rfc8731

PR #2222, opened 2023-04-07
add alias for KEX "curve25519-sha256" == "curve25519-sha566@libssh.org"

probabble:
to support servers that only declare support for the former;


bskinn:
Flagging this here also, as relevant to #387.

probabble:
should resolve #2221

dharrigan-lenovo:
Having the same issue.  Any outlook on when this patch will be pushed out?

PR #2226, opened 2023-04-12
Remove codecov dependency

alex:
They deleted their package from PyPI, see https://about.codecov.io/blog/message-regarding-the-pypi-package/. I don't see any actual usage of the package.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -13,7 +13,6 @@ black>=22.8,<22.9
 codespell>=2.2,<2.3
 # Coverage!
 coverage>=6.2,<7
-codecov==2.1.12
 # Documentation tools
 alabaster==0.7.13
 releases>=2.1


alex:
Ok turns out it's used via an orb. The correct solution is to use codecov's bash uploader, but I have no idea how to ship that into this orb.

I'd appreciate some guidance. This is currently breaking pyca/cryptography's CI, so we need to find a way to fix ASAP or we'll have to disable testing paramiko.

bskinn:
I'll elevate this to @bitprophet, thanks for the report.

bitprophet:
Thanks a fucking lot codecov. jackasses. I'll figure this out shortly.

alex:
FWIW the right invocation is basically `curl -Os https://uploader.codecov.io/latest/linux/codecov` to download it, then `chmod +x` it and put it somewhere on the path (or just run it by absolute path).

bitprophet:
Well no, that's _a possible_ invocation, but the _right_ invocation looks like having my orb reference codecov's published CircleCI orb, which has a `codecov/upload` action to call. I'm experimenting with this locally now.

bitprophet:
Jeff's having a bad day :( 

image

if the current run goes ok, I'll publish a non-dev orb version. @alex how is Cryptography getting its Paramiko for CI purposes? Main branch, tag, pypi release, other? I'm committing the Paramiko-side changes for this in my 3.0 branch for now but might backport it if needed.

alex:
I know them feels. Thanks for jumping on this.

We just clone main to run our downstream tests, so as soon as something lands which removes it from dev-requirements, we should be good to go: https://github.com/pyca/cryptography/blob/main/.github/downstream.d/paramiko.sh

bitprophet:
Great, so if the 3.0 push I just did, passes, I will merge-up to main and you should hopefully be fixed.

alex:
Sounds great. I'm going to close this, since this PR is clearly wrong :-)

bitprophet:
Yup, sounds good, thanks.

Also, lol???

```
[2023-04-12T22:59:32.345Z] ['error'] There was an error running the uploader: Error uploading to https://codecov.io: Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Commit sha does not match Circle build. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
CircleCI received exit code 0
```

I'll dig into this and just let you know when it's fixed-fixed (but also I'll just merge what's already pushed, to main, so it gets you past your immediate blockage - since your tests presumably don't care if my own tests' coverage gets pushed...)

bitprophet:
Sounds like maybe Codecov is just kind of sad nowadays. The errors I'm seeing might just be due to rate limiting of their attempt to automagically treat public repos on the big CI platforms as "not needing a token". Going to put this on my back burner until I care (which isn't often...I only ever really use these coverage platforms, vs local coverage reports, when they work seamlessly.)

PR #2241, opened 2023-05-16
Middlehunterr patch 1

middlehunterr:


bskinn:
This appears to be a patch for a specific use case you have @middlehunterr -- I'm afraid we won't be able to accept it, as not everyone is going to need a "Y" sent at this point in the auth process.

Instead, if you could open an issue explaining what your situation is and why you need to send this 'Y' at this point in auth, we might be able to identify a general solution that could be implemented.

---

Separately, how are you trying to use paramiko? Depending on your use case, [Fabric](https://www.fabfile.org/) might have the features you need.

Issue #122, opened 2012-12-17
Connection with CISCO by private-key throwing 'Illegal info request from server'


fmiceli:
I am trying to send remote commands to a CISCO ISR-891.

I have tried doing so with username and password without issues. But when I try to do it using private key authentication I get the following error:
# 

Traceback (most recent call last):
  File "CA_srv.py", line 17, in <module>
    ssh.connect(hostname, port=2222, username='control', password=None, pkey=mkey, look_for_keys=False)
  File "/usr/local/lib/python2.7/dist-packages/paramiko-1.9.0-py2.7.egg/paramiko/client.py", line 338, in connect
    self._auth(username, password, pkey, key_filenames, allow_agent, look_for_keys)
  File "/usr/local/lib/python2.7/dist-packages/paramiko-1.9.0-py2.7.egg/paramiko/client.py", line 519, in _auth
    raise saved_exception
# paramiko.SSHException: Illegal info request from server
# The output of the logfile is the following:

....................
....................
DEB [20121217-13:35:50.025] thr=2   paramiko.transport: Trying SSH key 32324d2e86db5b638bcb87a44774d9e7
DEB [20121217-13:35:50.213] thr=1   paramiko.transport: userauth is OK
ERR [20121217-13:35:50.229] thr=1   paramiko.transport: Exception: Illegal info request from server
ERR [20121217-13:35:50.230] thr=1   paramiko.transport: Traceback (most recent call last):
ERR [20121217-13:35:50.231] thr=1   paramiko.transport:   File "/usr/local/lib/python2.7/dist-packages/paramiko-1.9.0-py2.7.egg/paramiko/transport.py", line 1598, in run
ERR [20121217-13:35:50.231] thr=1   paramiko.transport:     self.auth_handler._handler_table[ptype](self.auth_handler, m)
ERR [20121217-13:35:50.231] thr=1   paramiko.transport:   File "/usr/local/lib/python2.7/dist-packages/paramiko-1.9.0-py2.7.egg/paramiko/auth_handler.py", line 384, in _parse_userauth_info_request
ERR [20121217-13:35:50.231] thr=1   paramiko.transport:     raise SSHException('Illegal info request from server')
# ERR [20121217-13:35:50.231] thr=1   paramiko.transport: SSHException: Illegal info request from server

Here is my code:
# 

import paramiko

privatekeyfile = '/root/.ssh/id_rsa'
hostname = '200.125.52.2'

paramiko.util.log_to_file('/root/log.para')

[...]

PR #2242, opened 2023-05-22
bugfix: pubkey auth: resending signature

j2s2:
Some Cisco device doesn't respect the RFC for pubkey auth. They ignore signature in the first message and ask again for signature (cf. #122 ). The patch resend the first message with signature in this case.

### DIFF
--- a/paramiko/auth_handler.py
+++ b/paramiko/auth_handler.py
@@ -87,6 +87,7 @@ class AuthHandler:
         self.banner = None
         self.password = None
         self.private_key = None
+        self.msg_pubkey_with_sig = None
         self.interactive_handler = None
         self.submethods = None
         # for server mode:
@@ -394,6 +395,8 @@ class AuthHandler:
                 )
                 sig = self.private_key.sign_ssh_data(blob, algorithm)
                 m.add_string(sig)
+                # save message for possible future resending
+                self.msg_pubkey_with_sig = m
             elif self.auth_method == "keyboard-interactive":
                 m.add_string("")
                 m.add_string(self.submethods)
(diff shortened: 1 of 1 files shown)


bskinn:
This is referenced in #122, but flagging here also as a use-case with needs relevant to the key/auth overhaul of #387.

j2s2:
well ! Should i do something ?

bskinn:
Sorry for the delay on this.

It looks like you've implemented this off of paramiko v3.1. There is now experimental machinery for a new way of handling authentication [available in v3.2](https://www.paramiko.org/changelog.html#3.2.0) (scroll down until you see the first red callout box).

I don't think bitprophet will be interested in these sorts of fixes off of the 3.1 line. Either (i) the experimental auth handling tools in 3.2 already will allow you to implement this solution, (ii) future work on that overhauled auth machinery will enable it, or (iii) it's something that will get implemented once the core auth overhaul is done.

I would recommend testing out the new [`auth_strategy`](https://docs.paramiko.org/en/latest/api/auth.html#module-paramiko.auth_strategy) machinery in v3.2, to see if that API will let you do what you need to do in your client code, rather than making a change inside paramiko.

Please report back here whether this succeeds or fails -- both types of feedback will be very helpful!

j2s2:
thank for answer !
I have re-apply the patch to the main branch and it fits well.
Should I open a new pull-request ?

bskinn:
@j2s2 Hm, I think I may not have expressed myself well. 

I didn't mean for you to test your patch against `main`/v3.2.

I meant that you should try to use the new, experimental features already in paramiko v3.2 to see if those features will let you achieve this without further modifications to the paramiko codebase.

j2s2:
Indeed, I misunderstood.
Well ! I already spent lot of time to find where modifying the code as I wanted.
I feel I will spend much more time to understand how to use "auth_strategy" !

Fortunately, the current patch allows the ssh to work on our fleet of switch, at least until v3.2.

bskinn:
Yes, if this patch will work for you, then that's great!

To note, the patch should still work for paramiko v3.2, and probably for several future versions. bitprophet won't be removing this old machinery for a while.

Issue #2251, opened 2023-06-08
Quadratic time complexity of known_hosts file parsing


jack-stanek-hpe:
Hello,
I'm running into poor performance when loading a large `known_hosts` file with around 7,000 entries. For instance:
```
$ wc -l ./known_hosts
    7000 ./known_hosts
$ python3
Python 3.9.16 (main, Dec  7 2022, 10:15:43)
[Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import paramiko
>>> import timeit
>>> timeit.timeit(lambda: paramiko.hostkeys.HostKeys('./known_hosts'), number=1)
23.574272852999997
```
(I threw together a `known_hosts` generator script, as well as a synthetic `known_hosts` file [here](https://gist.github.com/jack-stanek-hpe/67c306ed2c2086b7a4dd22edb2d3468e) as a reproducer for reference.)

After digging into it a bit, I think this is due to the quadratic time complexity of `HostKeys` loading due to the way that key entries are stored in a flat list, and so constant-time random access isn't possible. This seems to be the same issue as observed in #437.

Going off the discussion in that issue, I decided to take a crack at addressing this, and I have a [branch on my own fork](https://github.com/paramiko/paramiko/compare/main...jack-stanek-hpe:paramiko:faster-hostkeys-loading) which re-implements parts of the `HostKeys` class to speed up lookups. My approach is to split out entries with hashed hostnames and plaintext hostnames, and place entries with plaintext hostnames into a dict for constant-time lookups. With hashed hostnames, I don't think constant time lookups are actually possible, since looking up a hashed hostname entry for some plaintext hostname string requires hashing the hostname with every hashed entry's salt, which is going to be linear time anyway. For the use-case with a large known_hosts file with only plaintext hostnames, this approach should nonetheless provide substantial performance benefits.

Please let me know if you'd like me to open a PR from my branch, I'd be happy to be able to contribute! I'm not sure what your process is for contributions, and I don't want to step on any toes. 😅

bskinn:
Thanks for the report! And, thanks for mentioning the contributing info -- looks like it's not mentioned in the README anywhere, but paramiko.org [refers](https://www.paramiko.org/contributing.html) to this [general contributions guide](https://www.contribution-guide.org/).

In any event, a PR is definitely welcome -- please do create one.

[...]

PR #2254, opened 2023-06-12
Speed up loading of HostKeys

jack-stanek-hpe:
Use a `dict` to handle non-hashed keys in the `known_hosts` file to speed up loading and access. Addresses #2251.

### DIFF
--- a/paramiko/hostkeys.py
+++ b/paramiko/hostkeys.py
@@ -15,17 +15,17 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with Paramiko; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA.
-
-
 from base64 import encodebytes, decodebytes
 import binascii
 import os
 import re
 
 from collections.abc import MutableMapping
+from collections import defaultdict
 from hashlib import sha1
 from hmac import HMAC
-
+from itertools import chain
+import math
 
 from paramiko.pkey import PKey, UnknownKeyType
 from paramiko.util import get_logger, constant_time_bytes_eq, b, u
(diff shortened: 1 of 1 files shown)


paramiko/hostkeys.py
```diff
@@ -172,17 +199,13 @@ def __setitem__(self, key, val):
...
-                return [
-                    e.key.get_name()
-                    for e in self._entries
-                    if e.key is not None
-                ]
+                return [e.key.get_name() for e in self._entries if e.key is not None]
 
         entries = []
-        for e in self._entries:
+        for e in chain(self._entries[hostname], self._hashed_entries):
```
bskinn, line 208:
```suggestion
        # We have to search for both hashed and unhashed entries
        for e in chain(self._entries[hostname], self._hashed_entries):
```

paramiko/hostkeys.py
```diff
@@ -117,8 +139,13 @@ def save(self, filename):
         .. versionadded:: 1.6.1
         """
         with open(filename, "w") as f:
-            for e in self._entries:
-                line = e.to_line()
+            all_entries = set(chain(*self._entries.values(), self._hashed_entries))
```
bskinn, line 142: Is there some risk of loss of information here? E.g., a collision between an otherwise identical hashed and unhashed value (this might not be possible); or, duplicate entries for a hostname (this might also not be possible)?

I *think* there's not any concern here, but my grip on the data structures is still a bit hazy so I want to raise the question.

bskinn (changes requested):
Handful of questions/change requests.

Also, this should be pretty easy to add/augment tests for.

And, this will _definitely_ need a CHANGELOG entry in `/sites/www/changelog.rst`.

Thanks for the PR!

bskinn:
Also, looks like the CI errors were due to formatting issues. Blacken and those should go away.

PR #2257, opened 2023-06-22
Suppress superfluous exception reporting using PEP409

mhei:
I observed the following behavior when connecting to an embedded device. The SSH server is started but random other tasks slow down the whole system. Then the current behavior is:

    $ python3
    >>> import paramiko
    >>> s = paramiko.SSHClient()
    >>> s.set_missing_host_key_policy(paramiko.AutoAddPolicy())
    >>> s.connect("192.168.8.53", port=9000, username="root", password="superhero", timeout=2, allow_agent=False, look_for_keys=False)
    >>> Exception: Error reading SSH protocol banner
      File "/usr/local/lib/python3.6/dist-packages/paramiko/transport.py", line 2044, in _check_banner
        buf = self.packetizer.readline(timeout)
      File "/usr/local/lib/python3.6/dist-packages/paramiko/packet.py", line 353, in readline
        buf += self._read_timeout(timeout)
      File "/usr/local/lib/python3.6/dist-packages/paramiko/packet.py", line 555, in _read_timeout
        raise socket.timeout()
    socket.timeout

    During handling of the above exception, another exception occurred:

      File "/usr/local/lib/python3.6/dist-packages/paramiko/transport.py", line 1893, in run
        self._check_banner()
      File "/usr/local/lib/python3.6/dist-packages/paramiko/transport.py", line 2049, in _check_banner
        'Error reading SSH protocol banner' + str(e)
    paramiko.ssh_exception.SSHException: Error reading SSH protocol banner

Note: the code above triggers also a "No existing session" exception which I filtered out here to focus on my point.

The expected exception is SSHException("Error reading SSH protocol banner"). However, catching it in the calling program still prints the above to the console. In my eyes, reporting the socket.timeout stuff is not wanted since it is entirely covered by the "catchall" except clause and the intension is to replace all such exceptions with the new SSHException.
For this PEP409 [1] is implemented since Python 3.3 [2], so let's use it here.

After this fix, the same procedure looks like:

    >>> s.connect("192.168.8.53", port=9000, username="root", password="superhero", timeout=2, allow_agent=False, look_for_keys=False)
    >>> Exception: Error reading SSH protocol banner
      File "/usr/local/lib/python3.6/dist-packages/paramiko/transport.py", line 1893, in run
        self._check_banner()
      File "/usr/local/lib/python3.6/dist-packages/paramiko/transport.py", line 2050, in _check_banner
        ) from None
    paramiko.ssh_exception.SSHException: Error reading SSH protocol banner

Now, the calling program can catch the exception without any printout to the console anymore.

[1] https://peps.python.org/pep-0409/
[2] https://docs.python.org/3/whatsnew/3.3.html

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -2295,7 +2295,7 @@ class Transport(threading.Thread, ClosingContextManager):
             except Exception as e:
                 raise SSHException(
                     "Error reading SSH protocol banner" + str(e)
-                )
+                ) from None
             if buf[:4] == "SSH-":
                 break
             self._log(DEBUG, "Banner: " + buf)


bskinn:
While the full exception context does make for a noisy REPL, suppressing the traceback context like this would make future `paramiko` bug reports that hit this exception pathway _much_ harder to diagnose.

I suspect bitprophet will not want to make this change, but I'll leave this open and flag it for his review so that he can weigh in directly.

Thanks for taking the time to submit the PR!

mhei:
If you don't want to suppress the traceback in this case, then honestly I don't understand the introduction of the SSHException at this point. Wouldn't it then be easier, to not handle exceptions at all here and pass them all to "upper layers"? My current understanding of the current code is: whatever happened in the underground, we don't really care but summarize it with our own SSHException class.

bskinn:
I should note, before discussing further, that I'm neither the original author of the package, nor the current maintainer. So, I can't speak either to the past logic that led to this implementation nor any current reasoning for keeping it this way (if there is any, beyond not wanting to change the exception handling behavior). But, I can guess as best I can. :-)

> If you don't want to suppress the traceback in this case, then honestly I don't understand the introduction of the SSHException at this point. Wouldn't it then be easier, to not handle exceptions at all here and pass them all to "upper layers"?

Great question. Regardless of the ultimate decision here, it's a point worth raising.

Speculating, the main reason I can think of is that it's a tradeoff intended to make `paramiko` users' lives easier, so that they only have to catch an `SSHException`, regardless of what went wrong internally.

If this was/is the goal, I definitely wouldn't have achieved it this way. Instead, I would implemented a richer exception hierarchy, with SSHException being a parent class to however big a subclass tree. Then you can catch the `SSHException` parent if you don't care about exactly what went wrong.

---

But -- when considering your proposed change, all I can say is that, from a maintainer perspective, the _last_ thing I would want to do is detach the `SSHException` from its cause by adding `from None`.

If anything, I would want to make it better connected _programmatically_, by removing the `+ str(e)` from the exception message and adding `from e`, instead of `from None`.

I'll flag this for bitprophet's attention. My guess is that any changes here would take place as part of a broader, major overhaul of the exception hierarchy within `paramiko`, though -- and that's not on the [roadmap](https://bitprophet.org/projects/#roadmap), so I would guess this won't change for a while.

mhei:
> removing the + str(e) from the exception message and adding from e, instead of from None.

Yes, I considered this variant also and I'd also be fine with this way. Do you think I should now already change this PR, or still wait for bitprophet's feedback?

bskinn:
I'd say leave the PR with whichever change you most prefer. This discussion will be here for him to review, and likely his preference will be something not exactly like anything we've talked about.

mhei:
Anything I can do here to get fhis fixed?

mhei:
I rebased this PR to latest main branch. Still hoping, that we can get this merged.

PR #2258, opened 2023-06-22
auth_handler: properly check that public_blob attr exists

avalou:


bskinn:
Thanks for the report and proposed fix, @avalou!

Indeed, this seems like a good check to add. I'll leave it to bitprophet to evaluate whether this is the right logic flow in the event of a nonexistent `public_blob` member, though.

If you would, please add an entry to `/sites/www/changelog.rst` to document this fix.

Also, for our reference, would it be possible to provide the entire traceback? Never know what might be useful.

baltimorestrings:
Hello, came here to submit this exact PR. 

The hasattr is definitely needed, but I've found that there is still an error in situations where an agent is being forwarded.

What it *should* be doing is cycling through all keys in the agent and attempting to use them to log into the device in question one by one. 

What I've found is that an error `Oops, unhandled type 3 ('unimplemented')` occurs instead of properly logging into the device. 

I've made a gist with company info removed that can reproduce the original error. It uses netmiko, but forwards directly to paramiko, so hopefully it's helpful. 

I believe the hasattr is absolutely needed, but is incomplete

avalou:
@bskinn Hello, sorry for the delay, I will do so this weekend.

avalou:
Here is the full error stack : 
```
Unknown exception: public_blob
  File "/usr/local/lib/python3.10/dist-packages/paramiko/transport.py", line 2185, in run
    handler(m)
  File "/usr/local/lib/python3.10/dist-packages/paramiko/auth_handler.py", line 394, in _parse_service_accept
    key_type, bits = self._get_key_type_and_bits(self.private_key)
  File "/usr/local/lib/python3.10/dist-packages/paramiko/auth_handler.py", line 218, in _get_key_type_and_bits
    if key.public_blob:
  File "/usr/local/lib/python3.10/dist-packages/paramiko/agent.py", line 476, in __getattr__
    raise AttributeError(name)
AttributeError: public_blob

No corresponding virtual host in hostname.smthg.com was found.
```

Issue #2262, opened 2023-06-27
SSHClient.connect should return self


Robtom5:
As `SSHClient` subclasses `ClosingContextManager` we have access to both __enter__ and __exit__. As exit ensures closing a connection it would be helpful if we could also provide a clean way of entering.

If `SSHClient.connect` was changed to return `self` we could use the following syntax
```python

client = SSHClient(...)
with client.connect(hostname, username)  as ssh:
    client.exec_command("foo")
```

jun66j5:
I don't think so. Why not use `with SSHClient() as ...` like the following? (`.close` does nothing if `.connect` is not invoked yet).

```python
with paramiko.SSHClient() as client:
    client.connect(...)
    client.exec_command(...)
```

Robtom5:
In your case you are creating a new instance of the SSH client and have to perform any additional configuration (e.g. setting missing host key policy) each time you enter the context manager.

The proposed change would allow you to perform any setup once and then use the same client multiple times. Also it helps pair the connect and close explicitly which I think helps readability

```python3

ssh = paramiko.SSHClient()
key = paramiko.RSAKey.from_private_key_file(....)
ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy)

with ssh.connect("Target 1",..., pkey=key"):
    ssh.exec_command(...)
    
with ssh.connect("Target 2",..., pkey=key"):
    ssh.exec_command(...)

```

The change would not prevent you from operating your way but I think would be cleaner for operating as above

jun66j5:
Your changes breaks the API. The [`SSHClient.connect`](https://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient.connect) returns `AuthResult` or `None`.

See:
* https://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient.connect
* https://github.com/paramiko/paramiko/blob/3.2.0/paramiko/client.py#L338-L340
* https://github.com/paramiko/paramiko/blob/3.2.0/paramiko/client.py#L473-L475

Robtom5:
Ah, I had missed that when I forked and ran tests. I think I'll just make a wrapper locally for now as I think a whole new method is extreme

bskinn:
Thanks for submitting the idea though, @Robtom5 & for using paramiko. Good luck!  I'm going to close this issue for now.

PR #2263, opened 2023-06-27
Added return to SSHClient.connect

Robtom5:
Quick fix for issue #2262


Robtom5:
Discussion on #2262 highlights that this would change the documented behaviour of returning None if not returning an AuthResult

PR #2269, opened 2023-07-13
Fixed handling of "match" keyword

unisudo:
- do not update hosts where there is no "host" key, as in the "match" keyword in some cases
- make match type keywords case-insensitive


unisudo:
Closed because this [pull request](https://github.com/paramiko/paramiko/pull/1991) is better

PR #2270, opened 2023-07-20
Add markers to known_hosts parser re paramiko#771

MajorDallas:
Adds basic parser support for a marker in the first field. Additional handling of cert-authority should be discussed before being implemented.

Adds sample keys to test_hostkeys.py.

### DIFF
--- a/paramiko/hostkeys.py
+++ b/paramiko/hostkeys.py
@@ -18,6 +18,7 @@
 
 
 from base64 import encodebytes, decodebytes
+from enum import Enum
 import binascii
 import os
 import re
@@ -311,6 +312,11 @@ class InvalidHostKey(Exception):
         self.args = (line, exc)
 
 
+class HostKeyMarkers(str, Enum):
+    CERT_AUTHORITY = "@cert-authority"
+    REVOKED = "@revoked"
+
+
 class HostKeyEntry:
     """
     Representation of a line in an OpenSSH-style "known hosts" file.
@@ -342,7 +348,17 @@ class HostKeyEntry:
             msg = "Not enough fields found in known_hosts in line {} ({!r})"
             log.info(msg.format(lineno, line))
             return None
-        fields = fields[:3]
+
+        try:
+            marker = HostKeyMarkers(fields[0])
+            fields = fields[1:4]
+        except ValueError:
+            marker = None
+            fields = fields[:3]
+
+        if marker is HostKeyMarkers.REVOKED:
+            log.info("Key is marked as revoked")
+            return None
 
         names, key_type, key = fields
         names = names.split(",")
--- a/tests/test_hostkeys.py
+++ b/tests/test_hostkeys.py
@@ -41,6 +41,14 @@ modern.example.com ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIKHEChAIxsh2hr8Q\
 curvy.example.com ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlz\
 dHAyNTYAAABBBAa+pY7djSpbg5viAcZhPt56AO3U3Sd7h7dnlUp0EjfDgyYHYQxl2QZ4JGgfwR5iv9\
 T9iRZjQzvJd5s+kBAZtpk=
+@cert-authority ca.example.com ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAIEA8bP1ZA7DCZD\
+B9J0s50l31MBGQ3GQ/Fc7SX6gkpXkwcZryoi4kNFhHu5LvHcZPdxXV1D+uTMfGS1eyd2Yz/DoNWXNA\
+l8TI0cAsW5ymME3bQ4J/k1IKxCtz/bAlAqFgKoc+EolMziDYqWIATtW0rYTJvzGAzTmMj80/QpsFH+\
+Pc2M=
+@revoked revoked.example.com ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAIEA8bP1ZA7DCZD\
+B9J0s50l31MBGQ3GQ/Fc7SX6gkpXkwcZryoi4kNFhHu5LvHcZPdxXV1D+uTMfGS1eyd2Yz/DoNWXNA\
+l8TI0cAsW5ymME3bQ4J/k1IKxCtz/bAlAqFgKoc+EolMziDYqWIATtW0rYTJvzGAzTmMj80/QpsFH+\
+Pc2M=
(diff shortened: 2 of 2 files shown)


bskinn:
Applied `Needs patch` since more discussion on how to handle various bits is needed, and implementation changes will be required to suit.

bskinn:
@MajorDallas, since we know this is not ready for merge yet, please convert to a draft PR.

MajorDallas:
I hadn't seen #2251  / https://github.com/paramiko/paramiko/pull/2254 before now. I haven't looked closely at the code, though I did see the new `lineno` param from that PR in `main`, so maybe it's already implemented even though the PR is still open? If not, we would want to make sure this new parsing logic isn't causing a regression in time complexity.

bskinn:
100%, @MajorDallas, good catch.

bskinn:
Type annotations will be up to bitprophet. I don't believe there are currently plans to type the codebase, though.

(Now that Py2 is dropped, it might be less of a headache than in the past, though.)

bitprophet:
Thanks for this! Yes, the overall "host keys parser doesn't understand `@` annotations" thing is a missing feature, tho I gather this ticket is mostly just drawing a box around the exceptions from this problem, and isn't implementing actual CA checking support. I will try to merge this asap.

----

Also, for both the submitter and @bskinn: I don't mind patches adding mypy type annotations; it'll be awhile before it's worth explicitly annotating the entire codebase (we just did Invoke, a somewhat smaller and more modern codebase, a while ago and _that_ was even a chore) _but_ since new type hints are backwards compatible with "not caring about type hints"...it's fine! and means less work in the future.

bitprophet:
Right, specific feedback was requested too...

- I am +1 on storing the markers in HostKeyEntry, it's a custom object type so adding some new attribute or w/e for it shouldn't cause any consternation to existing code, AFAIK. Maybe name it/them with a leading underscore for now just to hedge our bets - the entire hostkey class set is old and creaky and due a replacement sometime.
- I think a host key lookup matching an explicitly revoked key is probably OK to be handled by returning None (as if there was a lookup-miss) but I'm curious what OpenSSH does in this case (that? or an error?) since we normally just default to what they do.

MajorDallas:
> Thanks for this! Yes, the overall "host keys parser doesn't understand `@` annotations" thing is a missing feature, tho I gather this ticket is mostly just drawing a box around the exceptions from this problem, and isn't implementing actual CA checking support. I'd be okay merging this at that level of implementation unless I missed an intent to extend it with the CA-signature support too (?).

Correct: my goal was merely to successfully parse `known_hosts`. For one thing, I'm not knowledgeable enough in cryptography to implement CA verification, and for another I don't have the cycles right now to learn it. Plus, it's removed enough from the task of parsing `known_hosts` that it deserves its own ticket.

MajorDallas:
One other small thing: I'm also looking at adding type annotations for `HostKeyEntry.from_line`. The `lineno` param defaults to `None`, so with no semantic changes the proper type should be `Optional[int]`. However, that method currently has exactly one caller within the library (if Pyright is to be believed), and the callsite\* basically guarantees that `lineno` will always be an `int` by virtue of being assigned by `enumerate`. Within the function, `lineno` is only used for the log message for a key with < 3 fields.

Assuming this method isn't supposed to be part of the public API, I'd recommend removing the default to `None`. 

\* There may be an off-by-one error happening here, too, but it's pretty minor:

```python
for lineno, line in enumerate(f, 1):
    ...
    try:
        entry = HostKeyEntry.from_line(line, lineno)
    ...
```

`enumerate` starts from 0, but line numbers in files start from 1 (if I'm not mistaken). The worst that happens is that an invalid hostkey entry gets logged as being on line `x` when it's actually on line `x-1`.

bskinn:
> Assuming this method isn't supposed to be part of the public API ...

Even if it's _not_ supposed to be part of the public API, there's a nonzero chance that this function has use in the wild, and changing this behavior would be considered a regression.

Given that this function could be used as an independent test of proper known_hosts line formatting, in which case the `lineno` is irrelevant, IMO this makes the nonzero chance even more nonzero.

> ... I'd recommend removing the default to `None`.

Indeed, as well as tightening the type to `int`, instead of `Optional[int]`, if we were to go this direction.

However, as long as a `None` case is gracefully handled inside the function, I think revising the type to `Optional[int]` is the better course.

MajorDallas:
The last couple of months have been _crazy_ busy with work, but I'm finally ready to take another look at this.

Two questions:

1. The main thing I'm not sure about for finishing the implementation is if there needs to be any explicit call to  `PKey.load_certificate` after `HostKeyEntry.from_line` in `HostKeyEntry.load`. I don't _think_ so: all `load` does is populate a list on the instance, and I assume that list then gets processed appropriately elsewhere..? Still trying to read backwards, but any pointers are appreciated.
2. Confession: I've made most of my git mistakes on small repos with few contributors, and always with merge rather than rebase. Since I've been away a while, `master` has pulled ahead of my fork. I've sync'd that branch, but now should I rebase `771-hostkeys` again onto the tip of `master` or just leave it where it is? (I figure this is one of those times I'll look less-bad for asking the "dumb" question than guessing and doing it wrong 🙂 )

mrmeszaros:
**!IMPORTANT** I am not the maintainer of this repository, but based on what I saw in the git history:

#### Short answer:
You should rebase.

#### Longer answer:
At least You should try - if You do not run into merge conflicts then it should be all good.
If You run into some, maybe abort, make a backup branch (in case the rebase breaks it beyond all hope).
Then you can try to resolve, and then check if your unit-tests run correctly.

#### Longest answer:
@MajorDallas I would be glad to help - as I see this is the currently worked / active PR.

PS.: I have an open PR I would like to get reviewed, so I hope the sooner this one is resolved, the sooner mine might :)

MajorDallas:
So I attempted a rebase in a back-up copy and it worked just fine... but there are only two commits in this branch, so I'm not sure it even makes a difference? I know (basically) how rebase is different from merge, but not sure why, in this case, it should be one or the other.

mrmeszaros:
Cool, good to know it was not a big deal :)

As far as I know, most projects prefer to have a linear history, because it is much easier to the eye and to argue about.

While one merge commit is not an issue, in the long run merge commits can introduce a lot of parallel branches with a git history that is hard to see through.

Lastly, a merge commit is usually dud - usually it does not convey much information, if not looked after these empty commits can take a big portion of the history.

I hope I was able to provide help with my experiences :)

mmb-davidsmith:
Are there any updates on getting this merged? It would be helpful to be able to handle `known_hosts` files containing markers as right now you either have to maintain a parallel host keys file or accept all keys if you have a cert authority in your `known_hosts`.

Issue #2271, opened 2023-07-21
[BUG] - forward.py demo: pipe write() after close() [Race Condition]


pettitpeon:
### Are you using paramiko as a client or server?

Client

### What feature(s) aren't working right?

SSH

### What version(s) of paramiko are you using?

3.2.0

### What version(s) of Python are you using?

3.10

### What operating system and version are you using?

ubuntu 20.4

### Expected/desired behavior

Channel should not be closed if there is pending data to be handled by the ssh client
https://github.com/paramiko/paramiko/blob/56f3c1625f32aa59b15a573a34ccc89134352edf/demos/forward.py#L98

### Actual behavior

There is a race condition between
https://github.com/paramiko/paramiko/blob/56f3c1625f32aa59b15a573a34ccc89134352edf/demos/forward.py#L98  
and  
https://github.com/paramiko/paramiko/blob/56f3c1625f32aa59b15a573a34ccc89134352edf/paramiko/transport.py#L2164  

If the channel is closed before the SSH client can handle the last data, we get **Bad file descriptor (9)** and the tunnel crashes.

### How to reproduce

Create a tunnel and send small packets on al loop. The tunnel will crash eventually

### Anything else?

https://github.com/paramiko/paramiko/blob/56f3c1625f32aa59b15a573a34ccc89134352edf/demos/forward.py#L98  
calls the _feed() handler  
https://github.com/paramiko/paramiko/blob/56f3c1625f32aa59b15a573a34ccc89134352edf/paramiko/channel.py#L1039  
which is not lock protected. So a race condition is possible.

PR #2274, opened 2023-07-23
Fix Channel.close() race condition in pipe.py

pettitpeon:
`self._closed = True`  
was moved to the top. Built-in assignment is python is atomic and there is no need to protect it with a lock.

If there is a `Bad file descriptor` (errno == 9). We can safely disregard it if `._closed` is set

### DIFF
--- a/paramiko/pipe.py
+++ b/paramiko/pipe.py
@@ -46,10 +46,9 @@ class PosixPipe:
         self._closed = False
 
     def close(self):
+        self._closed = True
         os.close(self._rfd)
         os.close(self._wfd)
-        # used for unit tests:
-        self._closed = True
 
     def fileno(self):
         return self._rfd
@@ -64,7 +63,13 @@ class PosixPipe:
         if self._set or self._closed:
             return
         self._set = True
-        os.write(self._wfd, b"*")
+        try:
+            os.write(self._wfd, b"*")
+        except OSError as e:
+            if e.errno == 9 and self._closed:
+                # The pipe was closed, no need to do anythin
+                return
+            raise e
 
     def set_forever(self):
         self._forever = True


pettitpeon:
Fixes this BUG
https://github.com/paramiko/paramiko/issues/2271

paramiko/pipe.py
```diff
@@ -64,7 +63,13 @@ def set(self):
...
             return
         self._set = True
-        os.write(self._wfd, b"*")
+        try:
+            os.write(self._wfd, b"*")
+        except OSError as e:
+            if e.errno == 9 and self._closed:
+                # The pipe was closed, no need to do anything
+                return
+            raise e
```
bskinn, line 72: Do we not also need something analogous to protect the `os.read(self._rfd, 1)` call above in `clear()`?

(We might not; again, just raising the question.)

paramiko/pipe.py
```diff
@@ -46,10 +46,9 @@ def __init__(self):
         self._closed = False
 
     def close(self):
+        self._closed = True
         os.close(self._rfd)
         os.close(self._wfd)
```
bskinn, line 51: @pettitpeon, is there any chance of either of these two `.close()` calls failing, making `self._closed == True` a possibly inaccurate representation of the pipe state?

(I would think that a failure of either of these two calls would signal an error state for the pipe such that `self._closed == True` is a _better_ state than `== False`... but I thought I'd raise the question.)

paramiko/pipe.py
```diff
@@ -64,7 +63,13 @@ def set(self):
         if self._set or self._closed:
             return
         self._set = True
-        os.write(self._wfd, b"*")
+        try:
+            os.write(self._wfd, b"*")
+        except OSError as e:
+            if e.errno == 9 and self._closed:
```
bskinn, line 69: Is it possible for this `if`condition to ever be `True`?

We've already trapped for `self._closed == True` with the first `if` in the method, so it seems like we'd only ever hit this point with `self._closed == False`.

bskinn (commented):
Couple of questions about the implementation. Thanks for the PR!

bskinn:
I would think it would be challenging to do, but: is there any good way to write a test for this? 

I would figure that any such test would require setting up a test harness with a server to connect to, and then run the open/close cycle you describe in the issue for some period of time long enough to trigger the race condition 90%+ of the time... but that would make for a flaky test, and that'd be not great.

Interested in any thoughts you have on it, though.

bskinn:
Oh, also -- I added `Needs changelog/docs` because this would need an entry in `/sites/www/changelog.rst`.

paramiko/pipe.py
pettitpeon: > is there any chance of either of these two .close() calls failing?

Yes, it could happen. If either `.close()` fails, it would raise an `OSError`. In that case some FD might stay open, but the Pipe is not usable anyways. There is no use-case of using it after `.close()`. Worst case scenario, we leak resources until the application exits. The most probable scenario is that the close() exception is not handled and your program terminates anyways.  
**A small improvement** would be to have single `try`s on each `.close()`. In that case if, the first fails and leaks, we might still close the second one correctly.  
In any case, I see `.close()` errors as non-handable, we close() file descriptors as a "best-effort" to release the resource, but cannot really do anything if it fails.

>  making `self._closed == True` a possibly inaccurate representation of the pipe state?

Not really, after `close()` has been called, the object is disposed and no one should it anyways. The "broken state" is meaningless.

Last, the `pipe` is most probably a child of `Channel` and when the channel gets destroyed it tries to re-close, which  fails if the pipe was previously closed, but it is disregarded by a catch-all try. This is OK, and aligns with the idea that it is a non-handable error. https://github.com/paramiko/paramiko/blob/f4d7fecc1e636248db0adc40f136c1a77220b9da/paramiko/channel.py#L136

paramiko/pipe.py
pettitpeon: From the `pipe`'s perspective, yes. And it would be wise to protect it. However, the way the pipe is used, this does not happen. The same thread
1. Opens,
2. clears and
3. closes
the pipe. So there is no race condition between `.clear()` and `.close().
A second thread sets it.

pettitpeon:
> I would think it would be challenging to do, but: is there any good way to write a test for this?

I will look into this. It might be possible by tunneling/forwarding an ssh connection and sending ssh echos in a loop.

paramiko/pipe.py
bskinn: Aha, this is the crux of the race condition, got it.

For the benefit of future readers of the code, please add a comment here describing the race condition as you just described it.

Otherwise we risk someone thinking the same thing I did and removing this.

paramiko/pipe.py
bskinn: Same request, please add a comment here explaining why a similar race catch is not (currently) required here.

pettitpeon:
> I would think it would be challenging to do, but: is there any good way to write a test for this?
> 
> I would figure that any such test would require setting up a test harness with a server to connect to, and then run the open/close cycle you describe in the issue for some period of time long enough to trigger the race condition 90%+ of the time... but that would make for a flaky test, and that'd be not great.
> 
> Interested in any thoughts you have on it, though.

Unfortunately, I could not think of a way of easily reproducing the bug. I tried tunneling SSH connections through a local forward (demo/forward.py), but SSH is too smart and rejects connections from the server side if it is being spammed (DoS attack mitigation). So I cannot open and close connections fast enough to run into the race condition.

paramiko/pipe.py
pettitpeon: done

paramiko/pipe.py
pettitpeon: done

paramiko/pipe.py
```diff
@@ -46,25 +46,50 @@ def __init__(self):
...
+
+        # This try fixes a race condition with .close()
+        # 1. The write/server thread sees ._closed == False and continues
+        # 2. The close/client thread closes the descriptors before the
+        #    write/server thread writes.
+        # 3. The write/server fails to write() because the FD has been closed
+        try:
+            os.write(self._wfd, b"*")
+        except OSError as e:
+            if e.errno == 9 and self._closed:
```
jun66j5, line 89: I'd suggest using `errno.EBADF` with `errno` module in `e.errno == 9`.

pettitpeon: Done

pettitpeon:
@bskinn @jun66j5 What else shall be done to merge? I fixed all the issues

paramiko/pipe.py
```diff
@@ -46,25 +47,50 @@ def __init__(self):
...
-        os.close(self._rfd)
-        os.close(self._wfd)
-        # used for unit tests:
         self._closed = True
 
+        # We cannot do anything about closing errors. It is only a
+        # "best effort" approach
+        try:
+            os.close(self._rfd)
+        except:
```
bskinn, line 56: Best practice is to catch `Exception` and subclasses, or else you get `BaseException` and subclasses, which would also trap things like `KeyboardInterrupt`.

```suggestion
        except Exception:
```

paramiko/pipe.py
```diff
@@ -46,25 +47,50 @@ def __init__(self):
...
+            pass
+        try:
+            os.close(self._wfd)
+        except:
```
bskinn, line 60: Same:

```suggestion
        except Exception:
```

bskinn (changes requested):

bskinn:
> @bskinn @jun66j5 What else shall be done to merge? I fixed all the issues

Minor tweak to the PR, and then I think it's ready for review.

It still needs a CHANGELOG entry - you said you're not sure what to do about that... take a look at `/www/sites/changelog.rst` to see the syntax & template of entries added for other bugfixes, and adapt to provide a description of this bugfix.

pettitpeon:
> > @bskinn @jun66j5 What else shall be done to merge? I fixed all the issues
> 
> Minor tweak to the PR, and then I think it's ready to flag for review by bitprophet.
> 
> It still needs a CHANGELOG entry - you said you're not sure what to do about that... take a look at `/www/sites/changelog.rst` to see the syntax & template of entries added for other bugfixes, and adapt to provide a description of this bugfix.

Gonna give it a try today

pettitpeon:
Dear @bskinn @jun66j5 @bitprophet, What else is needed for the merge?

thanks!

paramiko/pipe.py
```diff
@@ -46,25 +47,50 @@ def __init__(self):
...
-        os.close(self._rfd)
-        os.close(self._wfd)
-        # used for unit tests:
         self._closed = True
 
+        # We cannot do anything about closing errors. It is only a
+        # "best effort" approach
+        try:
+            os.close(self._rfd)
+        except Exception:
```
jun66j5, line 56: Why `OSError` is not used rather than `Exception`?

jun66j5:
This PR is to fix issues in `PosixPipe`. I think `WindowsPipe` has the same issue. Any plans to fix it?

paramiko/pipe.py
pettitpeon: In this case we want to be as broad as possible in the try/except. If any exception is raised during the closing of the rd file descriptor, we still want to guarantee the closing of the wr file descriptor. Catching only one exception type could be too narrow and leave the pipe in an invalid state (one descriptor open and the other one closed)

pettitpeon:
> This PR is to fix issues in `PosixPipe`. I think `WindowsPipe` has the same issue. Any plans to fix it?

TBH, I did not look into the windows pipe since I use paramiko only in linux. If desired I can look into it and replicate the changes in the WindowsPipe. Still I'd like to merge this fix asap to remove my local patches

paramiko/pipe.py
pettitpeon: Yes, the problem with catching `OSError` is that if os.close() raises a different exception for whatever unknow reason, the pipe could go into an invalid state and potentially leak opened file descriptors.

pettitpeon:
```py
        try:
            os.close(self._rfd)  # imagine this throws RuntimeError
        except OSError:
            pass
        try:
            os.close(self._wfd)  # in that case, the _wrd would not be closed
        except OSError:
            pass
```

bskinn: I'll leave this `except` as a decision for bitprophet.

bskinn:
I figure this is up to bitprophet whether he wants to merge with just the `PosixPipe` fix and leave the `WindowsPipe` fix for a later PR, or roll them all up in one.

sites/www/changelog.rst
```diff
@@ -1619,3 +1619,6 @@ Changelog
 - :feature:`80` Expose the internal "is closed" property of the file transfer
   class ``BufferedFile`` as ``.closed``, better conforming to Python's file
   interface.  Thanks to ``@smunaut`` and James Hiscock for catch & patch.
+- :bug:`2271` Fixes pipe ``write()`` after ``close()`` race condition in pipe.py.
```
bskinn, line 1622:
```suggestion
- :bug:`2271` Fixes pipe ``write()`` after ``close()`` race condition in ``pipe.py``.
```

bskinn (commented):
Formatting nit.

bskinn:
It looks like the two failing CircleCI checks were due to an error on CircleCI's side. Likely all is ok with the test suite here.

pettitpeon:
> It looks like the two failing CircleCI checks were due to an error on CircleCI's side. Likely all is ok with the test suite here.

Thanks I applied your formatting suggestion. I'll await the final review. Cheers

bskinn:
Great, thanks! And, thanks for your patience here. bitprophet is in the middle of a dry spell in terms of his open source bandwidth, which has slowed everything down across his projects.

PR #2275, opened 2023-07-29
Fix deprecated test methods in gssapi tests

mgorny:
Replace the deprecated test methods .assertEquals(), .assertNotEquals() and .assert_() in gssapi tests with their modern equivalents.  These methods have been deprecated since Python 3.3, and have finally been removed in Python 3.12.

I can't say this makes all the tests pass for me but it at least fixes the obvious errors, so it's an overall improvement.

### DIFF
--- a/tests/test_gssapi.py
+++ b/tests/test_gssapi.py
@@ -54,7 +54,7 @@ class GSSAPITest(KerberosTestCase):
 
         oid = encoder.encode(ObjectIdentifier(self.krb5_mech))
         mech, __ = decoder.decode(oid)
-        self.assertEquals(self.krb5_mech, mech.__str__())
+        self.assertEqual(self.krb5_mech, mech.__str__())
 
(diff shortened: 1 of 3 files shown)


bskinn:
Thanks for this, definitely need to modernize here.

Is this the only test file where the deprecated methods are used? It would be good to do a universal cleanup on this all in one pass, I think.

mgorny:
Well, the PR touches 3 files but yes, these are all I've been able to find. Note that I've been grepping for `assertEquals`, `assertNotEquals` and `assert_(`.

bskinn:
Lol, whoops - typed that in the singular without thinking.

Sounds good, marked as for-review.

dluyer:
This misses a couple of the assert_ changes, i.e.:

        self.assert_(self.event.is_set())
        self.assert_(self.ts.is_active())

Issue #2246, opened 2023-05-26
sdist is missing dev-requirements.txt


mtelka:
I'm packaging paramiko for OpenIndiana and I noticed that tests for paramiko fails since there are some packages required by tests not installed.  It looks like the list of test dependencies is in the `dev-requirements.txt` file, but such file is missing from sdist.

Please add `dev-requirements.txt` to sdist to make downstream testing easier.  Thank you.

bskinn:
Seems straightforward enough. While we're looking at this, can you tell if there are any other files missing from the sdist?  That is, if you copy over `dev-requirements.txt` and pip install it, do they run cleanly?

We'd welcome a PR adding the entry to `MANIFEST.in`.

marston19:
Hi, seems super simple, mind if I take this?

PR #2286, opened 2023-08-21
Include dev-requirements.txt in MANIFEST.in. Issue #2246

marston19:
Addresses #2246 - Include dev-requirements.txt in MANIFEST.in so sdist packaging get the necessary requirements to install and run tests.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,4 @@
-include LICENSE setup_helper.py pytest.ini
+include LICENSE setup_helper.py pytest.ini dev-requirements.txt
 recursive-include docs *
 recursive-include tests *.py *.key *.pub
 recursive-include tests/configs *


bskinn:
@mtelka, can you test your packaging workflow on @marston19's feature branch to make sure the only thing missing in the manifest is `dev-requirements.txt`?

mtelka:
I tried to run tests for `3.3.1` both github tarball and PyPI sdist and test results are same.  So the only difference in my use case is missing `dev-requirements.txt` file.  Sorry, I cannot test the branch because I have no sdist from it.  I can try to build it myself, but IIRC the result highly depends on tooling used to build sdist.

Thank you.

bskinn:
> Sorry, I cannot test the branch because I have no sdist from it.

Ahh, of course, you need the sdist as it would be built from the project side.

I'll look into this and see what I can do for you.

bskinn:
Although, looking at the CircleCI output, it [_did_ include `dev-requirements.txt` in the manifest when it built](https://app.circleci.com/pipelines/github/paramiko/paramiko/683/workflows/ed4bd1f2-bbbd-49a1-a686-782cba1816b5/jobs/6241?invite=true#step-107-184).

So, the easiest (albeit slow) approach might be to merge and release this MANIFEST.in change, and then see how it works. If the MANIFEST still isn't right, we can fix further.

mtelka:
That sounds like a plan :-).  Please note that I do not need new `paramiko` release with this fix only.  For the time being I just switched to github tarball.

Thank you!

mtelka:
While talking about this...  If you are open to make my life even more easier, then something like [this](https://github.com/typeddjango/pytest-mypy-plugins/pull/121) would be highly appreciated.

Thank you!

bskinn:
> While talking about this... If you are open to make my life even more easier, then something like [this](https://github.com/typeddjango/pytest-mypy-plugins/pull/121) would be highly appreciated.

That one would be bitprophet's call.

Could you create a separate PR for adding the stub `tox.ini` and other related changes? It's a separate question from this ticket, and having it split out will make it easier to review/evaluate?

mtelka:
> Could you create a separate PR for adding the stub `tox.ini` and other related changes? It's a separate question from this ticket, and having it split out will make it easier to review/evaluate.

Here it is: #2288

Issue #396, opened 2014-09-17
FIPS - cannot use MD5


cingeyedog:
can the pkey.py be updated so that it doesnt use MD5?  

MD5 cannot be used in conjunction with FIPS.  

Directions for turning on FIPS in RHEL6: https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Security_Guide/sect-Security_Guide-Federal_Standards_And_Regulations-Federal_Information_Processing_Standard.html

yaccz:
ssh -vvv <host-working-with-paramiko>
debug1: kex: server->client aes128-ctr hmac-md5-etm@openssh.com none

ssh -vvv <fips-enabled-host-not-working-with-paramiko>
debug1: kex: server->client aes128-ctr hmac-sha1 none

So I guess, exactly what the OP says

Ismael:
This pull request fixes this: https://github.com/paramiko/paramiko/pull/1103
To install run:

>  pip install -e git+https://github.com/ioparaskev/paramiko#egg=paramiko

dsethi:
Has this issue been solved? We are running into same issue where we have to use paramiko on a fips enabled kernel. In case not, are there alternate python ssh libraries for this?

thukk:
> This pull request fixes this: #1103
> To install run:
> 
> > pip install -e git+https://github.com/ioparaskev/paramiko#egg=paramiko

this does indeed allow Paramiko to work on FIPS enabled machines.  There is a CryptographyDeprecationWarning for ecdsakey.py and kex_ecdh_nist.py.  It recommends using EllipticCurvePublicKey instead of EllipticCurvePublicNumbers.

dsethi:
Thanks for that Thukk. Since the PR is not merged/failed some tests initially, did it lead to any issues? How long have you run this change?

thukk:
> Thanks for that Thukk. Since the PR is not merged/failed some tests initially, did it lead to any issues? How long have you run this change?

Thus far, no.  Set it up today.  I am using Paramiko as a part of Fabric.  My Fabric scripts are now working (aside from the warning I mentioned before).

avidspartan1:
We've found the paramiko transport to work well for us, except when it comes to FIPS. Would love to see this PR accepted.

subodh2691:
I am facing similar issue and cloning of repo outside org is blocked in my machine. When this change will be going in to latest version of paramiko?

naman0805:
[...]

PR #2287, opened 2023-08-22
Switch debug logging in `client.py` to use `key.fingerprint`

bskinn:
Per #396, FIPS mode rejects .get_fingerprint() because it uses MD5.

The .fingerprint property uses SHA256, which is FIPS-compatible.

This addresses the latest FIPS problem reported in #396. We'll see if it solves all the FIPS problems.

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -640,7 +640,7 @@ class SSHClient(ClosingContextManager):
         # when #387 is released, since this is a critical log message users are
         # likely testing/filtering for (bah.)
         msg = "Trying discovered key {} in {}".format(
-            hexlify(key.get_fingerprint()), key_path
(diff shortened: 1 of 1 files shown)


bskinn:
@gubenkoved, can you test with my fork branch to see if this clears the path for FIPS?

bskinn:
No specific plans, but it seems like something that should be pretty quickly mergeable. I'll push it up to bitprophet to see if he has bandwidth to tackle.

gubenkoved:
Thank you @bskinn, I appreciate that and I think that change is going to be helpful for others as well!

Evan0000000000:
This has been sitting for quite a while, is there any chance someone could take a look to get it finalized? It would address something that's been outstanding since 2014.

PR #2288, opened 2023-08-23
Add tox.ini

### LINKED ISSUE #2246: sdist is missing dev-requirements.txt

mtelka:
A minimal `tox.ini` file.  It will allow fully automatic packaging and testing of `paramiko` package for OpenIndiana.

Closes #2246
Closes #2286

Thank you.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,4 @@
-include LICENSE setup_helper.py pytest.ini
+include LICENSE setup_helper.py pytest.ini dev-requirements.txt tox.ini
 recursive-include docs *
 recursive-include tests *.py *.key *.pub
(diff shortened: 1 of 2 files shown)


bskinn:
Mm, I don't think this supersedes #2286, at least not in general, even if it does close #2246.

There may be other repackaging situations where `dev-requirements.txt` would need to be in the manifest, so I still want to keep #2286 on the table.

PR #2291, opened 2023-08-28
Support certificate-based known hosts

liuyangc3:


bskinn:
Thanks for submitting this!

However, #2270 is already covering this, and is on course to be a more feature-complete implementation of handling `@cert-authority` entries. Accordingly, I'm going to close this in favor of that PR.

PR #2323, opened 2023-11-03
do not abort connection on unsupported public key

### LINKED ISSUE #2322: [BUG] -

Server

Keys/auth

2.8.1 - 3.3.1

2.7, 3.10.12

Ubuntu 22.04

_No response_

_No response_

### Expected/desired behavior

At any moment of a time, paramiko supported set of keys may be less than the new official one, or some proprietary extensions.
OpenSSH in such cases just ignore authentication attempts, while paramiko forbids and closes connection.
Since client's ssh-agent can contain such unsupported keys, and can attempt them in any order, it is very hard to use it simultaneously with paramiko powered servers, and servers, that accept these keys.

It would be right to just to ignore unknown keys too.

### Actual behavior

Paramiko logs: `Auth rejected: unsupported or mangled public key (...)` and sends `_disconnect_no_more_auth` to the client.

### How to reproduce

Right now I have no good public example, but the case is pretty obvious. As an example (already supported in paramiko) I can give recent lack of support for rsa-sha2-256 kind of keys and certificates.

### Anything else?

This code originates from 17 years ago and I found no real reason why it was done so, so I assume it is not intentional behavior, and can safely be changed to match OpenSSH one.

torkve:
This PR fixes #2322

### DIFF
--- a/paramiko/auth_handler.py
+++ b/paramiko/auth_handler.py
@@ -625,13 +625,15 @@ Error Message: {}
                 msg = "Auth rejected: unsupported or mangled public key ({}: {})"  # noqa
                 self._log(INFO, msg.format(e.__class__.__name__, e))
                 key = None
+
             if key is None:
-                self._disconnect_no_more_auth()
-                return
-            # first check if this key is okay... if not, we can skip the verify
-            result = self.transport.server_object.check_auth_publickey(
-                username, key
-            )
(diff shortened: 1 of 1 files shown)


torkve:
@bitprophet Hi, may I kindly ping you to review this one?

PR #2338, opened 2023-12-19
avoid KeyError when canonicaldomains is missing

totaam:
paramiko assumes that `canonicaldomains` is always present when `CanonicalizeHostname` is set, and it errors out when it is not, preventing the whole host config from being parsed.

Xpra-org/xpra#3868 (comment)


paramiko/config.py
```diff
@@ -233,7 +233,7 @@ def lookup(self, hostname):
         if canon and hostname.count(".") <= maxdots:
             # NOTE: OpenSSH manpage does not explicitly state this, but its
             # implementation for CanonicalDomains is 'split on any whitespace'.
-            domains = options["canonicaldomains"].split()
+            domains = options["canonicaldomains", ""].split()
```
bskinn, line 236: If you're going for a fallback to the empty string, it should be:

```suggestion
            domains = options.get("canonicaldomains", "").split()
```

bskinn:
It would be good to add a test case for the bugged situation here, to guard against future regression -- is that something you could tackle, @totaam?

Also, we'll want a CHANGELOG entry for this -- please add a bullet to `/sites/www/changelog.rst` following the format used there.

Thanks for this!

totaam:
I can take a look, not sure when.
(I am travelling, hence the botched conversion of the working patch into a borken PR)

bskinn:
No worries on timeline, @totaam, happy to take care of this whenever you're able.

totaam:
The updated PR adds a test which verifies that paramiko no longer blows up when a config specifies `canonicalizehostname` without also specifying `canonicaldomains`.

The failing check is not related to the changes I made so I am ignoring it.

totaam:
Please review this PR, the failing test is not caused by it.

totaam:
> No worries on timeline, @totaam, happy to take care of this whenever you're able.

@bskinn how about now?

stdedos:
Any _bump_ luck in someone or you reviewing this? @bskinn

PR #2420, opened 2024-07-22
Fix detection of SHA1 signing support

pghmcfc:
`e._reason` is an enum from `cryptography.exceptions._Reasons` so `is` should be the correct comparison, but it doesn't always work. This is apparently triggered by `_Reasons` moving to the part of `cryptography` that is implemented in rust, which doesn't yet implement enums as singletons.

https://github.com/pyca/cryptography/issues/11332
https://github.com/PyO3/pyo3/issues/3059

### DIFF
--- a/tests/_util.py
+++ b/tests/_util.py
@@ -188,7 +188,7 @@ def sha1_signing_unsupported():
         )
         return False
     except UnsupportedAlgorithm as e:
-        return e._reason is _Reasons.UNSUPPORTED_HASH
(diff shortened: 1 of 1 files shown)


alex (approved):

bitprophet:
Thanks @pghmcfc + @alex! I wasn't able to repro the underlying issue on my end (Python 3.8, Cryptography 43 and 42) but I trust the fix and have backported it to Paramiko 3.3/3.4. Will be released later today.

PR #2435, opened 2024-08-13
diffie-hellman-group18-sha512 Kex Algorithm Support Addition To Paramiko

takhtart:
Paramiko's Current SSHClient does not support SSH to devices that use the diffie-hellman-group18-sha512 Kex Algorithm. This small change adds the ability to do so.


takhtart:
SSH Connection Tested On Python 3.11, Paramiko version 3.4.1, On Commercial Ruggedcom ROX RX1524

takhtart:
Any update for this addition? Would like to interface with devices utilizing this key exchange algorithm to connect.

simplymacbook:
Hey @takhtart, This is great. Have you had any issues with the new paramiko not supporting ssh-dss. I believe the ruggedcom ROS switches still use ssh-dss but they have removed this since 2.10 or something. All the new paramiko 4.0.0 version only do ssh-rsa. Did you have any work around for ROS?

takhtart:
Hey @simplymacbook 

Newer firmwares of ROS have (since the 3.11 or 3.12 version) quite some time ago moved away from ssh-dss being the default key pair for a while, other than (if you can) upgrading the firmware of the switches the only other way is to use an older release of paramiko for your needs. I don't encounter this particular issue because the switches I interact with have newer firmware.

> Hey @takhtart, This is great. Have you had any issues with the new paramiko not supporting ssh-dss. I believe the ruggedcom ROS switches still use ssh-dss but they have removed this since 2.10 or something. All the new paramiko 4.0.0 version only do ssh-rsa. Did you have any work around for ROS?

KaylaSolace:
does this add support for the ssh server to use diffie-hellman? Hoping to avoid clients connecting to a paramiko ssh server from getting the following warning:
```
** WARNING: connection is not using a post-quantum key exchange algorithm.
** This session may be vulnerable to "store now, decrypt later" attacks.
** The server may need to be upgraded. See https://openssh.com/pq.html
```

Edit: I am not very familiar with key exchange, I just know there are post quantum diffie-hellman setups out there and could not tell if this is one of them!

Edit again: Yeah I probably misunderstood, sorry!

nielsendres:
Just adding a note to say this worked for me on Python 3.14 with Paramiko 4.0.0.   Also against an RX1524, which had no way of connecting otherwise.

takhtart:
> does this add support for the ssh server to use diffie-hellman? Hoping to avoid clients connecting to a paramiko ssh server from getting the following warning:
> ```
> ** WARNING: connection is not using a post-quantum key exchange algorithm.
> ** This session may be vulnerable to "store now, decrypt later" attacks.
> ** The server may need to be upgraded. See https://openssh.com/pq.html
> ```
> 
> Edit: I am not very familiar with key exchange, I just know there are post quantum diffie-hellman setups out there and could not tell if this is one of them!
> 
> Edit again: Yeah I probably misunderstood, sorry!

Hey @KaylaSolace!

No worries If I understand correctly when sshing to a device the destination device will tell your ssh client the available key exchange algorithms you can use to complete the connection. For example in my case only the diffie-hellman-group18-sha512 was the only key exchange algorithm offered by the device I was trying to connect to, which Paramiko had no notion of and so SSHing into the device would fail, and this small change simply added that algo.

Normally a device would offer various kex algos, and your SSH Client I assume would pick the highest security level algo that it recognizes and supports, in my case the RX1524 I was trying to reach uses a higher security algo that is not yet available in Paramiko which I would like to see in a future update, which is what is keeping me on this version with my personal additions at the moment.

I believe post quantum is an entirely different key exchange algo type rather than the diffie-hellman series that this pull req addresses, but I have not looked into if Paramiko supports this at all. I assume the device you're trying to reach offers both post quantum and diffie hellman kex algorithms and I'm assuming Paramiko chooses diffie- hellman, the destination device recognizes this and provides a warning, but allows SSH access (Post Quantum is not enforced).

Hope that makes a little more sense :)

-Tarnveer

takhtart:
> Just adding a note to say this worked for me on Python 3.14 with Paramiko 4.0.0. Also against an RX1524, which had no way of connecting otherwise.

Hey @nielsendres!

Glad you we're able to integrate my fix to v4.0 from my 3.4.1 build base. Luckily this addition is simple enough to add even with the rolling releases ahead, but would be great if its integrated in future releases as well. Just for flexibility my release roof is 3.5.1 at the moment in case I want to work with a legacy device that uses DSA (DSS) through Paramiko so I did not get the chance to integrate and test this personally for v4.0.0. 

I'm glad you're able to connect to that RX1524 now! This will be helpful for other's who are are willing to add this small bit of code for themselves in these newer releases :)

-Tarnveer

Issue #2440, opened 2024-08-31
[BUG] - BufferedFile.read can fail to return buffered when timeouts occur


smoofra:
### Are you using paramiko as a client or server?

Client

### What feature(s) aren't working right?

SSH

### What version(s) of paramiko are you using?

3.4.1

### What version(s) of Python are you using?

3.12.5

### What operating system and version are you using?

Mac OS 14.6.1

### If you're connecting as a client, which SSH server are you connecting to?

BitVise ssh server

### Expected/desired behavior

I'm executing a command that using `sss.exec_command(.. timeout=0.1)` 
The command produces some output and then hangs on the remote side.
I expect that reading the output will return all the output that the remote sent. and then raise TimeoutError.

### Actual behavior

It raises TImeoutError while keeping some of the output stuck in BufferedFile._rbuffer

### How to reproduce

use paramiko to run a command like "cat some_large_file && sleep 9999999"  and try to read all the output

Issue #515, opened 2015-04-24
Creating a new SSH connection sometimes hangs forever


appidman:
I use paramiko to create SSH sessions and send commands to Linux machines, which works great 95% of the time. But sometimes, the Python thread which is used to execute the command hangs forever in wait() of threading.py:

```
(gdb) py-list
 239            waiter.acquire()
 240            self.__waiters.append(waiter)
 241            saved_state = self._release_save()
 242            try:    # restore state no matter what (e.g., KeyboardInterrupt) 
 243                if timeout is None:
>244                    waiter.acquire()
 245                    if __debug__:
 246                        self._note("%s.wait(): got it", self)
 247                else:
 248                    # Balancing act:  We can't afford a pure busy loop, so we
 249                    # have to sleep; but if we sleep the whole timeout time,


(gdb) py-bt
#5 Frame 0xb459a5c, for file /usr/lib/python2.7/threading.py, line 244, in wait (self=<_Condition(_Verbose__verbose=False, _Condition__lock=<thread.lock at remote 0xaf7e3a0>, acquire=<built-in method acquire of thread.lock object at remote 0xaf7e3a0>, _Condition__waiters=[<thread.lock at remote 0xaf7e100>], release=<built-in method release of thread.lock object at remote 0xaf7e3a0>) at remote 0xb73708c>, timeout=None, waiter=<thread.lock at remote 0xaf7e100>, saved_state=None)
    waiter.acquire()
#9 Frame 0xb66118c, for file /usr/lib/python2.7/dist-packages/paramiko/buffered_pipe.py, line 137, in read (self=<BufferedPipe(_closed=False, _lock=<thread.lock at remote 0xaf7e3a0>, _buffer=<array.array at remote 0xb737ea0>, _event=None, _cv=<_Condition(_Verbose__verbose=False, _Condition__lock=<thread.lock at remote 0xaf7e3a0>, acquire=<built-in method acquire of thread.lock object at remote 0xaf7e3a0>, _Condition__waiters=[<thread.lock at remote 0xaf7e100>], release=<built-in method release of thread.lock object at remote 0xaf7e3a0>) at remote 0xb73708c>) at remote 0xb737a4c>, nbytes=8192, timeout=None, out='', then=<float at remote 0xb57efe4>)
    self._cv.wait(timeout)
[...]

PR #2441, opened 2024-09-02
check for TimeoutError in BufferedFile

smoofra:
Check for TimeoutError in BufferedFile so that data does not languish in the buffer after a timeout occurs.

Fixes: #2440

### DIFF
--- a/paramiko/file.py
+++ b/paramiko/file.py
@@ -201,6 +201,11 @@ class BufferedFile(ClosingContextManager):
                 read_size = max(self._bufsize, read_size)
             try:
                 new_data = self._read(read_size)
+            except TimeoutError:
+                if not self._rbuffer:
+                    raise
+                else:
+                    new_data = b""
             except EOFError:
                 new_data = None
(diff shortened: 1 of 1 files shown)


annonm:
Are configured timeouts actually respected in a way that would cause a TimeoutError to be raised here?

smoofra:
> Are configured timeouts actually respected in a way that would cause a TimeoutError to be raised here?

Yes, they seem to be.

Here is a script I use to run go tests remotely, using paramiko.   It sets a timeout.

https://gist.github.com/smoofra/63dee4d4337705716d5fc203e7084df4

annonm:
Ah okay, thanks for sharing that. I see you are setting the timeout directly on the Channel, which is why this works.

I'm not sure if you saw #515  - but I believe, at least for users who aren't directly setting the Channel timeout like this, they would never actually time out.

Curious - without your code to add a catch to **TimeoutError,** wouldn't some sort of "PipeException" occur?

Issue #2445, opened 2024-09-12
[BUG] - Sometimes loading an ECDSAKey or DSSKey with a comment of certain lengths results in "Invalid key".


Styca96:
### Are you using paramiko as a client or server?

Client

### What feature(s) aren't working right?

Keys/auth

### What version(s) of paramiko are you using?

3.4.1

### What version(s) of Python are you using?

3.10.12

### What operating system and version are you using?

WSL on Windows 11

### Expected/desired behavior

Load any private file with ECDSAKey without getting an error

### Actual behavior

Sometimes when I try lo load a private ECDSAKey from a file I get an error:

> paramiko.ssh_exception.SSHException: Invalid key curve identifier

The problemis is in the function `_unpad_openssh` in `pkey.py` module.

```python
def _unpad_openssh(data):
    # At the moment, this is only used for unpadding private keys on disk. This
    # really ought to be made constant time (possibly by upstreaming this logic
    # into pyca/cryptography).
    
    padding_length = data[-1]
    if 0x20 <= padding_length < 0x7:
        return data  # no padding, last byte part comment (printable ascii)
    if padding_length > 15:
        raise SSHException("Invalid key")
    for i in range(padding_length):
        if data[i - padding_length] != i + 1:
            raise SSHException("Invalid key")
    return data[:-padding_length]
```

When the `padding_length` is equal to 0, this function return an empty list that cause an Exception. 
I don't kwon about padding and comment so I don't kwon if it is the correct behaviour.

The key is valid if we change this line:

> if 0x20 <= padding_length <  0x7F:

in this one:

> if 0x20 <=  padding_length < 0x7F or padding_lenght == 0x00:

### How to reproduce

By using this snippet of code you can see that sometimes the load of the key fail
```python
import os
import paramiko

def path(_type, n):
    return f'/tmp/key_{_type}_{n}'

passphrase = ''  #  an empty passphrase is much faster
comment = ""
CYCLE = 10

for _ in range(CYCLE):
    for _type in ['rsa','ecdsa','ed25519','dsa']:
        os.system('rm -f %s' % path(_type, 0))
        os.system('ssh-keygen -t %s -N "%s" -C "%s" -f %s > /dev/null' % (_type, passphrase, comment, path(_type, 0)))

        #  see which keys can be loaded
        
        s = comment
        try:
            match _type:
                case 'rsa':
                    pkey = paramiko.RSAKey.from_private_key_file(filename=path(_type,0), password=passphrase)
[...]

PR #2446, opened 2024-09-12
Fix ECDSA and DSS key zero padding_lenght

Styca96:


Styca96:
fixes https://github.com/paramiko/paramiko/issues/2445

shapovalovts:
Any reason why this is not merged to 3.5 or 4.0?

PR #2456, opened 2024-09-22
Return new file position from SFTPFile.seek

### LINKED ISSUE #2452: [FEAT] - Implement io.IOBase protocol for opened SFTP file objects
### Is this feature for paramiko acting as a client or a server?

Client

### What functionality does this feature request relate to?

SFTP

### For client-side features, does this relate to a specific type of SSH server?

_No response_

_No response_

### Desired behavior

I am opening a file object via [SFTPClient.open](https://docs.paramiko.org/en/latest/api/sftp.html#paramiko.sftp_client.SFTPClient.open). The returned file object looks very much like a typical Python file object, but differs in two ways noticable for my use case:

 - It inherits from `paramiko.sftp_file.SFTPFile` and does not have `io.IOBase` in its inheritance hierarchy, i.e., `isinstance(fileobj, io.IOBase)` does not work.
 - [`paramiko.sftp_file.SFTPFile.seek`](https://docs.paramiko.org/en/latest/api/file.html#paramiko.file.BufferedFile.seek) does not return the new file position. This is in stark contrast to [io.IOBase.seek](https://docs.python.org/3/library/io.html#io.IOBase.seek), which "return[s] the new absolute position".

I guess, I will need to either inject-replace a new seek method or more cleaner, but maybe worse for performance and code conciseness, wrap `paramiko.sftp_file.SFTPFile` in a new class that correctly implements the `io.IOBase` protocol.

### Anything else?

_No response_

mxmlnkn:
Fixes #2452.

### DIFF
--- a/paramiko/file.py
+++ b/paramiko/file.py
@@ -354,6 +354,7 @@ class BufferedFile(ClosingContextManager):
         :param int whence:
             type of movement: 0 = absolute; 1 = relative to the current
             position; 2 = relative to the end of the file.
+        :returns: file position (`number <int>` of bytes).
 
         :raises: ``IOError`` -- if the file doesn't support random access.
         """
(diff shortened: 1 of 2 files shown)


SupraSummus:
I would really appreciate merging and releasing this.

PR #2465, opened 2024-10-09
Add check for public_blob attr.

### LINKED ISSUE #2462: [BUG] - AttributeError(public_blob)

Client

Keys/auth

3.5.0

3.9.6

MacOS Sonoma, RHEL 7, RHEL8

OpenSSH

Fabric

### Expected/desired behavior

Successful authentication

### Actual behavior

An `AttributeError: public_blob` exception is being thrown by `AgentKey.__getattr(self, name)`.

### How to reproduce

```
import fabric

c = fabric.Connection('hostname')
output = c.run('hostname').stdout.strip()
print(output)
```

The code here isn't very interesting.  This may only happen when using certificates, and this doesn't seem very common.

There are several other bug reports that seem to be hitting this issue for various reasons.  Eg, #2320 and #2299.  

### Anything else?

The primary issue seems to be how the inner_key value of the `AgentKey` class is being used.

In the `__init__` method of `AgentKey` where `self.inner_key` is being initialized, there is a comment in the exception handler that inner_key is best-effort and to not explode.  That tells me that its allowable for this value to be None.

In the `__get_attr__` method of `AgentKey`, the code checks if `self.inner_key` is None and raises an AttributeError.  This matches the expected behavior for `getattr` as described at https://docs.python.org/3/library/functions.html#getattr.

The issue seems to be happening in the `AuthHandler._get_key_type_and_bits` method.  This is the code in question:
```
# Use certificate contents, if available, plain pubkey otherwise
if key.public_blob:
    return key.public_blob.key_type, key.public_blob.key_blob
else:
    return key.get_name(), key
```

This code assumes that the public_blob attribute exists on key but might be None.  Behind the scenes, this is calling the `AgentKey.__get_attr__` method.  If `inner_key` isn't set, this causes the exception.

I've tested two fixes, and they both seem to work.

The first fix is to not throw an exception in AgentKey, and instead return None.  That looks like this:
```
    def __getattr__(self, name):
        """
        Proxy any un-implemented methods/properties to the inner_key.
        """
        if self.inner_key is None:  # nothing to proxy to
            return None
        return getattr(self.inner_key, name)
```
While this works, it does seem to break the behavior described by the documentation for `getattr`.

The other fix, which is probably better, is to check for the attribute in `AuthHandler`.  That would look like this:
```
# Use certificate contents, if available, plain pubkey otherwise
if hasattr(key, public_blob):
    return key.public_blob.key_type, key.public_blob.key_blob
else:
    return key.get_name(), key
```

The `hasattr` function calls `getattr`, and checks if it raises the `AttributeError`.  The exception happens, but its handled gracefully and sends it down the else path successfully.

I'm happy to PR either of these fixes, but both are pretty trivial.

Thanks,

Paul

J08nY:
This fixes #2462.

Thanks for the clear write up @paulf69487623.

### DIFF
--- a/paramiko/auth_handler.py
+++ b/paramiko/auth_handler.py
@@ -215,7 +215,7 @@ class AuthHandler:
         Intended for input to or verification of, key signatures.
         """
         # Use certificate contents, if available, plain pubkey otherwise
-        if key.public_blob:
+        if hasattr(key, "public_blob"):
             return key.public_blob.key_type, key.public_blob.key_blob
         else:
             return key.get_name(), key


johannes85 (commented):
I have the same problem at the moment and in a search for a solution I found your pull request. In my case this change breaks the case if key.public_blob is None. Since the hasattr check passes but key.public_blob.key_type ends in a `AttributeError: 'NoneType' object has no attribute 'key_type'` error.

oversize:
I have the same issue as @johannes85 . The attribute is there but None so if not done as @described here https://github.com/paramiko/paramiko/issues/2462#issuecomment-2418896272 it will still fail for me. 
I do have the private key stored on a yubikey and do not actually have a public key localy anyhow.

paulf69487623:
I haven't seen any further activity on the PR to add the extra fixes, so I've opened #2475 .

PR #2475, opened 2024-11-20
Fix AttributeError(public_blob) #2462

### LINKED ISSUE #2462: [BUG] - AttributeError(public_blob)

paulf69487623:
This should fix #2462.  This extends the solution in PR #2465, which wasn't 100% complete.  This incorporates the extra check from @johannes85.

### DIFF
--- a/paramiko/auth_handler.py
+++ b/paramiko/auth_handler.py
@@ -215,7 +215,7 @@ class AuthHandler:
         Intended for input to or verification of, key signatures.
         """
         # Use certificate contents, if available, plain pubkey otherwise
-        if key.public_blob:
+        if hasattr(key, "public_blob") and key.public_blob:
             return key.public_blob.key_type, key.public_blob.key_blob
         else:
             return key.get_name(), key


Sakoes:
+1 For merging this fix

austinhallett:
+1 let's get this merged in

supertunaman:
This patch works great for my use-case. I auth using an ed25519 key that gets signed by a CA in order to allow auth. This bug was preventing my script from working with this setup, but this patch fixes it.

paulf69487623:
Hey @bitprophet,

When you get a free moment, could you take a look at this PR?  It seems like its helping a variety of people who need to use certificate-based authentication.

Thanks,

Paul

staticglobal:
+1 for merging this. Paramiko is broken out of the box for us

JakkuSakura:
This should be fixed ASAP

jonxstill:
Just adding another +1 to this as currently Paramiko isn't working straight out the gate.

iain-henderson:
Another +1

Sakoes:
@paulf69487623 Could you add @bitprophet as a reviewer and maybe add some kind of "Ready for review tag" to the PR? He seems like one of the few people actively maintaining the project.

mherrb:
+1 too to merge this...

PR #2496, opened 2025-02-10
Flag md5 usage as not for security

### LINKED ISSUE #2383: [BUG] - Use of MD5 algorithm in FIPs system

Client

SSH, SFTP, Keys/auth, known_hosts

3.4

3.11

Rocky 9

_No response_

_No response_

### Expected/desired behavior

Being able to open a ssh connection.

### Actual behavior

When using a systems with FIPS enabled md5 hash will fail.
In my case i couldn't even open a connection because of PKey when using key auth or when adding a new host. Due to its use in the 
 debug log 
```python
        client._log(
            DEBUG,
            "Adding {} host key for {}: {}".format(
                key.get_name(), hostname, hexlify(key.get_fingerprint())
            ),
        )
```` 
https://github.com/paramiko/paramiko/blob/51eb55debf2ebfe56f38378005439a029a48225f/paramiko/pkey.py#L344 

### How to reproduce

_No response_

### Anything else?

Not use md5, not sure if sha1 would also be problematic with FIP. At least not depend on hashlib with depends on openssl which will fail. 
Maybe use sha256 instead?

steveb:
This is required for FIPS compliance. Without this change attempting to use md5 will raise an error.

Applied patch mopi3456 in bug #2383

Closes: https://github.com/paramiko/paramiko/issues/2383

### DIFF
--- a/paramiko/pkey.py
+++ b/paramiko/pkey.py
@@ -25,7 +25,15 @@ from base64 import encodebytes, decodebytes
 from binascii import unhexlify
 import os
 from pathlib import Path
-from hashlib import md5, sha256
+from hashlib import sha256
+import hashlib
+import sys
+from functools import partial
+
+if sys.version_info >= (3, 9):
+    md5 = partial(hashlib.md5, usedforsecurity=False)
+else:
+    md5 = hashlib.md5
 import re
 import struct
 


steveb:
The Test 3.7 failure appears to be unrelated

tiran:
Hi, author of the usedforsecurity flag here. This is a misuse of the feature and wrong approach to address the problem. The digest is used in a security context.

bskinn:
Concur, these changes are not appropriate. Thanks for the PR, but we won't be able to accept it.

PR #2498, opened 2025-02-13
NETNKM-2562 : Introducing OSError timebound busy-wait in packetizer _read_timeout

mdear:
While testing on an embedded automotive target from a Windows 11 client, the target was gracefully rebooted (which resulted in the system being taken down slowly and then hard-rebooted).

Around the time of hard-rebooting we noticed that `packet.py:Packetizer:_read_timeout` was not catching the following OSError raised when the `self.__socket.recv(128)` was attempted. 

`[WinError 10054] An existing connection was forcibly closed by the remote host`

This PR now catches the `OSError` as long as the timeout has not been exceeded.  If the issue resolves and data is able to be read from the socket when it is returned normally.  If the issue does not resolve and `OSError` continues to be raised then when the timeout has been exceeded it is re-raised.

### DIFF
--- a/paramiko/common.py
+++ b/paramiko/common.py
@@ -243,3 +243,7 @@ MIN_PACKET_SIZE = 2**12
 
 # Max windows size according to http://www.ietf.org/rfc/rfc4254.txt
 MAX_WINDOW_SIZE = 2**32 - 1
+
+# When OSError is thrown, wait this long if we want to spin and
+# wait for error conditions to resolve.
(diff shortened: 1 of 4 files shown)


mdear:
In case this question gets asked : I know there is another instance of socket.recv() call in `Packetizer:read_all`, but I wasn't seeing explicit problems so did not update the code there.  I only want to fix issues that I can prove are occurring.

mdear:
Created an issue https://github.com/paramiko/paramiko/issues/2499 that points to this PR.

PR #2503, opened 2025-03-11
new optional paramiko host policy: prompt

anarcat:
This policy tries to match a little closer what a "normal" SSH client does which is "trust on first use" (TOFU).

We do not actually set it as default, as I want to leave room for discussion and maximize the possibility of this being merged.

This is part of repeated attempts I have made at improving host key policy handling in Fabric, see:

https://github.com/fabric/fabric/pull/2076
https://github.com/paramiko/paramiko/pull/1647

To have this useful in Fabric, the former would be necessary.

Hopefully one of those will stick...

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -861,6 +861,34 @@ class AutoAddPolicy(MissingHostKeyPolicy):
         )
 
 
+def no_yes(prompt: str) -> bool:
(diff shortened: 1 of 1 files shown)


anarcat:
A possible improvement to this would be to accept the key hash as well, just like OpenSSH now does. I'm open to extending the PR if you think that's useful.

PR #2519, opened 2025-06-19
fix a race in ftp prefetching when using max_concurrent_requests

davidszotten:
when `max_concurrent_requests` is set, it's possible to empty the list of prefetch extents by reading faster than they are enqueued. currently this makes us mistakenly thing the prefetching is all done. reading still works, but it starts ignoring the prefetch which, in particular, means that in the common case of reading an entire file, the prefetch buffers continue filling but are never emptied causing excessive memory use.

fix this by explicitly keeping track of when we're done enqueueing.


ddelange:
does this mean we can pass `max_concurrent_requests` to control network i/o AND `bufsize` to control RAM usage (i.e. to stop prefetching when the buffer is full)?

bwinston-sdp:
nice find @davidszotten! this seems like the correct diagnosis to me, but I will defer to @bitprophet and/or other maintainers with a better mental model of the prefetch code.

PR #2543, opened 2025-08-20
Make `invoke` an optional dependency again

akx:
I was doing a routine dependency upgrade on a project and noticed `invoke` started getting pulled in by paramiko 4.0.0. I'm allergic to extra dependencies being pulled in, so I decided to investigate.

https://github.com/paramiko/paramiko/commit/dbfd52c50594164ed0cb884d90b1009bf48fccc6 (inadvertently? The commit message doesn't make a note of the change?) made `invoke`, only needed for `exec` credential helpers a non-optional dependency.

This pulls it back to a dependency group – and while at it, moves importing it to a very late stage, so no extra time is being spent importing it if exec credentials aren't used. This also simplifies the code a bit.

### DIFF
--- a/paramiko/config.py
+++ b/paramiko/config.py
@@ -31,12 +31,6 @@ from hashlib import sha1
 from io import StringIO
 from functools import partial
 
-invoke, invoke_import_error = None, None
-try:
-    import invoke
-except ImportError as e:
-    invoke_import_error = e
-
 from .ssh_exception import CouldNotCanonicalize, ConfigParseError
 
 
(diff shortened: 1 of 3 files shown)


bitprophet:
> I'm allergic to extra dependencies being pulled in

While I sympathize (and thank you for the feedback!) I don't share your allergy - it'd be one thing if I was, say, grabbing a _large_ transitive dependency tree for little gain, or requiring some annoying C-extension, but this gets a more compact and powerful subprocess-wrapping API from a single small pure-Python dependency.

So I file it alongside other deps like pynacl ("only used if you're relying on ed25519 keys") where it's simpler from a dev and support level to just require all the bits n pieces required for most functionality.

(FWIW, GSSAPI - which _is_ still an optional-only dep - is exempt from this due to being a lot more complicated along most axes [dependency tree, implementation, etc] _and_ being comparatively less commonly used.)

So, apologies, but I won't accept the patch.

PR #2544, opened 2025-08-23
Drop `invoke` dependency

akx:
Alternate take for #2543 – shouldn't need a dependency just to see whether a process returns `1`.


akx:
@bskinn Hmm. Based on the reply there, this would still be a viable approach though, since as far as I can tell, the standard library can do what the dependency would do here. Could we consider reopening this?

PR #2559, opened 2026-01-02
Fix CVE-2025-69277

loganaden:
https://00f.net/2025/12/30/libsodium-vulnerability/ pynacl is also impacted.


loganaden:
After some discussions, it might be a good idea to approach this differently. What is the policy about backwards compatibility ?

Issue #2320, opened 2023-11-01
[BUG] - Unable to use ssh-ed25519-cert-v01@openssh.com 


sagraw2:
### Are you using paramiko as a client or server?

Client

### What feature(s) aren't working right?

SSH, Keys/auth

### What version(s) of paramiko are you using?

3.3.1

### What version(s) of Python are you using?

3.8.10

### What operating system and version are you using?

Ubuntu 20.04

### If you're connecting as a client, which SSH server are you connecting to?

OpenSSH

### Expected/desired behavior

Trying to use a 'ssh-ed25519-cert-v01@openssh.com' loaded in ssh-agent to connect to a remote OpenSSH server.
This works fine with paramiko library version 3.1
This seems to be the offending commit :
https://github.com/paramiko/paramiko/commit/0146cf72012e46505851abe17d8eb3555a481967

Test Code:
`ssh = SSHClient()
 ssh.set_missing_host_key_policy(AutoAddPolicy())
 try:
     ssh.connect(hostname=some_server, port=22, username=some_username)
     print("Connected successfully !!")
except Exception:
     print("Exception from ssh.connect !!")
`

### Actual behavior

Fails with below error:

Traceback (most recent call last):
  File "paramiko_test.py", line 154, in <module>
    sys.exit(main())
  File "paramiko_test.py", line 62, in main
    ssh.connect(hostname=some_server, port=22, username=some_username)
  File "/home/someuser/.local/lib/python3.8/site-packages/paramiko/client.py", line 485, in connect
    self._auth(
  File "/home/someuser/.local/lib/python3.8/site-packages/paramiko/client.py", line 754, in _auth
    self._transport.auth_publickey(username, key)
  File "/home/someuser/.local/lib/python3.8/site-packages/paramiko/transport.py", line 1658, in auth_publickey
    return self.auth_handler.wait_for_response(my_event)
  File "/home/someuser/.local/lib/python3.8/site-packages/paramiko/auth_handler.py", line 248, in wait_for_response
    raise e
  File "/home/someuser/.local/lib/python3.8/site-packages/paramiko/transport.py", line 2185, in run
    handler(m)
  File "/home/someuser/.local/lib/python3.8/site-packages/paramiko/auth_handler.py", line 394, in _parse_service_accept
    key_type, bits = self._get_key_type_and_bits(self.private_key)
[...]

Issue #2299, opened 2023-09-07
[FEAT] - Support for ecdsa-sk and ed25519-sk


crabhi:
### Is this feature for paramiko acting as a client or a server?

Client

### What functionality does this feature request relate to?

Keys/auth

### For client-side features, does this relate to a specific type of SSH server?

OpenSSH

### If you're using paramiko as part of another tool, which tool/version?

Fabric, Ansible

### Desired behavior

Openssh v.8.2 added support for FIDO/U2F tokens introducing sk-ssh-ed25519@openssh.com and sk-ssh-ecdsa@openssh.com. See: https://www.openssh.com/txt/release-8.2

When such a key is present in the SSH agent, the client connection fails with (traceback from Ansible, but Fabric fails similarly)

```
Traceback (most recent call last):
  File "/home/krab/.virtualenvs/ansible/lib/python3.10/site-packages/paramiko/transport.py", line 2185, in run
    handler(m)
  File "/home/krab/.virtualenvs/ansible/lib/python3.10/site-packages/paramiko/auth_handler.py", line 394, in _parse_service_accept
    key_type, bits = self._get_key_type_and_bits(self.private_key)
  File "/home/krab/.virtualenvs/ansible/lib/python3.10/site-packages/paramiko/auth_handler.py", line 218, in _get_key_type_and_bits
    if key.public_blob:
  File "/home/krab/.virtualenvs/ansible/lib/python3.10/site-packages/paramiko/agent.py", line 476, in __getattr__
    raise AttributeError(name)
AttributeError: public_blob
```

It would be nice if this type of keys would be supported and Paramiko could be used instead of the OpenSSH client.

jakubman1:
Hi,
is there any progress on this issue?
Thanks.

strokovok:
Found some dirty workaround

Just load your key as usual, either from agent like this:
```python3
mykey = paramiko.Agent().get_keys()[0]
```

Or from file (if you have ECDSA):
```python3
mykey = paramiko.ecdsakey.ECDSAKey(filename="/path/to/id_ecdsa")
```

Or if you have ED25519:
```python3
mykey = paramiko.ed25519key.Ed25519Key(filename="/path/to/id_ed25519")
```

And then just do this:
```python3
if not hasattr(mykey, "public_blob"):
    mykey.public_blob = None
```

Then connect using this key:
```python3
client.connect("a.b.c.d", 22, username="user", pkey=mykey)
```

[...]

PR #2561, opened 2026-01-09
Fix Paramiko parsing of SSH certificate public_blob

calderonth:
When trying to use Paramiko with ssh certificates stored in the `ssh-agent`, we can observe failures to connect when using a simple connection script:

```
import paramiko
import os

def test_agent_cert_auth():
    client = paramiko.SSHClient()

    # This should ideally pick up the certificate from your running ssh-agent
    try:
        client.connect('localhost', username='calderonth', allow_agent=True)
        print("Authentication successful!")
    except Exception as e:
        print(f"Authentication failed: {e}")

if __name__ == "__main__":
    test_agent_cert_auth()
```
Running it produces the following stack trace:
```
sys-python test.py
Unknown exception: public_blob
  File ".../python3.11/site-packages/paramiko/transport.py", line 2262, in run
    handler(m)
  File ".../python3.11/site-packages/paramiko/auth_handler.py", line 394, in _parse_service_accept
    key_type, bits = self._get_key_type_and_bits(self.private_key)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/paramiko/auth_handler.py", line 218, in _get_key_type_and_bits
    if key.public_blob:
       ^^^^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/paramiko/agent.py", line 476, in __getattr__
    raise AttributeError(name)
AttributeError: public_blob

Authentication failed: public_blob
```

Example content of the ssh-agent:
```
ssh-add -l
256 SHA256:E2MBaXQgjsV7KJo1sAO7REyu5WI0U1kU4X9ElT25Owo calderonth@example.com (ECDSA-CERT)
```

The following PR implements a `@property` in the AgentKey class that satisfies the interface expected by AuthHandler without breaking the existing proxy logic.

Changes in paramiko/agent.py:
  * Added a public_blob property to the AgentKey class.
  * The property returns the inner_key.public_blob if it exists (for standard keys).
  * For keys where the inner key is missing (like certificates the agent handles but Paramiko hasn't fully parsed), it returns a namedtuple containing key_type (from self.name) and key_blob (from self.blob).
  
Testing & Validation
Added regression test to tests/agent.py using the pytest-relaxed style familiar to the Paramiko codebase

Other issues this may help with #2462 #2320  #2299

### DIFF
--- a/paramiko/agent.py
+++ b/paramiko/agent.py
@@ -29,6 +29,7 @@ import time
 import tempfile
 import stat
 from logging import DEBUG
+from collections import namedtuple
 from select import select
 from paramiko.common import io_sleep, byte_chr
 
@@ -467,6 +468,15 @@ class AgentKey(PKey):
         if self.inner_key is not None:
             return self.inner_key.get_bits()
         return super().get_bits()
+
+    @property
+    def public_blob(self):
+        # If Paramiko managed to parse a real PKey object (RSA/Ed25519), use its blob
+        if self.inner_key is not None and hasattr(self.inner_key, 'public_blob'):
+            return self.inner_key.public_blob
+        # Otherwise, wrap the raw agent blob so AuthHandler can still see
+        # the key_type (self.name) and the raw bytes (self.blob)
+        return namedtuple('BlobWrapper', 'key_type key_blob')(self.name, self.blob)
 
     def __getattr__(self, name):
         """
--- a/paramiko/agent.py
+++ b/paramiko/agent.py
@@ -472,11 +472,15 @@ class AgentKey(PKey):
     @property
     def public_blob(self):
         # If Paramiko managed to parse a real PKey object (RSA/Ed25519), use its blob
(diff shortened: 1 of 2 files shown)


calderonth:
PS: this is an alternative fix related to #2462 (includes unit-tests).

@bitprophet or @alex  would it be possible to take a look? We're heavy users of SSH user certificates and this make us unable to do agent-based authentication.

santosh0705:
I am having issue with Teleport ssh-multiplexer https://github.com/gravitational/teleport/issues/63015
This PR fix the issue.

link2xt:
I got the same problem while playing with https://github.com/openpubkey/opkssh, which generates SSH certificates, and later using [pyinfra](https://pyinfra.com/), which uses Paramiko. Worked around with `ssh-add -D`, `opkssh logout` (to delete the certificate) and then loading the normal key back.

PR #2566, opened 2026-02-12
Fixed transport join, moved host keys subdict, reduced tuple deep copying

WattsUp:
SSHClients do not release all their memory after close. This patch helps Python realize objects are not referenced and can be freed.

Transport `stop_thread()`: `packetizer.close()` is called which sets `packetizer.closed` to `True` and prevents the following while loop from every evaluating. I removed `packetizer.closed` clause from while loop

Transport `_handler_table`: Python got confused with a dict of function pointers and didn't automatically free them. I cleared `_handler_table` when complete.

Transport `_filter_algorithms()`: It would deep copy the algorithms every time regardless of if filtering was necessary. I added a check to return the unfiltered tuple if able.

HostKeys `lookup()`: The anonymous class `SubDict` had a reference to `hostname` which led it to be unique for each call and require unique allocation. I moved the class outside of the function call and repaired the `hostname` reference.

### DIFF
--- a/paramiko/hostkeys.py
+++ b/paramiko/hostkeys.py
@@ -32,6 +32,51 @@ from paramiko.util import get_logger, constant_time_bytes_eq, b, u
 from paramiko.ssh_exception import SSHException
 
 
+class _SubDict(MutableMapping):
+    def __init__(self, hostname, entries, hostkeys):
+        self._hostname = hostname
+        self._entries = entries
+        self._hostkeys = hostkeys
+
+    def __iter__(self):
+        for k in self.keys():
+            yield k
+
+    def __len__(self):
+        return len(self.keys())
+
+    def __delitem__(self, key):
+        for e in list(self._entries):
+            if e.key.get_name() == key:
+                self._entries.remove(e)
+                break
+        else:
+            raise KeyError(key)
+
+    def __getitem__(self, key):
+        for e in self._entries:
+            if e.key.get_name() == key:
+                return e.key
+        raise KeyError(key)
+
+    def __setitem__(self, key, val):
+        for e in self._entries:
+            if e.key is None:
+                continue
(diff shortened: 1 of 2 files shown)


WattsUp:
I ran tests of creating and closing 1000 clients and monitoring how RSS increased after each.

Here is the results of the patch 
rss-no-gc

Versus current main line
rss-main-no-gc


Here are the results of the patch with garbage collect called after each client
rss-gc

Versus current main line
rss-main-gc

Since the RSS usage with garbage collection looks the same, I'd conclude it isn't a true memory leak. But the garbage collector is having a hard time telling what can be freed after `client.close()`.

Here is the test script:

```python
import paramiko
import psutil
import os
import gc
import matplotlib.pyplot as plt


def test():
    ssh = paramiko.SSHClient()
    try:
        ssh.load_system_host_keys()
        ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
        ssh.connect(
            "192.168.1.3",
            username="root",
            password="root",
            timeout=1,
        )
    finally:
        ssh.close()


process = psutil.Process(os.getpid())

test()
gc.collect()

N = 1000

rss: list[int] = [0] * N

end = process.memory_info().rss
start = process.memory_info().rss
for n in range(1, N + 1):
    test()
    gc.collect()
    end = process.memory_info().rss
    diff = end - start
    rss[n - 1] = diff
    if (n % 10) == 0:
        print(f"RSS increased by: {diff}B for {n} clients")

plt.plot([rss[i] - rss[i - 1] for i in range(1, N)])
plt.xlabel("# of clients")
plt.ylabel("RSS (B) / client")
plt.savefig("rss-main-gc.png")
```

tgarc:
noticed the same issue with transport.stop_thread. Would be nice to see this merged.

PR #2582, opened 2026-03-17
fix: [BUG] wrong documentation on how known hosts are merged

### LINKED ISSUE #2572: [BUG] wrong documentation on how known hosts are merged

Client

known_hosts

4.0.0

3.13.12

Debian sid

_No response_

_No response_

### Expected/desired behavior

Hey.

https://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient.load_system_host_keys
and https://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient.load_host_keys
both claim that “(new replacing old if there are conflicts)”.

I've tried that out and this doesn't seem to be the case and rather only the first added key for a hostname is used and any later (and thus newer) ignored.

This happens for both, multiple entries in a single file loaded in one function call and also if in multiple files loaded in multiple function calls.

Interestingly, the order (first one wins) seem to work even of keys added via either `load_system_host_keys()` or `load_host_keys()`, so if the former loaded a wrong key for a host, an entry with the good one loaded via the latter would still loose.

OpenSSH, seems to rather check for any conflicts (within a file and over multiple files) and simply reject all keys if there are any, which I guess would be the better behaviour?

Cheers,
Chris.

### Actual behavior

see above

### How to reproduce

_No response_

### Anything else?

_No response_

JiwaniZakir:
Fixes #2572

The docstrings in `paramiko/client.py` (`load_host_keys`, `load_system_host_keys`) and `paramiko/hostkeys.py` (`HostKeys.load`) incorrectly stated that new entries replace existing ones during a merge conflict. The actual behavior preserves existing entries when a conflict is detected -- the documentation was simply wrong relative to the implementation. Updates the affected docstrings in all three locations to accurately describe the skip-on-conflict behavior. Verified by inspecting the merge logic directly against the corrected documentation.

### DIFF
--- a/paramiko/client.py
+++ b/paramiko/client.py
@@ -82,8 +82,8 @@ class SSHClient(ClosingContextManager):
         this method will not be saved back by `save_host_keys`.
 
         This method can be called multiple times.  Each new set of host keys
-        will be merged with the existing set (new replacing old if there are
-        conflicts).
+        will be merged with the existing set (existing entries are kept if
+        there are conflicts).
 
         If ``filename`` is left as ``None``, an attempt will be made to read
         keys from the user's local "known hosts" file, as used by OpenSSH,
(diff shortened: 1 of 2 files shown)


JiwaniZakir:
Closing this PR as it hasn't received maintainer review. Happy to resubmit if there's interest.

PR #2583, opened 2026-03-18
ProxyCommand.close: close pipes, reap process, and handle already-exited child

### LINKED ISSUE #2568: ProxyCommand.close() leaks file descriptors — does not close subprocess pipes or wait for process exit
## Description

`ProxyCommand.close()` in `paramiko/proxy.py` only sends `SIGTERM` to the child process but does not close the `stdin`, `stdout`, or `stderr` pipes, nor does it wait for the process to actually exit. This causes file descriptor leaks in long-running processes that create many SSH connections over time.

## Current Implementation

```python
# paramiko/proxy.py, line 121-122
def close(self):
    os.kill(self.process.pid, signal.SIGTERM)
```

The `__init__` method creates three pipe file descriptors via `subprocess.Popen`:

```python
self.process = subprocess.Popen(
    self.cmd,
    stdin=subprocess.PIPE,
    stdout=subprocess.PIPE,
    stderr=subprocess.PIPE,
    bufsize=0,
)
```

After `close()` is called, these three pipe FDs remain open and the process becomes a zombie (no `wait()`).

## Impact

In applications that repeatedly create and destroy SSH connections (e.g., test suites, monitoring tools, orchestration systems), the leaked file descriptors accumulate until the process hits the `select()` limit of 1024 (`FD_SETSIZE`), causing:

```
SSHException: Error reading SSH protocol bannerfiledescriptor out of range in select()
```

## Reproduction

```python
import gc
import os
import resource

import paramiko


def count_open_fds():
    return len(os.listdir(f"/proc/{os.getpid()}/fd"))


# Raise limit so we can accumulate past 1024 FD numbers
soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
resource.setrlimit(resource.RLIMIT_NOFILE, (4096, hard))

# Disable GC so leaked FDs are not intermittently reclaimed
gc.disable()

initial_fds = count_open_fds()
print(f"Starting FDs: {initial_fds}")

try:
    for i in range(500):
        proxy = paramiko.ProxyCommand("cat")
        client = paramiko.SSHClient()
        try:
            client.connect("localhost", sock=proxy, timeout=1, banner_timeout=1)
        except Exception as e:
            if "out of range in select" in str(e):
                current_fds = count_open_fds()
                print(f"\nIteration {i}: GOT THE BUG!")
                print(f"FDs: {current_fds} (leaked: {current_fds - initial_fds})")
                print(f"Error: {e}")
                break
        finally:
            client.close()  # Properly closing — should release all FDs

        if i % 50 == 0:
            current_fds = count_open_fds()
            print(f"Iteration {i}: FDs = {current_fds} (leaked: {current_fds - initial_fds})")

    else:
        print(f"Final FDs: {count_open_fds()} (leaked: {count_open_fds() - initial_fds})")
finally:
    gc.enable()
    resource.setrlimit(resource.RLIMIT_NOFILE, (soft, hard))
```

Run with `2>/dev/null` to suppress paramiko's stderr tracebacks:

```bash
python paramiko_fd_leak.py 2>/dev/null
```

**Output:**

```
Starting FDs: 5
Iteration 0: FDs = 8 (leaked: 3)
Iteration 50: FDs = 158 (leaked: 153)
Iteration 100: FDs = 308 (leaked: 303)
Iteration 150: FDs = 458 (leaked: 453)
Iteration 200: FDs = 608 (leaked: 603)
Iteration 250: FDs = 758 (leaked: 753)
Iteration 300: FDs = 908 (leaked: 903)

Iteration 340: GOT THE BUG!
FDs: 1028 (leaked: 1023)
Error: Error reading SSH protocol bannerfiledescriptor out of range in select()
```

3 FDs leak per iteration (one per pipe). Despite calling `client.close()` on every iteration, FDs grow linearly until `select()` fails at FD number >= 1024.

**Note:** GC is disabled in the reproduction to make the leak deterministic. With GC enabled, Python's cyclic garbage collector intermittently reclaims some leaked FDs, making the failure non-deterministic but still inevitable with enough iterations.

## Suggested Fix

```python
def close(self):
    try:
        self.process.terminate()
        self.process.wait(timeout=5)
    except subprocess.TimeoutExpired:
        self.process.kill()
        self.process.wait()
    finally:
        if self.process.stdin:
            self.process.stdin.close()
        if self.process.stdout:
            self.process.stdout.close()
        if self.process.stderr:
            self.process.stderr.close()
```

This would:
1. Gracefully terminate the process (SIGTERM), then force-kill if needed
2. Wait for process exit to prevent zombies
3. Explicitly close all pipe file descriptors

## Environment

- Paramiko version: 3.x (issue exists in all versions)
- Python: 3.14
- OS: Linux

---

*This bug was discovered and analyzed with the help of [Claude Code](https://claude.ai/code) (Anthropic's AI coding assistant).*

ReinerBRO:
Fixes #2568.

`ProxyCommand.close()` previously only sent `SIGTERM`, which could leave pipe FDs open and could also raise when the subprocess had already exited. This patch updates `close()` to:

- call `terminate()` and `wait(timeout=5)` to reap the child process
- fall back to `kill()` + `wait()` on timeout
- gracefully handle `ProcessLookupError` / `ChildProcessError`
- always close `stdin`/`stdout`/`stderr` pipes in a `finally` block

Tests added/updated in `tests/test_proxy.py`:

- close terminates, waits, and closes all pipes
- timeout path kills then waits
- already-exited process path is handled cleanly while still closing pipes
- close is idempotent

Validation:

- `python -m pytest tests/test_proxy.py -q`
- `python -m flake8 paramiko/proxy.py tests/test_proxy.py`

### DIFF
--- a/paramiko/proxy.py
+++ b/paramiko/proxy.py
@@ -19,7 +19,6 @@
 
 import os
 import shlex
-import signal
 from select import select
 import socket
 import time
@@ -119,7 +118,32 @@ class ProxyCommand(ClosingContextManager):
             raise ProxyCommandFailure(" ".join(self.cmd), e.strerror)
 
     def close(self):
-        os.kill(self.process.pid, signal.SIGTERM)
+        try:
+            self.process.terminate()
+        except ProcessLookupError:
+            pass
+        try:
+            self.process.wait(timeout=5)
+        except subprocess.TimeoutExpired:
+            try:
+                self.process.kill()
+            except ProcessLookupError:
+                pass
+            self.process.wait()
+        except ChildProcessError:
+            pass
+        finally:
+            for pipe in (
+                self.process.stdin,
+                self.process.stdout,
+                self.process.stderr,
+            ):
+                if pipe is None:
+                    continue
+                try:
+                    pipe.close()
+                except OSError:
+                    pass
 
     @property
     def closed(self):
--- a/tests/test_proxy.py
+++ b/tests/test_proxy.py
@@ -1,7 +1,7 @@
-import signal
 import socket
+import subprocess
 
-from unittest.mock import patch
+from unittest.mock import call, patch
 from pytest import raises
 
 from paramiko import ProxyCommand, ProxyCommandFailure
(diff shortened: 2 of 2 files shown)


ReinerBRO:
Thanks for the review and for calling that out. I kept the 5-second timeout fixed here to keep this PR narrowly focused on the shutdown/leak fix and avoid broadening the API surface in the same change. If maintainers would prefer configurability, I can follow up on that separately.

PR #2593, opened 2026-04-16
Modify cryptography dependency for Windows AMR64 compatibility

MugundanMCW:
PR description:

- Installing Paramiko wheels on Windows ARM64 is failing due to the absence of Cryptography wheels.
- The Cryptography release for Windows ARM64 is currently on hold due to issues with Windows ARM64 GitHub Actions[See the [PR](https://github.com/pyca/cryptography/pull/14216)].
- To resolve this, pin Cryptography to version 46.0.3, which is available for Windows ARM64 to enable successful installation of Paramiko.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,10 @@ authors = [{ name = "Jeff Forcier", email = "jeff@bitprophet.org" }]
 
 dependencies = [
   "bcrypt>=3.2",
-  "cryptography>=3.3",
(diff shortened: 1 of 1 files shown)


bskinn:
Pinning `cryptography` in paramiko's requirements is not the correct solution here.

Any user who needs a version of `cryptography` with a wheel available should install as `pip install --only-binary cryptography`. See [pip docs](https://pip.pypa.io/en/latest/cli/pip_install/#cmdoption-only-binary).

PR #2604, opened 2026-05-10
Add ML-KEM hybrid post-quantum key exchange support

alex:
Implements `mlkem768x25519-sha256` from
draft-ietf-sshm-mlkem-hybrid-kex, combining ML-KEM-768 (FIPS 203) with X25519. The shared secret is SHA-256(K_PQ || K_CL); per the draft, K is encoded as an SSH string (rather than an mpint) in both the exchange hash and the symmetric key derivation.

ML-KEM is provided by pyca/cryptography (>=48), which itself requires OpenSSL 3.5+, AWS-LC, or BoringSSL. The kex method is auto-registered when ML-KEM is importable and is preferred over the existing classical methods. Verified interoperable with OpenSSH 10.0 in both directions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,7 +84,7 @@ workflows:
               # somewhere between yaml, pip and bash all escapes get fucky
               # TODO: what is our current oldest/newest cryptography supported
               # versions? DOES that change from py3.6->3.9?
-              pip-overrides: ["cryptography==3.3.2", "cryptography==39.0.0"]
+              pip-overrides: ["cryptography==3.3.2", "cryptography==39.0.0", "cryptography==48.0.0"]
       # Kerberos tests. Currently broken :(
       #- kerberos:
       #    name: Test 3.6 w/ Kerberos support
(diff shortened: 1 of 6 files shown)


tomkins:
Not a code review (this definitely isn't my area of expertise!).

But - applied this patch to a currently installed paramiko 5.0.0 (Python 3.13, macOS 26) - and does work against a server configured for post-quantum only KexAlgorithms.

mauricioharley:
Nice work here, the K-as-string vs mpint handling and reusing the small-order X25519 check are the kind of details that are easy to get wrong in a hybrid kex port.

Following this because I'm evaluating PQC SSH options for some Python tooling built on paramiko. Any idea what's blocking review, or just waiting on maintainer time? Also curious if `sntrup761x25519-sha512@openssh.com` is on the radar too, since it covers a lot more OpenSSH servers out there than ML-KEM does right now.

alex:
a) This is just waiting for maintainer review bandwidth.
b) No, I have no plans to do sntrup, since pyca/cryptography doesn't have any plans to expose it.

alex:
(Re-created as https://github.com/paramiko/paramiko/pull/2668)

PR #2624, opened 2026-05-19
Update MD5 fingerprint calculation for FIPS compatibility

davidesba:
Use usedforsecurity=False to enable compatibility with FIPS when calculating public MD5


bskinn:
This duplicates #1928 and #2189 to one degree or another, and `usedforsecurity=False` is not the right way to handle things in `_read_private_key_pem()`. Thank you for putting this together, but I'm going to close in favor of the other PRs.

PR #2633, opened 2026-06-14
fix: default RSAKey.sign_ssh_data to rsa-sha2-256 when algorithm not in HASHES

### LINKED ISSUE #2628: [BUG] - RSAKey sign_ssh_data raises KeyError: 'ssh-rsa' with default algorithm

Client

Keys/auth

5.0.0

3.10.12

WSL on Windows 11

_No response_

_No response_

### Expected/desired behavior

RSAKey.sign_ssh_data successfully signed SSH data.

### Actual behavior

```
  File "/home/ultraplanet/.local/lib/python3.10/site-packages/paramiko/rsakey.py", line 142, in sign_ssh_data
    algorithm=self.HASHES[algorithm](),
KeyError: 'ssh-rsa'
```

### How to reproduce

key = RSAKey.generate(1024)
msg = key.sign_ssh_data(b"jerri blank")

### Anything else?

'ssh-rsa' was removed in 5.0.0, I guess we should use 'rsa-sha2-256' by default. It would be better to add a check for the algorithm parameter.

kaizeenn:
## Summary

Fixes #2628.

In paramiko 5.0, `'ssh-rsa'` (SHA-1) was removed from `RSAKey.HASHES` because SHA-1 signing is no longer acceptable. However `sign_ssh_data()` still defaulted to `self.name` (`'ssh-rsa'`) when `algorithm=None`, causing an immediate `KeyError` for any caller that didn't pass an explicit algorithm.

```python
key = RSAKey.generate(1024)
key.sign_ssh_data(b"data")  # KeyError: 'ssh-rsa'
```

The same crash occurs when a caller passes `algorithm='ssh-rsa'` explicitly (e.g. legacy code).

## Fix

When the resolved algorithm is absent from `HASHES`, fall back to `'rsa-sha2-256'` — the recommended default per RFC 8332 and the algorithm paramiko's own transport layer negotiates first.

```python
if algorithm is None or algorithm not in self.HASHES:
    algorithm = "rsa-sha2-256"
```

## Testing

- All 47 RSA-related tests pass (`pytest tests/ -k rsa`)
- Smoke-tested `algorithm=None`, `algorithm='ssh-rsa'`, and `algorithm='rsa-sha2-512'` manually

### DIFF
--- a/paramiko/rsakey.py
+++ b/paramiko/rsakey.py
@@ -167,9 +167,7 @@ class RSAKey(PKey):
             sign = b"\x00" * ((diff + 7) // 8) + sign
 
         try:
-            key.verify(
-                sign, data, padding.PKCS1v15(), self.HASHES[sig_algorithm]()
-            )
+            key.verify(sign, data, padding.PKCS1v15(), self.HASHES[sig_algorithm]())
         except InvalidSignature:
             return False
         else:
--- a/paramiko/rsakey.py
+++ b/paramiko/rsakey.py
@@ -132,8 +132,12 @@ class RSAKey(PKey):
         return isinstance(self.key, rsa.RSAPrivateKey)
 
     def sign_ssh_data(self, data, algorithm=None):
-        if algorithm is None:
-            algorithm = self.name
+        if algorithm is None or algorithm not in self.HASHES:
+            # ssh-rsa (SHA1) was removed from HASHES in 5.0; fall back to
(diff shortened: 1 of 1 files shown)


kaizeenn:
Hi, I've formatted the code with black to resolve the style check failure. Locally, all tests are passing successfully. Ready for a new CI run!

PR #2634, opened 2026-06-18
fix: skip Match blocks with unknown criteria

### LINKED ISSUE #2632: [BUG] - Configuration under Match statements with invalid or unknown Match criteria words is

Client

sshconfig

5.0.0

3.14.5

Tested on Ubuntu 24.04 and CachyOS

Not relevant (bug is before connection)

_No response_

### Expected/desired behavior

If paramiko doesn't understand a Match criteria it should not read its configuration (and probably even error or at least warn) 

### Actual behavior

Paramiko still reads the configuration, even though it can't verify if the Match statement actually matches

### How to reproduce

Config
```sshconfig
Match localnetwork 9.8.7.6/32
    ProxyJump THISSHOULDNOTMATCH
```

Test program
```python
import paramiko
print(paramiko.SSHConfig.from_path("/tmp/poc/config").lookup("1.2.3.4")["proxyjump"])
```

### Anything else?

The wrong logic is in the function `_does_match` in `paramiko/config.py`

I found this error because I had a localnetwork match in the config, which is a valid keyword in OpenSSH, but not handled by paramiko.

decembercomposer697-hue:
Fixes #2632: when an SSH config Match directive uses an unknown
criteria keyword (e.g. localnetwork), paramiko would silently apply
the block instead of skipping it.

### DIFF
--- a/paramiko/config.py
+++ b/paramiko/config.py
@@ -362,6 +362,17 @@ class SSHConfig:
             configured_host = options.get("hostname", None)
             configured_user = options.get("user", None)
             type_, param = candidate["type"], candidate["param"]
+            # If the Match criteria is unknown (e.g. localnetwork), skip this
+            # block rather than silently applying it (paramiko#2632).
+            if self._unknown_match_criteria(type_):
+                import warnings
+                warnings.warn(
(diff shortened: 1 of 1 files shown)


wahajahmed010:
test

PR #2651, opened 2026-07-16
Reject oversized BPP packet lengths in Packetizer

### LINKED ISSUE #2626: Missing MAX_PACKET_SIZE check in BPP layer allows memory exhaustion from malicious server
## Summary
The `read_message()` function in `packet.py` does not validate 
`packet_size` against a maximum value before calling `read_all()`.

A malicious SSH server can send a BPP packet with `packet_size = 0xFFFFFFFF`
(4GB), causing the client to attempt reading 4GB from the socket,
resulting in indefinite blocking or memory exhaustion.

## Location
`paramiko/packet.py` — `read_message()`:

```python
packet_size = struct.unpack(">I", header[:4])[0]
# No maximum size check here
buf = self.read_all(packet_size + self.__mac_size_in - len(leftover))
```

## Suggested Fix
Add a check before allocating:

```python
MAX_BPP_PACKET_SIZE = 35000  # OpenSSH default
if packet_size > MAX_BPP_PACKET_SIZE:
    raise SSHException(f"Packet too large: {packet_size}")
```

## Impact
- Prerequisite: connection to a malicious SSH server
- Effect: client process blocks indefinitely or exhausts memory
- Severity: Low/Medium hardening suggestion

## Environment
- paramiko 2.11.0
- Python 3.x

pctablet505:
## Summary

`Packetizer.read_message()` trusts the 4-byte packet length pulled directly out of the incoming (still-encrypted, in most cipher modes) BPP header, and uses it as-is to size subsequent socket reads. A malicious or misbehaving server can send a length close to `0xFFFFFFFF`, causing the client to block for a very long time (or attempt a huge allocation) trying to read that many bytes from the socket.

This adds a bounds check applied everywhere `packet_size` is derived from the header: the encrypt-then-mac path, the AEAD/GCM path, and the plain/CBC path. The limit (35000 bytes) matches the floor recommended by RFC 4253 section 6.1 and the value OpenSSH itself enforces.

## Testing

Added `test_read_rejects_oversized_packet` in `tests/test_packetizer.py`, which sends a crafted header claiming an oversized (but still block-aligned) packet length and asserts `read_message()` raises `SSHException` instead of trying to read that much data. Without the fix, this exact input causes `read_message()` to hang waiting on the socket, which I confirmed locally.

Ran the full `test_packetizer.py`, `test_transport.py`, and `test_kex.py` suites; all pass.

Closes #2626

### DIFF
--- a/paramiko/packet.py
+++ b/paramiko/packet.py
@@ -74,13 +74,17 @@ class Packetizer:
 
     # Maximum size (in bytes) accepted for an incoming BPP (binary packet
     # protocol) packet, prior to decryption/decompression. RFC 4253, Section
-    # 6.1 says implementations should support packets of at least 35000
-    # bytes and may reject larger ones; OpenSSH uses the same value. This
-    # guards against malicious/corrupt peers advertising an absurd packet
-    # length (e.g. via a spoofed/garbled length header) and forcing us to
-    # allocate huge buffers or block for a long time waiting for that many
-    # bytes to arrive.
-    MAX_PACKET_SIZE = 35000
+    # 6.1 only sets a MINIMUM that implementations must be able to accept
+    # (>= 35000 bytes) -- it is not a ceiling, and legitimate channels may
+    # negotiate a max_packet_size well above that (see
+    # `.Transport._sanitize_packet_size`, which allows values up to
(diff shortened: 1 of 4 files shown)


FloorIsGround (approved):
very clean and concise implementation, looks good to me 👍

bskinn:

PR #2660, opened 2026-07-24
Fix SFTPFile.seek() to return the new absolute position

### LINKED ISSUE #2218: SFTPFile.seek should return the new position
Hi,

in order to match the behaviour of standard file objects `SFTPClient.seek` should return the new position, currently `None` is returned.

agu2347:
Fixes #2218.

`seek()` updated `self._pos`/`self._realpos` but didn't return anything, implicitly returning `None`. Standard Python file objects (per `io.IOBase.seek()`'s documented contract) return the new absolute position from `seek()`, and `SFTPFile`'s own `tell()` already does exactly that (`return self._pos`), so this brings `seek()` in line with both the general Python file-object convention and this class's own sibling method.

This is a purely additive change: any existing caller not using the return value (the overwhelming majority, since it was always `None`) is completely unaffected; callers wanting the new position previously had to make a separate `tell()` call immediately after `seek()`.

**Testing**: verified against a real, live in-process SFTP server (this project's own test fixture, not a mock): confirmed `seek()` now returns the correct new position for `SEEK_SET`, `SEEK_CUR`, and `SEEK_END` modes (and the default whence), matching `tell()` called immediately afterward in each case.

Added a regression test exercising all three whence modes plus the default-whence call against a real remote file. I confirmed the test fails with the original code (returns `None`) and passes with the fix. Ran the full existing `test_sftp.py` suite (35 passed: 34 baseline + 1 new, 4 skipped for unrelated reasons), no regressions.

### DIFF
--- a/paramiko/sftp_file.py
+++ b/paramiko/sftp_file.py
@@ -260,6 +260,10 @@ class SFTPFile(BufferedFile):
         Set the file's current position.
 
         See `file.seek` for details.
+
+        :returns:
+            the new absolute position, in bytes, matching the standard
+            Python file-object convention (as `io.IOBase.seek` does).
         """
         self.flush()
         if whence == self.SEEK_SET:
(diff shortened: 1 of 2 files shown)


bskinn:

Issue #2050, opened 2022-05-12
Writing in append mode fails on TypeError when stat does not return st_size


UOndro:
Writing in append mode fails on the following error:

`TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'`

The problem is that [_get_size](https://github.com/paramiko/paramiko/blob/main/paramiko/sftp_file.py#L519) method in SFTPFile can return None, which is then used to assign value to [_realpos](https://github.com/paramiko/paramiko/blob/29c4d3210b745e7d56c328ffebc2513ebfb482fb/paramiko/file.py#L508).

The [_get_size](https://github.com/paramiko/paramiko/blob/main/paramiko/sftp_file.py#L519) method can return None when [stat](https://github.com/paramiko/paramiko/blob/29c4d3210b745e7d56c328ffebc2513ebfb482fb/paramiko/sftp_attr.py#L100) does not return st_size in response. 

Stack trace:
```  
File "/usr/local/lib/python3.8/site-packages/paramiko/file.py", line 405, in write
    self._write_all(data)
  File "/usr/local/lib/python3.8/site-packages/paramiko/file.py", line 522, in _write_all
    count = self._write(data)
  File "/usr/local/lib/python3.8/site-packages/paramiko/sftp_file.py", line 199, in _write
    long(self._realpos),
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

bskinn:
Do you know, what should the correct behavior be when `_get_size()` returns `None`? Can it successfully continue the append-write? Or should paramiko surface this as an error/exception condition?

Which Python and paramiko versions are you using here?

jun66j5:
Related issue: [#1980 (comment)](https://github.com/paramiko/paramiko/issues/1980#issuecomment-1029310765)

Are you using what ssh server?

UOndro:
> Do you know, what should the correct behavior be when _get_size() returns None? Can it successfully continue the append-write? Or should paramiko surface this as an error/exception condition?

Not sure. However, the code already handles if an exception is raised during calling stat(). If the exception is raised, paramiko defaults to 0. So, I think when st_size is None,  `_get_size()` should also default to 0. 

> Which Python and paramiko versions are you using 

We use paramiko 2.7.2(however, I check also a new version and this bug is still there) and we use python 3.8.


> Are you using what ssh server?
It's the client's server, so I don't know, but I asked them if they could provide this information. 

[...]

PR #2670, opened 2026-08-15
SFTPFile.seek(SEEK_END) should propagate stat errors, not swallow them

### LINKED ISSUE #2460: [BUG] - A SFTPFile.seek can cause SFTPFile.tell to lie

Client

SFTP

3.3.0

3.11.5

Debian bullseye

Not 100% sure but banner report "Maverick_SSHD", so maybe something from https://jadaptive.com/

_No response_

### Expected/desired behavior

Start with a poorly-behaved "Maverick" SFTP server that has a `test_file.txt` that is non-empty.  Say, for example, it's 12345 bytes long.

```python
import os
import paramiko

client = paramiko.SSHClient()
client.connect(hostname="example.com", username="foo", password="bar")
sftp = client.open_sftp()
fh = sftp.open("test_file.txt", "rb")
fh.seek(0, os.SEEK_END)
print(fh.tell())
```

Expected: prints 12345

### Actual behavior

Actual: Prints 0 (zero)

### How to reproduce

See above bit about expected behavior, but I'm guessing the hardest part will be getting an SFTP server to reproduce this.

### Anything else?

If you call `SFTPFile.stat` directly using an open file from this particular SFTP server, you'll see that you get

```
OSError: The message [test_file.txt] is not extractable!
```

(This error message makes no sense to me, BTW.)

So, CMD_FSTAT raises an error.  FYI calling `SFTPClient.stat` (which does CMD_STAT instead) does return a proper result, including the file size.

IMHO, I think the problem is that [`stat` with `SEEK_END` tries to update `_realpos` using `self._get_size()`](https://github.com/paramiko/paramiko/blob/23f92003898b060df0e2b8b1d889455264e63a3e/paramiko/sftp_file.py#L271), but [`SFTPFile._get_size` swallows exceptions and returns 0](https://github.com/paramiko/paramiko/blob/23f92003898b060df0e2b8b1d889455264e63a3e/paramiko/sftp_file.py#L531-L535).  If you seek to the end then you just set `_realpos` to 0.

Obviously I wish this SFTP server would make CMD_FSTAT work.  In the absence of a working server, though, I'd rather get an exception from Paramiko, not incorrect information.  If you, for example, pass an `SFTPFile` to boto3's [`S3Client.upload_fileobj`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/upload_fileobj.html), in order to transfer a file from SFTP to S3, boto3 will silently truncate the file after it determines the file is 0 bytes long according to `SFTPFile.tell`.

My first suggestion is that `_get_size` shouldn't eat exceptions, but after reading #2050, does it _need_ to eat exceptions?  I hope not!

If `_get_size` can't raise exceptions, then maybe Paramiko should do some kind of test before `SFTPFile.seekable` returns true, to find out whether CMD_FSTAT works?

Thank you to all the authors and maintainers of Paramiko!

agu2347:
### Problem

`SFTPFile.seek(offset, SFTPFile.SEEK_END)` computes the new position via `self._get_size()`:

```python
def _get_size(self):
    try:
        return self.stat().st_size
    except:
        return 0
```

`_get_size()` swallows *any* exception from `stat()` (i.e. `CMD_FSTAT`) and returns `0`. If the server's `CMD_FSTAT` implementation is broken or unsupported for a given file (as reported against a "Maverick" SFTP server), `seek(0, SEEK_END)` silently sets the position to `0` instead of raising - so `tell()` afterwards reports `0` instead of the real file size, with no indication anything went wrong.

This isn't just cosmetic: reporter noted that passing such an `SFTPFile` to `boto3`'s `upload_fileobj()` causes boto3 to determine (via `tell()`) that the file is empty, silently truncating the "transferred" file to 0 bytes.

Closes #2460

### Fix

`SFTPFile.seek()`'s `SEEK_END` branch now calls `self.stat().st_size` directly instead of `self._get_size()`, so a `stat()`/`CMD_FSTAT` failure propagates as an exception instead of being silently treated as "file is 0 bytes".

`_get_size()` itself is left unchanged, since it's also used by `BufferedFile._set_mode()` for append-mode opens (see #2050), which already depends on it defaulting to `0` on failure - changing that call site's behavior is a separate, more involved question I didn't want to bundle into this fix.

### Testing

- Added `test_seek_end_propagates_stat_errors`, which opens a real file against the test suite's stub SFTP server, confirms normal `seek(0, SEEK_END)` still works, then mocks `stat()` to raise and asserts `seek(0, SEEK_END)` now raises instead of silently resetting the position to 0.
- Confirmed the new test fails with `Failed: DID NOT RAISE <class 'OSError'>` when the fix is reverted (stashed `sftp_file.py` only, test unchanged).
- Full `tests/test_sftp.py` suite: 35 passed, 4 skipped (pre-existing skips, unrelated to this change).
- `black --line-length 79` (pinned project version 22.12.0) and `flake8` both clean on the changed files.

### DIFF
--- a/paramiko/sftp_file.py
+++ b/paramiko/sftp_file.py
@@ -268,7 +268,14 @@ class SFTPFile(BufferedFile):
             self._pos += offset
             self._realpos = self._pos
         else:
-            self._realpos = self._pos = self._get_size() + offset
+            # Intentionally use stat() (which propagates errors) instead of
+            # _get_size() (which silently swallows them and returns 0) here.
+            # Silently treating an unknown/unreadable size as 0 would make
+            # SEEK_END seek to position 0 without any indication that
+            # anything went wrong, which can cause silent data loss/
+            # truncation for callers relying on tell() afterwards.
+            # See https://github.com/paramiko/paramiko/issues/2460.
+            self._realpos = self._pos = self.stat().st_size + offset
         self._rbuffer = bytes()
 
     def stat(self):
--- a/tests/test_sftp.py
+++ b/tests/test_sftp.py
@@ -30,6 +30,7 @@ import warnings
 from binascii import hexlify
 from io import StringIO
 from tempfile import mkstemp
+from unittest import mock
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


bskinn:

PR #2671, opened 2026-08-15
docs: clarify PKey.sign_ssh_data return types

### LINKED ISSUE #2539: [BUG] - Return value for sign_ssh_data in PKey is incorrect in subclasses

Client

Keys/auth

3.5.1

3.12.11

MacOS, Linux

_No response_

_No response_

### Expected/desired behavior

The documentation and typing information for PKey.sign_ssh_data is not correct for all subclasses.

The documentation states that the return type is:

> Returns
an SSH signature [message](https://docs.paramiko.org/en/stable/api/message.html#paramiko.message.Message).

https://docs.paramiko.org/en/stable/api/keys.html#paramiko.pkey.PKey.sign_ssh_data

However, for some subclasses of PKey, this is not correct. For instance, AgentKey, the return type is `bytes`.

https://docs.paramiko.org/en/stable/api/agent.html?highlight=agentkey#paramiko.agent.AgentKey

https://github.com/paramiko/paramiko/blob/d9ab89a0f8ae37a25d44565d5eb03a5d93fed5b9/paramiko/agent.py#L484

### Actual behavior

Actual behavior is that depending on the key that does the signing, you might need to handle both Message and bytes return values.

### How to reproduce

_No response_

### Anything else?

This is probably already known and handled by users of this package, but trying to build a generic signing part of my application. I ran into this issue when I expected things to be a Message but in fact there were two types other than handle here.

Most likely the fix here is to just update the documentation and return types for types-paramiko.

nightcityblade:
Fixes #2539

`AgentKey.sign_ssh_data()` returns an encoded byte string, while locally held key implementations return a `Message`. Clarify the generic `PKey` API documentation and its return type so the inherited AgentKey documentation no longer promises only `Message`.

Tests:
- `uv run flake8 paramiko/pkey.py`
- `uv run codespell paramiko/pkey.py`
- `uv run pytest tests/test_pkey.py -q` (39 passed, 1 skipped)
- `uv run sphinx-build -E -W -b html sites/docs sites/docs/_build`

### DIFF
--- a/paramiko/pkey.py
+++ b/paramiko/pkey.py
@@ -416,15 +416,18 @@ class PKey:
 
     def sign_ssh_data(self, data, algorithm=None):
         """
-        Sign a blob of data with this private key, and return a `.Message`
-        representing an SSH signature message.
+        Sign a blob of data with this private key, and return an SSH signature
+        message.
 
         :param bytes data:
             the data to sign.
         :param str algorithm:
             the signature algorithm to use, if different from the key's
             internal name. Default: ``None``.
-        :return: an SSH signature `message <.Message>`.
+        :rtype: `.Message` or bytes
+        :return:
+            an SSH signature `message <.Message>`, or the encoded message as
+            ``bytes`` for agent-backed keys.
 
         .. versionchanged:: 2.9
             Added the ``algorithm`` kwarg.


nightcityblade:
CI note: the only failing job is `Lint`, and current `main` at d60d5c1 has the same failing CircleCI workflow. All other PR jobs passed, including Docs, Style check, Spellcheck, and every test matrix job. The changed file also passes `flake8 paramiko/pkey.py` locally, so I have kept this documentation PR focused rather than including an unrelated baseline lint fix.

bskinn:

Issue #2550, opened 2025-10-24
[SUPPORT] - Failed to import in latest python 3.14


ruiyang2015:
### Are you using paramiko as a client or server?

Client

### What feature(s) aren't working right?

Something else

### What version(s) of paramiko are you using?

4.0.0

### What version(s) of Python are you using?

3.14

### What operating system and version are you using?

MacOS

### If you're connecting as a client, which SSH server are you connecting to?

SFTP

### What are you trying to do with paramiko?

when import paramiko in python 3.14, getting following import error:
```
Traceback (most recent call last):
  File "<python-input-0>", line 1, in <module>
    import paramiko
  File "/Users/ruiyang/Go/src/github.com/ascend-io/ascend-core/.venv/lib/python3.14/site-packages/paramiko/__init__.py", line 24, in <module>
    from paramiko.transport import (
    ...<3 lines>...
    )
  File "/Users/ruiyang/Go/src/github.com/ascend-io/ascend-core/.venv/lib/python3.14/site-packages/paramiko/transport.py", line 98, in <module>
    from paramiko.ed25519key import Ed25519Key
  File "/Users/ruiyang/Go/src/github.com/ascend-io/ascend-core/.venv/lib/python3.14/site-packages/paramiko/ed25519key.py", line 22, in <module>
    import nacl.signing
  File "/Users/ruiyang/Go/src/github.com/ascend-io/ascend-core/.venv/lib/python3.14/site-packages/nacl/signing.py", line 16, in <module>
    import nacl.bindings
  File "/Users/ruiyang/Go/src/github.com/ascend-io/ascend-core/.venv/lib/python3.14/site-packages/nacl/bindings/__init__.py", line 186, in <module>
    from nacl.bindings.crypto_secretstream import (
    ...<15 lines>...
    )
  File "/Users/ruiyang/Go/src/github.com/ascend-io/ascend-core/.venv/lib/python3.14/site-packages/nacl/bindings/crypto_secretstream.py", line 14, in <module>
    from typing import ByteString, Optional, Tuple, cast
ImportError: cannot import name 'ByteString' from 'typing' (/Users/ruiyang/.local/share/uv/python/cpython-3.14.0b1-macos-aarch64-none/lib/python3.14/typing.py)
```

### How are you trying to do it, and what's happening instead?

i am trying to upgrade to latest stable python release 3.14 and facing import error for the paramiko package

bskinn:
[...]

Issue #2527, opened 2025-07-29
State of removing PyNaCl?


bseidenberg:
Hi!

We are very interested in removing the dependency on PyNaCl from Paramiko. This is currently limited to paramiko/ed25519key.py. I was working on preparing a patch when I discovered that there was existing work for this. Both https://github.com/paramiko/paramiko/pull/1806 and https://github.com/paramiko/paramiko/pull/2103 have been pending for some time. It looks like they were blocked waiting for a dependency version bump that happened with Paramiko 3.0. 

What would it take to realize this? Given our interest, I am able to dedicate some bandwidth to working on this. Would refreshing [PR 1806](https://github.com/paramiko/paramiko/pull/1806) be the best path forward?

Thanks!
Benjamin

crlorentzen:
Following up on this issue. 

We continue to be interested in removing PyNaCl from Paramiko. We believe this is worthwhile reducing dependencies and deployment size. We are looking for agreement on the path forward.

crlorentzen:
@bitprophet With paramiko 5.0 out, we wanted to check, is there appetite for removing PyNaCl in favor of the existing cryptography dependency? 

We see requests since 2019 (#1419) and working PRs since 2021 (#1806, #2103). In early 2023 there appeared to be appetite post-3.0, is anything still blocking this? 

We're happy to refresh PR(s) if you would be open to accepting PyNaCl's removal.

PR #2679, opened 2026-08-26
Use Cryptography instead of PyNaCl for Ed25519 keys

toabctl:
`Ed25519Key` is the only thing left in Paramiko that uses PyNaCl, so this moves
it over to Cryptography and drops `pynacl` from the requirements.

Ed25519 support landed in 2.2 (#325 via #972) on PyNaCl because Cryptography
had no Ed25519 primitive at the time. It got one in 2.6, and our floor is
already 3.3, so it has been available to us the whole time. Same approach as
@adiroiban's #1806 and @boyxuper's #2103, both of which have gone stale against
main — credit for it is theirs. Refs #1419 and #2527.

Besides shedding a dependency that generates a steady trickle of install and
import reports (#1265, #1435, most recently #2550), the reason I care about
this: Paramiko already lets OpenSSL decide what is usable, e.g.
`KexCurve25519.is_available()` drops curve25519 when OpenSSL refuses it.
Ed25519 is the one algorithm that cannot take part, because libsodium does not
know the OpenSSL configuration exists. On a host with a restricted algorithm
policy, OpenSSL will happily do Ed25519 — its FIPS provider has registered it
`fips=yes` since 3.4.0 — while libsodium simply is not in the picture. This
does not make Paramiko work under such a policy by itself, mind: `hashlib.md5()`
in `pkey.py` is still there, which is #396 and what #1928 / #2189 are about.
Nothing here overlaps those in `paramiko/`.

Three things that are not obvious from the diff:

- I used `public_bytes(Encoding.Raw, PublicFormat.Raw)` rather than the nicer
  `public_bytes_raw()`, which only exists in Cryptography 40+; the floor cell is
  still 3.3.2.
- `_fields` now holds raw bytes rather than the key object. Cryptography's key
  objects compare by identity on 3.3.2 and 39.0.0, and are unhashable on recent
  versions — and `PKey.__eq__`/`__hash__` are both built on `_fields`. Swap the
  bytes back for the object and `test_ed25519_compare` fails either way, once I
  added a `hash()` assertion to it; without that assertion only the equality
  half was caught.
- Cryptography's `verify()` takes `(signature, data)`, the other way round from
  PyNaCl's.

The one API change is `Ed25519Key.verifying_key`, which now returns an
`Ed25519PublicKey` instead of a `nacl.signing.VerifyKey`; there is a warning on
the changelog entry. I filed that as `:support:` — happy to move it if you would
rather it waited for a major.

Wire format is unchanged, and as Ed25519 is deterministic the new test pins the
exact signature the PyNaCl code produced for the existing fixture.

I diffed behaviour against main across certificates, `PKey.from_path`,
`from_type_string`, deepcopy, equality/hashing/set membership, fingerprints,
`get_bits`, `identifiers`, malformed key blobs and partly-initialised objects.
All identical, including error types. Three things do differ:

- `pickle.dumps()` on an `Ed25519Key` now raises `TypeError`, as Cryptography's
  key objects are not picklable. Worth knowing, though `RSAKey` and `ECDSAKey`
  are already unpicklable on main for exactly this reason — Ed25519Key was the
  only one that worked, by accident of PyNaCl. `copy.deepcopy()` still works.
- `verify_ssh_sig()` with a signature that is not 64 bytes used to raise
  `ValueError` out of a remote-supplied value; it now returns `False`. That
  looks like an improvement, but it is a change.
- On a host whose OpenSSL refuses Ed25519 — a FIPS provider older than 3.4.0,
  or OpenSSL below 1.1.1b — loading a key or parsing a peer's host key now
  raises `UnsupportedAlgorithm` where PyNaCl was simply unaffected by the
  policy. Obeying the policy is the point, but the failure is not graceful:
  `import paramiko` is fine, however `ssh-ed25519` stays in `_preferred_keys`,
  since there is no availability gate for host key algorithms like the one
  `KexCurve25519` has. Happy to add one if you want it here rather than
  separately.

Tested with `pynacl` uninstalled: the full suite on 3.13, which leaves exactly
the same 17 failures as unpatched 142f593e (the `AgentKey.public_blob` ones,
#2462), and `test_pkey.py`, `pkey.py`, `test_hostkeys.py` on 3.9 against all
three of your cryptography cells — 3.3.2, 39.0.0 and the newly added 48.0.0.
Worth noting 48.0.0 makes the `_fields` change load-bearing rather than
theoretical: `Ed25519PublicKey` is unhashable there, so the object form would
raise.
`test_ed25519_32bit_collision` is skipped off 32-bit, so I checked by hand that
the two badhash fixtures still differ in `_fields`.

Rebased onto 142f593e. This previously carried a second commit fixing the E501
in `tests/pkey.py`; you fixed that line yourself with a `# noqa` in the
meantime, so I dropped mine — this is now a single commit touching five files,
none of which you've changed upstream.

### DIFF
--- a/paramiko/ed25519key.py
+++ b/paramiko/ed25519key.py
@@ -17,14 +17,32 @@
 from typing import Union
 
 import bcrypt
-import nacl.signing
+from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric.ed25519 import (
+    Ed25519PrivateKey,
+    Ed25519PublicKey,
+)
 from cryptography.hazmat.primitives.ciphers import Cipher
 
 from paramiko.message import Message
 from paramiko.pkey import OPENSSH_AUTH_MAGIC, PKey, _unpad_openssh
 from paramiko.ssh_exception import PasswordRequiredException, SSHException
 from paramiko.util import b
+
+
+def _public_bytes(key: Ed25519PublicKey) -> bytes:
+    """
+    Return the 32 raw bytes of ``key``, as used in the SSH ``ssh-ed25519``
+    key blob.
+
+    Cryptography's own ``public_bytes_raw`` is tidier, but only exists as of
+    version 40; this spelling works on every version we support.
+    """
+    return key.public_bytes(
+        serialization.Encoding.Raw, serialization.PublicFormat.Raw
+    )
 
 
 class Ed25519Key(PKey):
@@ -54,7 +72,9 @@ class Ed25519Key(PKey):
                 key_type=self.name,
                 cert_type="ssh-ed25519-cert-v01@openssh.com",
             )
-            self._verifying_key = nacl.signing.VerifyKey(msg.get_binary())
+            self._verifying_key = Ed25519PublicKey.from_public_bytes(
+                msg.get_binary()
+            )
         elif filename is not None:
             with open(filename, "r") as f:
                 pkformat, data = self._read_private_key("OPENSSH", f)
@@ -147,10 +167,10 @@ class Ed25519Key(PKey):
             key_data = message.get_binary()
             # The second half of the key data is yet another copy of the public
             # key...
-            signing_key = nacl.signing.SigningKey(key_data[:32])
+            signing_key = Ed25519PrivateKey.from_private_bytes(key_data[:32])
             # Verify that all the public keys are the same...
             assert (
-                signing_key.verify_key.encode()
+                _public_bytes(signing_key.public_key())
                 == public
                 == public_keys[i]
                 == key_data[32:]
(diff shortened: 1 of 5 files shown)


toabctl:
@bitprophet is this something you would accept/merge ?

bskinn:

Closing this PR due to inferred AI authorship. My apologies for adding the `Ready for review` and then walking back; I labeled before I took a close look.

toabctl:
@bskinn if I do this now manually, would there be a chance to get this accepted? Including the technical direction?

adiroiban:
I am 100% onboard with rejected LLM contributions.

The original contribution from 2021 was not LLM generated.

@bskinn if you have time and want to review and merge that work, I am happy to resume working on it.

I am using paramiko only to support some automated testing...so I created a fork that works well for me since 2021 ... so for me it's not a big blocker if this is not resolved.
But I am happy to help so that I don't have to keep my patching.

Thanks

PR #2681, opened 2026-08-31
Validate default_window_size and default_max_packet_size in Transport

### LINKED ISSUE #2545: [BUG] - Validate ftpPort in paramiko.Transport(ftpHost, ftpPort)

Client

SFTP

3.5.1

3.10.8

Windows 11

_No response_

_No response_

### Expected/desired behavior

That a `.Transport()` initialize without error.  But it should reports errors if it is passed bad values.

### Actual behavior

Consider this working code `paramiko.Transport('myURL', 21)` If 21 is instead '21' the transport fails at CONSUMPTION as `paramiko.SFTPClient.from_transport(transport)` claiming "TypeError: '<' not supported between instances of 'int' and 'str'"

1) port value should be validated as initiated in `.Transport()`; not at consumption.
2) '22' (a string) is more type accurate for its intention, vs. an int, because this value is concatenated into a string as part of a URL.

### How to reproduce

`paramiko.Transport('myURL', '21')`

### Anything else?

The IntelliSense & online signature of .Transport() is nothing like every example out there.  Common uses are `.Transport(stringHOST, intPORT)`.  The definition is parameter#1=Socket(Sock) & parameter#2=default_window_size(int).

afonsojanu:
Closes #2545.

Transport's constructor takes default_window_size and default_max_packet_size, but never checked that either was actually an int. Pass a string (which is easy to do by accident, since Transport's signature is sock plus these kwargs, not a two-argument host/port constructor the way it reads at first glance) and it gets stored just fine. The failure only shows up later, inside clamp_value, when it tries min(your_value, MAX_WINDOW_SIZE) and hits a TypeError comparing str against int, deep inside channel or SFTP setup rather than at the call that actually caused it.

That's the exact scenario in the linked issue: someone called Transport('host', '21') expecting host/port semantics, and instead got a confusing crash from SFTPClient.from_transport much later.

Both values now get a type check right where they're assigned, so a bad type raises TypeError immediately with a message naming which argument was wrong, instead of surfacing wherever the value first gets compared against something.

Added two tests alongside the existing test_sanitze_window_size/test_sanitze_packet_size, confirmed they fail without the fix (constructor happily accepts the string, and I reproduced the original clamp_value TypeError manually to make sure the new check catches the same case) and pass with it. Ran the full transport test suite, 99 passed / 2 skipped, same skips as before this change. flake8 clean on both files.

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -513,6 +513,14 @@ class Transport(threading.Thread, ClosingContextManager):
         self.channel_events = {}  # (id -> Event)
         self.channels_seen = {}  # (id -> True)
         self._channel_counter = 0
+        for arg_name, value in (
+            ("default_window_size", default_window_size),
+            ("default_max_packet_size", default_max_packet_size),
+        ):
+            if not isinstance(value, int):
+                raise TypeError(
+                    "{} must be an int, got {!r}".format(arg_name, value)
+                )
         self.default_max_packet_size = default_max_packet_size
         self.default_window_size = default_window_size
         self._forward_agent_handler = None
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -519,7 +519,7 @@ class Transport(threading.Thread, ClosingContextManager):
         ):
             if not isinstance(value, int):
                 raise TypeError(
-                    "{} must be an int, got {!r}".format(arg_name, value)
+                    f"{arg_name} must be an int, got {value!r}"
                 )
         self.default_max_packet_size = default_max_packet_size
         self.default_window_size = default_window_size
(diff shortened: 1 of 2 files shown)


paramiko/transport.py
```diff
@@ -513,6 +513,14 @@ def __init__(
         self.channel_events = {}  # (id -> Event)
         self.channels_seen = {}  # (id -> True)
         self._channel_counter = 0
+        for arg_name, value in (
+            ("default_window_size", default_window_size),
+            ("default_max_packet_size", default_max_packet_size),
+        ):
+            if not isinstance(value, int):
+                raise TypeError(
+                    "{} must be an int, got {!r}".format(arg_name, value)
```
bskinn: Any particular reason you chose `.format()` over an f-string?

afonsojanu:
No particular reason, just habit from matching the surrounding code before I looked closer. Switched it to an f-string.

Issue #2658, opened 2026-07-20
[FEAT] - make BufferedFile io compatible


asmodehn:
### Is this feature for paramiko acting as a client or a server?

Client

### What functionality does this feature request relate to?

SSH

### If you're using paramiko as part of another tool, which tool/version?

OpenWisp

### Desired behavior

paramiko implement [BufferedFile](https://github.com/paramiko/paramiko/blob/main/paramiko/file.py#L31), and this serves as the base of [ChannelFile](https://github.com/paramiko/paramiko/blob/main/paramiko/channel.py#L1334), which is the interface through which a client see the output/error of a command that was sent to the server.

There are missing functionalities here, and in recent python versions [io](https://docs.python.org/3/library/io.html#) specifies what a file should provide as interface. BufferedFile only support parts of that.

### Anything else?

In my specific usecase, I want the client to retrieve the various output lines before a specific timeout, or before the command terminates (without actually breaking the connection, so it s not detected as exit)...
As far as I understand at this stage, this requires [read1](https://docs.python.org/3/library/io.html#io.BufferedIOBase.read1) to be implemented to be able to loop on it until a potential client side timeout.

There are probably quite a few other usecases that can be enabled with this as well. I ended up there after a few frustrations trying to deal with [get_exit_status like other users](https://github.com/paramiko/paramiko/issues?q=is%3Aissue%20state%3Aopen%20exit_status)

PR #2682, opened 2026-09-01
Fix duplicate Transport auth docs (#2556)

### LINKED ISSUE #2556: [BUG] - documentation: copy pasted doc appearing twice in Transport section?

Client

Something else

stable (according to https://docs.paramiko.org/en/stable/api/transport.html)

not relevant

not relevant

not relevant

not relevant

### Expected/desired behavior

Hey guys,

https://docs.paramiko.org/en/stable/api/transport.html

if you search through the documentation , auth_interactive() appears twice and auth_interactive_dumb( ) also appears twice. With different descriptions. 

also auth_password(username, password, fallback=True) mentions something about an event? but the method signature does not show any event handler present?

is it someone ctrl+v by error here?

### Actual behavior

Non-duplicated documentation section in https://docs.paramiko.org/en/stable/api/transport.html

### How to reproduce

Read the doc here https://docs.paramiko.org/en/stable/api/transport.html

on your browser, control+f and search for auth_interactive. 4 instances, 2 duplicated for dumb/not dumb

### Anything else?

Merry xmas, thank you for the library, i'll be willing to submit examples how to write a client that auths with 2FA and how to handle all the possible flow during logging, for standard ssh client/sshd configs.

sankalpsthakur:
#2658 already has an open draft PR (#2665) implementing BufferedIOBase on BufferedFile, so this change addresses the smallest unassigned non-security bug that did not already have an open PR.

`sites/docs/api/transport.rst` used `automodule`, which listed both `Transport` and `ServiceRequestingTransport`. The subclass overrode `auth_interactive`, `auth_interactive_dumb`, `auth_password`, and `auth_publickey` without its own docstrings, so:

- `auth_interactive` / `auth_interactive_dumb` appeared twice on the Transport API page
- `ServiceRequestingTransport.auth_password(username, password, fallback=True)` inherited the parent docstring that describes an `event` argument the method does not accept

This PR documents `ServiceRequestingTransport` separately (only its unique session/auth-handler methods) and gives the overridden auth methods short docstrings that match their signatures.

Fixes #2556

### DIFF
--- a/paramiko/transport.py
+++ b/paramiko/transport.py
@@ -3173,6 +3173,11 @@ class ServiceRequestingTransport(Transport):
     """
     Transport, but also handling service requests, like it oughtta!
 
+    Opt-in via `SSHClient.connect <paramiko.client.SSHClient.connect>`'s
+    ``transport_factory``. Authentication methods match `.Transport` but
+    always block: ``auth_password`` and ``auth_publickey`` have no ``event``
+    parameter.
+
     .. versionadded:: 3.2
     """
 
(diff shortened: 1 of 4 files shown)


bskinn:

Closing this PR due to declared AI authorship.

